PR #2, opened 2018-11-02
starlark: disallow list element update during iteration over list

alandonovan:
The notion of "structural" vs "nonstructural" updates was apparently
never more than a vague idea and was never implemented in Bazel.

Fixes google/skylark#141


alandonovan:
@laurentlb

PR #3, opened 2018-11-02
starlark: fix bugs in str.{title,capitalize}

alandonovan:
Behavior now matches the spec (and Python3).
+ Tests.

Fixes google/skylark#140

### DIFF
--- a/doc/spec.md
+++ b/doc/spec.md
@@ -3505,11 +3505,15 @@ See also: `string·elems`.
 <a id='string·capitalize'></a>
 ### string·capitalize
 
-`S.capitalize()` returns a copy of string S with all Unicode letters
-that begin words changed to their title case.
+`S.capitalize()` returns a copy of string S with its first code point
+changes to its title case and all subsequent letters changed to their
+lower case.
 
 ```python
-"hello, world!".capitalize()		# "Hello, World!"
+"hello, world!".capitalize()		# "Hello, world!"
+"hElLo, wOrLd!".capitalize()		# "Hello, world!"
+"¿Por qué?".capitalize()		# "¿por qué?"
+
 ```
 
 <a id='string·codepoint_ords'></a>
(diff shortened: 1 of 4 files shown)


alandonovan:
@laurentlb

doc/spec.md
```diff
@@ -3505,11 +3505,15 @@ See also: `string·elems`.
...
 
-`S.capitalize()` returns a copy of string S with all Unicode letters
-that begin words changed to their title case.
+`S.capitalize()` returns a copy of string S with its first code point
+changes to its title case and all subsequent letters changed to their
+lower case.
 
 ```python
 "hello, world!".capitalize()		# "Hello, World!"
+"hElLo, wOrLd!".capitalize()		# "Hello, World!"
```
laurentlb: the two comments look wrong (lowercase "w")

alandonovan: Well spotted! Thanks.

doc/spec.md
```diff
@@ -3505,11 +3505,14 @@ See also: `string·elems`.
 <a id='string·capitalize'></a>
 ### string·capitalize
 
-`S.capitalize()` returns a copy of string S with all Unicode letters
-that begin words changed to their title case.
+`S.capitalize()` returns a copy of string S with its first code point
```
jayconrod: Python says "first character" instead of "first code point". Is the difference intended?

In Python:

```
>>> s = '\uFB03'
>>> s
'ﬃ'
>>> s.capitalize()
'FFI'
>>> len(s.capitalize())
3
```

jayconrod: Sorry, this comment was too nitpicky. I don't think the Python behavior actually makes sense: given the wording, I'd expect it to return three characters: F f i.

I think having the spec say this applies to code points, not characters makes sense. It may be good to point out somewhere that this is different than Python though.

alandonovan: The Python function is even more underspecified than ours: it seems to attempt to uppercase (not capitalize) the decomposition of a lower case letter that has no corresponding uppercase letter. I don't want to go there.

Thanks for educating me about this particular naughty string. :)

alandonovan:
Scratch the surface of any program that uses Unicode and you find bugs---I found some more. Please take another look.

PR #11, opened 2018-11-05
fix nits found by various static tools

### LINKED ISSUE #10: Nits found by various static check tools
```
$ unconvert ./...
internal/chunkedfile/chunkedfile.go:61:18: unnecessary conversion
starlark/int.go:43:42: unnecessary conversion
starlark/value.go:427:16: unnecessary conversion
$ ineffassign .
starlark/library.go:286:2: ineffectual assignment to ok
$ staticcheck  ./... |grep -v ST1006 # ignore different receiver names
internal/compile/compile.go:1:1: package comment should be of the form "Package compile ..." (ST1000)
repl/repl.go:1:1: package comment should be of the form "Package repl ..." (ST1000)
starlark/hashtable.go:284:22: func (*hashtable).dump is unused (U1000)
starlark/library.go:140:6: func builtinMethodOf is unused (U1000)
starlark/library.go:1809:6: func string_lstrip is unused (U1000)
starlark/library.go:1867:6: func string_rstrip is unused (U1000)
starlark/value.go:1100:7: identical expressions on the left and right side of the '!=' operator (SA4000)
starlark/value.go:1115:7: identical expressions on the left and right side of the '!=' operator (SA4000)
syntax/scan.go:380:2: field triple is unused (U1000)
```

I can file a PR to fix these.  Just wondering if you had any opinions on them.  For example,  the `identical expressions` are using `y != y` directly instead of `math.IsNaN()`.  Will `string_lstrip` will be used in the future or can it be removed?   The `hashtable.dump` method is probably for debugging, so I'll leave that.  Package comment tweaks are minor and inoffensive.

alandonovan:
Fixes #10 
@dgryski

### DIFF
--- a/internal/chunkedfile/chunkedfile.go
+++ b/internal/chunkedfile/chunkedfile.go
@@ -58,7 +58,6 @@ func Read(filename string, report Reporter) (chunks []Chunk) {
 	}
 	linenum := 1
 	for i, chunk := range strings.Split(string(data), "\n---\n") {
-		chunk := string(chunk)
 		if debug {
 			fmt.Printf("chunk %d at line %d: %s\n", i, linenum, chunk)
 		}
--- a/internal/compile/compile.go
+++ b/internal/compile/compile.go
@@ -1,4 +1,4 @@
-// The compile package defines the Starlark bytecode compiler.
+// Package compile defines the Starlark bytecode compiler.
 // It is an internal package of the Starlark interpreter and is not directly accessible to clients.
 //
 // The compiler generates byte code with optional uint32 operands for a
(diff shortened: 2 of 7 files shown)


dgryski:
`ineffassign` and `unconvert` have no false positives, although occasionally `ineffassign` will complain if you declare variables with `:=` to the zero value before use rather than with `var`.  Many projects run `staticcheck` during CI, but for starlark I filtered out a few checks that I didn't consider worth fixing.  If you were going to run it as part of CI, you'd need to either fix those or add appropriate comments to let staticcheck to know ignore them (which generally is about as much work as fixing the issues to begin with...)

dgryski:
Why did the CI build fail with Go tip?

alandonovan:
> Why did the CI build fail with Go tip?

Because of longstanding flakiness in the ExampleThread_Load_parallelCycle test, fixed recently.

PR #20, opened 2018-11-21
all: use explicit URLs to refer to old Starlark issues

alandonovan:
GitHub allows individual issues to be transferred but provides no bulk
mechanism---also, we've started reusing issue numbers---so I
transferred the few open issues and left the closed ones.


alandonovan:
Thanks, I'll stick to URLs for clarity.

PR #22, opened 2018-11-21
starlark: reject dup keyword args in call to func with **kwargs param

### LINKED ISSUE #21: starlark: reject duplicate keyword arguments in function call even when callee has **kwargs
```
$ cat a.star
def f(*args, **kwargs):
        print(args, kwargs)

f(x=1, y=2)
f(x=1, x=2)

$ starlark a.star
() {"x": 1, "y": 2}
() {"x": 2} # should be an error
```

Python2, Python3 and Starlark-in-Java all reject repetition of a keyword argument in a function call. Starlark-in-Go should to.  (It does when the callee is a regular function but not when it has a **kwargs parameter.)

alandonovan:
Fixes #21

### DIFF
--- a/starlark/testdata/function.star
+++ b/starlark/testdata/function.star
@@ -174,6 +174,23 @@ assert.fails(lambda: f(
     57, 58, 59, 60, 61, 62, 63, 64, 65,
     mm = 100), 'multiple values for keyword argument "mm"')
 
+---
+# Regression test for github.com/google/starlark-go/issues/21.
+
+load("assert.star", "assert")
+
+def f(*args, **kwargs):
+  return args, kwargs
+
+assert.eq(f(x=1, y=2), ((), {"x": 1, "y": 2}))
+assert.fails(lambda: f(x=1, x=2), 'multiple values for keyword argument "x"')
+assert.fails(lambda: f(x=1, **dict(x=2)), 'multiple values for keyword argument "x"')
+
+def g(x, y):
+  return x, y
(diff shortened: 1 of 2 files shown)


starlark/testdata/function.star
```diff
@@ -174,6 +174,22 @@ assert.fails(lambda: f(
...
+  return args, kwargs
+
+assert.eq(f(x=1, y=2), ((), {"x": 1, "y": 2}))
+assert.fails(lambda: f(x=1, x=2), 'multiple values for keyword argument "x"')
+
+def g(x, y):
+  return x, y
+
+assert.eq(g(1, y=2), (1, 2))
+assert.fails(lambda: g(1, y=2, y=3), 'multiple values for keyword argument "y"')
```
laurentlb: Add test:
  `f(x=1, **{"x": 2})`

alandonovan: Good idea. Done.

alandonovan:
Oops, must stop force-pushing. This is not gerrit. :(

dgryski:
@alandonovan force pushing to a PR branch before a merge is fine. Did you want to maintain the old commits?

alandonovan:
I was advised that GitHub code reviews are most intelligible if one pushes a sequence of commits and only squashed them when merging, rather than rewriting history on the review branch, as we do in the Go project's Gerrit workflow, so that it always appears as a single commit.

PR #30, opened 2018-11-26
Update README with information about https://github.com/bazelbuild/starlark

laurentlb:


README.md
```diff
@@ -105,7 +105,8 @@ avoid duplication of effort.
 Despite some differences, the Go implementation of Starlark strives to
 match the behavior of the Java implementation used by Bazel.
 For that reason, proposals to change the language itself should
-generally be directed to [the Bazel team](https://github.com/bazelbuild/bazel/issues),
+generally be directed to [the Starlark repository](
```
alandonovan: The text must make the organizational distinction clear.  How about:

```
Despite some differences, the Go implementation of Starlark strives to
match the behavior of the
[Java implementation](https://github.com/bazelbuild/starlark/)
used by Bazel and maintained by the Bazel team.
For that reason, proposals to change the language itself should
generally be directed to the Bazel team.
```

laurentlb: Not exactly, because https://github.com/bazelbuild/starlark/ aims to be implementation-independent.

PR #34, opened 2018-11-28
starlark: fix overflow in repeat operator

alandonovan:
We limit repeats arbitrarily to 10,000,000 elements
to avoid accidental OOM. It's still possible to
OOM piecewise but we can't fix that.

Also, handle integer overflow in multiplication
and excessive repeats counts.

### DIFF
--- a/starlark/eval.go
+++ b/starlark/eval.go
@@ -544,20 +544,15 @@ func Binary(op syntax.Token, x, y Value) (Value, error) {
 			case Float:
 				return x.Float() * y, nil
 			case String:
-				if i, err := AsInt32(x); err == nil {
-					if i < 1 {
-						return String(""), nil
-					}
-					return String(strings.Repeat(string(y), i)), nil
-				}
+				return stringRepeat(y, x)
 			case *List:
-				if i, err := AsInt32(x); err == nil {
-					return NewList(repeat(y.elems, i)), nil
+				elems, err := tupleRepeat(Tuple(y.elems), x)
+				if err != nil {
+					return nil, err
 				}
+				return NewList(elems), nil
 			case Tuple:
-				if i, err := AsInt32(x); err == nil {
-					return Tuple(repeat([]Value(y), i)), nil
-				}
+				return tupleRepeat(y, x)
 			}
 		case Float:
 			switch y := y.(type) {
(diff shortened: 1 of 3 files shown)


alandonovan:
@josharian

starlark/eval.go
```diff
@@ -836,14 +826,50 @@ unknown:
...
+func tupleRepeat(elems Tuple, n Int) (Tuple, error) {
+	if len(elems) == 0 {
+		return nil, nil
+	}
+	i, err := AsInt32(n)
+	if err != nil {
+		return nil, fmt.Errorf("repeat count %s too large", n)
+	}
+	if i < 1 {
+		return nil, err
```
jayconrod: `nil, nil` would be more clear.

starlark/eval.go
```diff
@@ -836,14 +826,50 @@ unknown:
...
+	}
+	// Inv: i > 0, len > 0
+	sz := len(elems) * i
```
jayconrod: Is this actually 32-bit multiplication? `len(elems)` should be `int`, which would be 64-bit on`amd64`, so it would be 64-bit multiplication, right?

I think it would be clearer to check that `len(elems)` can be cast to a 32-bit integer without loss, then do the multiplication. You shouldn't need to check `sz < 0` after that.

starlark/eval.go
```diff
@@ -836,14 +826,50 @@ unknown:
...
+	}
+	i, err := AsInt32(n)
+	if err != nil {
+		return "", fmt.Errorf("repeat count %s too large", n)
+	}
+	if i < 1 {
+		return "", nil
+	}
+	// Inv: i > 0, len > 0
+	sz := len(s) * i
```
jayconrod: Same comment as above.

starlark/eval.go
```diff
@@ -836,14 +826,50 @@ unknown:
...
 
-func repeat(elems []Value, n int) (res []Value) {
-	if n > 0 {
-		res = make([]Value, 0, len(elems)*n)
-		for i := 0; i < n; i++ {
-			res = append(res, elems...)
-		}
+// It's always possible to overeat in small bites but we'll
+// try to stop someone swallowing the world in one gulp.
+const maxAlloc = 10000000
```
jayconrod: I'm not really confident in having a limit like this. It's small enough that a large program could conceivably hit it (particularly if you want to allocate `[0] * N`, then fill in the values later). But it's large enough that minimal platforms could still OOM. And it doesn't protect from malicious inputs that OOM.

starlark/eval.go
alandonovan: It sure would. Thanks, that was an oversight.

starlark/eval.go
alandonovan: No, it's int multiplication, which is either 32 or 64. It can overflow on 64, though that's not realistic because i is < 2^32 which means the len would have to be larger than 2^32. I removed the "32" from the comment to make it more accurate.

I would rather not downgrade to 32 bits if the machine is 64, as is the norm. The <0 check is the usual way to detect multiplication overflow.

starlark/eval.go
alandonovan: The purpose of the limit is to prevent the most trivial mistakes from killing the program (and possibly the machine---sadly this happens all too often), not to stop malicious DoS, which is essentially impossible in a GC language.

I agree it's imperfect, but every memory allocator imposes some limit. The Go runtime caps malloc at 0x1000 << 32 on 64-bits or 1<<32 on a small machine, but in reality bad things happen long before you get there. I would rather start with a small limit and raise it as needed (probably never).

jayconrod:
PS, GitHub seems to be happy with force pushes to PR branches now. Comments are showing up on earlier commits without any trouble. I'm glad they fixed that.

starlark/eval.go
alandonovan: I upped the limit to about a billion elements since that's large (by Starlark standards) but still falls into the range malloc can easily handle.

alandonovan:
> PS, GitHub seems to be happy with force pushes to PR branches now. Comments are showing up on earlier commits without any trouble. I'm glad they fixed that.

Me too, as I keep forgetting not to force-push.

Issue #35, opened 2018-11-30
[feature request] Support python-like docstrings


aarzilli:
The language already allows an statement consisting only of a literal string, to appear as the first statement of a function. I propose to add a `Doc()` method to `*Function` that will return the value of this string.

**Backwards compatibility**: this doesn't change the syntax of starlark nor its behavior.

**Motivation**: I want to use starlark as scripting language for delve, I intend to scan the list of global variables returned by ExecFile and bind any global function that starts with "command_" to a command accessible from delve's command line. All the predefined commands have a documentation string that is consumed by the "help" command. I would like starlark-defined commands to also have this.

**Alternatives**: collect the comments preceding the function definition and return them somehow. Looking at the syntax package it seems to do some of this already but I don't think it's exposed anywhere, also poking at it with a debugger it doesn't look like the commentsRef of DefStmt actually ever gets set to anything (but maybe I just don't get what it's supposed to do).

aarzilli:
The change is probably trivial and I can probably do it myself, I just want to know if this is acceptable.

alandonovan:
Sounds fine. The only downside is that this will bloat the serialized bytecode files, since the comments can easily double the size of the compiled program. But it should be an easy change---send me a patch and I will measure whether it has performance implications on a large benchmark I can't easily share.

deckarep:
Question on this: since this got merged should the grammar.txt file be updated to reflect that an optional docstring is now supported within a Suite?

alandonovan:
> Question on this: since this got merged should the grammar.txt file be updated to reflect that an optional docstring is now supported within a Suite?

There is no change to the grammar. A string literal is an expression, and any expression is a valid statement. All that has changed is the representation of function values whose first statement is a string literal.

deckarep:
@alandonovan - thx for clarifying.

PR #37, opened 2018-11-30
Support doc strings

aarzilli:
If the first statement of the body of a function is a string its value
will be returned when calling the Doc() method of starlark.(*Function).

Implements #35

### DIFF
--- a/internal/compile/compile.go
+++ b/internal/compile/compile.go
@@ -300,6 +300,7 @@ type Funcode struct {
 	Prog                  *Program
 	Pos                   syntax.Position // position of def or lambda token
 	Name                  string          // name of this function
+	Doc                   string          // docstring of this function
 	Code                  []byte          // the byte code
 	pclinetab             []uint16        // mapping from pc to linenum
 	Locals                []Ident         // for error messages and tracing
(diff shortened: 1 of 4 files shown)


starlark/exec_file_test.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"testing"
+)
+
+func TestDocstring(t *testing.T) {
+	globals, _ := ExecFile(&Thread{}, "", `
+def somefunc():
+	"somefunc doc"
+	return 0
+`, StringDict{})
```
alandonovan: s/StringDict{}/nil/

starlark/exec_file_test.go
```diff
@@ -0,0 +1,17 @@
+package starlark
+
+import (
+	"testing"
+)
+
+func TestDocstring(t *testing.T) {
```
alandonovan: Cold you move this function into starlark/eval_test?

starlark/exec_file_test.go
```diff
@@ -0,0 +1,17 @@
...
+	globals, _ := ExecFile(&Thread{}, "", `
```
alandonovan: Use "doc.star", not an empty string.

aarzilli:
Rebased.

PR #43, opened 2018-12-05
starlark: drop support for dict+dict

alandonovan:
This feature (which Python does not have) has been deprecated for a long time.


alandonovan:
@laurentlb

laurentlb:
Thanks!

PR #58, opened 2018-12-10
syntax: fix bug in parser

alandonovan:
(introduced in 6be536f)

Also factor similar *args/*kwargs parsing in the same way.


alandonovan:
(Jay: consider this a TBR commit.)

PR #59, opened 2018-12-10
[parser] Unify handling of * and **.

ttsugriy:
This is similar to a recent [change](https://github.com/google/starlark-go/pull/57/files) to `parseParams` method.


ttsugriy:
hm, looks like it's fixed now

PR #62, opened 2018-12-10
Support keyword-only function parameters

### LINKED ISSUE #61: support kwonlyargs
Starlark in Java, like Python3 but not Python3, has the concept of keyword-only arguments (kwonlyargs).

This statement:
  def f(a, b, *args, c=1, d=2, **kwargs):  pass
declares a function with two kwonlyargs, c and d, as does this one:
  def f(a, b, *, c=1, d=2, **kwargs):  pass
except the allows variadic calls and the second does not.

The representation of function values will need to record the length of the suffix of the parameter tuple that are kwonlyargs, and the call operator must reject calls that would assign to any of them using a positional argument.

alandonovan:
Following an undocumented feature of Skylark-in-Java,
which in turn follows Python3,
a function declaration may now include optional parameters
after the *args parameter:

	def f(a, b, c=1, *args, d=2, **kwargs)

The parameter d is a "keyword-only" parameter as it can
never by assigned from a positional parameter; all positional
arguments surplus to a, b, and c are put in a tuple and
assigned to args.

To declare a non-variadic function with keyword-only arguments,
the *args parameter is replaced by just *:

        def f(a, b, c=1, *, d=2, **kwargs)

The * parameter is not a real parameter; it just serves as a
separator between the parameter that may be specified positionally
and the keyword-only ones.

Fixes #61

### DIFF
--- a/doc/spec.md
+++ b/doc/spec.md
@@ -2321,6 +2321,7 @@ LambdaExpr = 'lambda' [Parameters] ':' Test .
 Parameters = Parameter {',' Parameter} .
 Parameter  = identifier
            | identifier '=' Test
+           | '*'
            | '*' identifier
            | '**' identifier
            .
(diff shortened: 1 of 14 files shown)


alandonovan:
@laurentlb

alandonovan:
Hmm, hold off on this for now: there appears to be some confusion about whether this feature is to be added or taken away; see https://github.com/bazelbuild/starlark/issues/13#issuecomment-446005938.

adonovan:
Not needed.

PR #70, opened 2018-12-13
better error message for misuse of break and continue

aarzilli:
```
better error message for misuse of break and continue

Output an error message instead of a compiler panic when break and
continue are used without any loop.

```


aarzilli:
I see that resolve is already checking this and the problem doesn't exist after #71. So this isn't useful.

PR #71, opened 2018-12-13
make break and continue work with while

aarzilli:
```
make break and continue work with while

break and continue should work inside a while loop instead of causing a
compiler panic.

```

### DIFF
--- a/starlark/eval_test.go
+++ b/starlark/eval_test.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"math"
 	"path/filepath"
+	"strings"
 	"testing"
 
 	"go.starlark.net/internal/chunkedfile"
@@ -109,6 +110,7 @@ func TestExecFile(t *testing.T) {
 		"testdata/set.star",
 		"testdata/string.star",
 		"testdata/tuple.star",
+		"testdata/recursion.star",
 	} {
 		filename := filepath.Join(testdata, file)
 		for _, chunk := range chunkedfile.Read(filename, t) {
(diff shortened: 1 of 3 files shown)


starlark/eval_test.go
```diff
@@ -497,10 +497,48 @@ sum5 = sum(5)
 	if err != nil {
 		t.Fatal(err)
 	}
-	if got, _ := globals["fib5"].(starlark.Int).Int64(); int(got) != 8 {
-		t.Fatalf("wrong value for fib5, got %d expected 8\n", got)
+	if got, _ := globals["fib5"].(starlark.Int).Int64(); got != 8 {
```
alandonovan: The production code change looks good---thanks (and sorry for completely failing to notice the obvious omission.)

But I think we can improve the tests. Let's move the work of TestRecursion into the main test suite, which is a set of *.star files.  The only reason it is written in Go is because we need a way to temporarily set  resolve.AllowRecursion. If you look at TestResolve in the resolve directory, you'll see it has a simple mechanism to allow chunks of the test files to set the global dialect flags:
```
		// A chunk may set options by containing e.g. "option:float".
		resolve.AllowNestedDef = option(chunk.Source, "nesteddef")
		resolve.AllowLambda = option(chunk.Source, "lambda")
		resolve.AllowFloat = option(chunk.Source, "float")
		resolve.AllowSet = option(chunk.Source, "set")
		resolve.AllowGlobalReassign = option(chunk.Source, "global_reassign")
...
func option(chunk, name string) bool {
	return strings.Contains(chunk, "option:"+name)
}
```

Let's use the same mechanism in the evaluator tests. Then we can move all of the Starlark code into proper Starlark files and use the assert module for the assertions.

aarzilli:
Like this?

starlark/eval_test.go
```diff
@@ -109,13 +110,17 @@ func TestExecFile(t *testing.T) {
...
+		"testdata/recursion.star",
 	} {
 		filename := filepath.Join(testdata, file)
 		for _, chunk := range chunkedfile.Read(filename, t) {
 			predeclared := starlark.StringDict{
 				"hasfields": starlark.NewBuiltin("hasfields", newHasFields),
 				"fibonacci": fib{},
 			}
+
+			resolve.AllowRecursion = option(chunk.Source, "recursion")
```
alandonovan: Ideally all the other flags would be off by default and enabled in this way too. We can do that as a followup.

PR #72, opened 2018-12-14
resolve: statically reject duplicate keyword args in a call

alandonovan:
This complements the dynamic checks, which is already implemented
in setArgs and in UnpackeArgs.

Updates bazelbuild/starlark#21


alandonovan:
@laurentlb

PR #76, opened 2018-12-17
starlark: API additions for improved debugging

alandonovan:
This change adds a number of small features to improve debugging.
The actual debugger API will come in a later change.

- Thread.Name : an optional string field that describes the purpose of
  the thread, for use in debugging.
- (*Program).Filename: a method that reports the file of the program.
  Also, a String method that returns the same thing.
  Also, a test that it reports the correct location even for
  an empty file (a special case of the previous implementation).
- (*Frame).Local(i int): a method to return the value of a local
  variable of an active frame, such as one might use in debugger's
  stack trace.
- ExprFunc: creates a starlark.Function from a given expression,
  such as one might use in a debugger REPL.

### DIFF
--- a/starlark/debug.go
+++ b/starlark/debug.go
@@ -1,0 +1,18 @@
+package starlark
+
+// This file defines an experimental API for the debugging tools.
+// Some of these declarations expose details of internal packages.
+// (The debugger makes liberal use of exported fields of unexported types.)
+// Breaking changes may occur without notice.
+
+// Local returns the value of the i'th local variable.
+// It may be nil if not yet assigned.
+//
+// Local may be called only for frames whose Callable is a *Function (a
+// function defined by Starlark source code), and only while the frame
+// is active; it will panic otherwise.
+//
+// This function is provided only for debugging tools.
+//
+// THIS API IS EXPERIMENTAL AND MAY CHANGE WITHOUT NOTICE.
+func (fr *Frame) Local(i int) Value { return fr.locals[i] }
--- a/starlark/eval_test.go
+++ b/starlark/eval_test.go
@@ -337,11 +337,11 @@ f()
 			caller.Position(), caller.Callable().Name(), msg)
 	}
 	thread := &starlark.Thread{Print: print}
-	if _, err := starlark.ExecFile(thread, "foo.go", src, nil); err != nil {
+	if _, err := starlark.ExecFile(thread, "foo.star", src, nil); err != nil {
 		t.Fatal(err)
 	}
-	want := "foo.go:2: <toplevel>: hello\n" +
-		"foo.go:3: f: hello, world\n"
+	want := "foo.star:2: <toplevel>: hello\n" +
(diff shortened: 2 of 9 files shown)


alandonovan:
@cococlyde55

alandonovan:
@cococlyde

starlark/example_test.go
```diff
@@ -90,7 +90,7 @@ func ExampleThread_Load_sequential() {
 
 			// Load and initialize the module in a new thread.
 			data := fakeFilesystem[module]
-			thread := &starlark.Thread{Load: load}
+			thread := &starlark.Thread{Name: "exec " + module, Load: load}
```
jayconrod: There's one more `&starlark.Thread{...}` above. Does that need a `Name`?

starlark/debug.go
```diff
@@ -0,0 +1,18 @@
...
+// Local returns the value of the i'th local variable.
+// It may be nil if not yet assigned.
+//
+// Local may be called only for frames whose Callable is a *Function (a
+// function defined by Starlark source code), and only while the frame
+// is active; it will panic otherwise.
+//
+// This function is provided only for debugging tools.
+//
+// THIS API IS EXPERIMENTAL AND MAY CHANGE WITHOUT NOTICE.
```
jayconrod: Since users are encouraged not to call this, I wonder if it would be better to name this `DebugLocal`? People might not read the docs.

starlark/eval_test.go
```diff
@@ -452,6 +452,21 @@ func TestRepeatedExec(t *testing.T) {
 	}
 }
 
+// TestEmptyFilePosition ensures that even Programs
+// from empty files have a value position.
```
jayconrod: s/value/valid/

starlark/interp.go
```diff
@@ -561,5 +563,8 @@ loop:
 			err = fr.errorf(f.Position(savedpc), "%s", err.Error())
 		}
 	}
+
+	fr.locals = nil
```
jayconrod: This would be a bit more readable in a `defer` after the original assignment. Is there a significant overhead for  that?

starlark/debug.go
alandonovan: There are going to be a number of functions that are intended for debugger-like tools and other diagnostic machinery but that shouldn't generally be used to influence the behavior of a Starlark program. I don't want to have to have to put Debug in all their names. Most clients will have little reason to find themselves using this; the comment is fair warning to "advanced" aka "overly curious" users.

starlark/eval_test.go
alandonovan: Fixed.

starlark/example_test.go
alandonovan: Done.

starlark/interp.go
alandonovan: Yes, which is why this function has exactly one return statement. At some point I should add a static check.

PR #77, opened 2018-12-18
starlark: permit if/for/while at toplevel if -globalreassign

alandonovan:
Also:
- make the option handling work the same across all tests
- be explicit about which options are needed in each test chunk
- improve documentation of options
- rename fp option to float consistently
- rename global_reassign option to globalreassign consistently

### DIFF
--- a/doc/spec.md
+++ b/doc/spec.md
@@ -16,8 +16,7 @@ used by Bazel.
 We identify places where their behaviors differ, and an
 [appendix](#dialect-differences) provides a summary of those
 differences.
-We plan to converge both implementations on a single specification
-in early 2018.
+We plan to converge both implementations on a single specification.
 
 This document is maintained by Alan Donovan <adonovan@google.com>.
 It was influenced by the Python specification,
@@ -241,11 +240,10 @@ characters are tokens:
 identifiers:
 
 ```text
-and            else           load
-break          for            not
-continue       if             or
-def            in             pass
-elif           lambda         return
+and            elif           in             or
+break          else           lambda         pass
+continue       for            load           return
+def            if             not            while
 ```
 
 The tokens below also may not be used as identifiers although they do not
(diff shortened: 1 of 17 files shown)


glycerine:
It could be my merge mistake, but its worth checking at the top level that `while` will work, even with resolve.AllowGlobalReassign = true
~~~
n=6
while n > 0:
   n=n-1
   print(n)

~~~
I see endings 5s:
~~~
5
5
5
5
5
....
~~~

glycerine:
possibly related
~~~
m=6 # at toplevel
def g():
  while m>0:
    m=m-1
    print(m)

~~~
gives
~~~
>>> g()
  <stdin>:1: in <expr>
  <stdin>:2: in g
Error: local variable m referenced before assignment
~~~

alandonovan:
Works for me at this commit:
```
$ git co options
Already on 'options'
Your branch is up to date with 'origin/options'.
$ go build -o x ./cmd/starlark
$ cat a.go
n=6
while n > 0:
   n=n-1
   print(n)

$ ./x -recursion -globalreassign a.go
5
4
3
2
1
0
```
If you can reproduce the problem at this or any other commit, please do file a bug. 

Regarding "local variable m referenced before assignment": yes, that's the intended behavior. Starlark does not have Python's global (or nonlocal) keyword so you just have to choose different variable names. The global and nonlocal statements would be easy enough to implement, but given the easy workaround, there's little urgency.

doc/spec.md
```diff
@@ -254,14 +252,11 @@ appear in the grammar; they are reserved as possible future keywords:
...
-from           with
-global         yield
+as             finally        nonlocal
+assert         from           raise
+class          global         try
+del            import         with
+except         is             yield
 ```
 
 <b>Implementation note:</b>
```
jayconrod: Consider adding implementation note for `while`, since it's not a keyword in the default grammar.

PR #79, opened 2018-12-24
StringDict implements Value and HasAttrs

glycerine:
allows StringDict to act as a package, containing Values
  and responding to attribute queries.


alandonovan:
This change is unnecessary. We already have a data type that has the characteristics this PR proposes for StringDict: starlarkstruct.

That said, I think we may want a variant of starlarkstruct to represent modules, if only so that when printed it doesn't show all its keys, but StringDict is not it.

PR #86, opened 2019-01-02
starlark: add StringDict.Keys method

alandonovan:
...and remove three ad-hoc implementations.

### DIFF
--- a/cmd/starlark/starlark.go
+++ b/cmd/starlark/starlark.go
@@ -12,7 +12,6 @@ import (
 	"log"
 	"os"
 	"runtime/pprof"
-	"sort"
 	"strings"
 
 	"go.starlark.net/repl"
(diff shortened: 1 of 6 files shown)


josharian:
LGTM.

Two observations from reading surrounding code:

We might want to migrate from bytes.Buffer to strings.Builder in a few places. I can send a PR for that later.

The handling of path in writeValue is subtle. I had to read carefully to see that it didn’t need to be reset on every iteration through the dict keys. I’d suggest adding some docs. Worth mentioning in those docs that the path argument passed in should be short-lived on pain of pinning memory. (Or better, nil out recently appended elements after each recursive call.)

josharian:
(I can send a PR for the second comment as well. Just let me know for each whether you’ll handle or you’re waiting for me.)

josharian:
> Thanks. I keep thinking of strings.Builder as too new, but now that 1.12 is out, it's fair game. I would be grateful for that CL.

1.12 should come out in Feb. I'll file a reminder issue to do this once it has been released.

> I agree the use of append in this way is a little subtle, but what would you add beyond

Given that path doesn't escape, there's not that much value to being able to re-use it across iterations. I'd consider having writeValue lazily initialize path as needed, and making most callers pass in `nil`. (And maybe add a wrapper that doesn't take a path parameter at all.) Would need to check that perf is ok for that.

If it was a doc-only change, I'd suggest something like:

```go
// writeValue writes x to out.
// path is used to detect cycles.
// It contains the list of *List and *Dict values we're currently printing.
// (These are the only potentially cyclic structures.)
// Callers should generally pass a temporary slice of length zero but non-zero capacity.
// It is safe to re-use the same path slice for multiple calls.
```

alandonovan:
Thanks. I've added the comment, removed the preallocations, and verified that it does not escape.

PR #87, opened 2019-01-02
starlark: HasUnary: allow extension types to define +x, -x, ~x

alandonovan:
This mirrors the treatment of HasBinary.

### DIFF
--- a/internal/compile/compile.go
+++ b/internal/compile/compile.go
@@ -266,8 +266,10 @@ var stackEffect = [...]int8{
 	SLICE:       -3,
 	STAR:        -1,
 	TRUE:        +1,
+	UMINUS:      0,
 	UNIVERSAL:   +1,
 	UNPACK:      variableStackEffect,
+	UPLUS:       0,
 }
 
 func (op Opcode) String() string {
--- a/starlark/eval.go
+++ b/starlark/eval.go
@@ -478,26 +478,20 @@ func setIndex(fr *Frame, x, y, z Value) error {
 
 // Unary applies a unary operator (+, -, ~, not) to its operand.
 func Unary(op syntax.Token, x Value) (Value, error) {
-	switch op {
-	case syntax.MINUS:
-		switch x := x.(type) {
-		case Int:
-			return zero.Sub(x), nil
-		case Float:
-			return -x, nil
-		}
-	case syntax.PLUS:
-		switch x.(type) {
-		case Int, Float:
-			return x, nil
-		}
-	case syntax.TILDE:
-		if xint, ok := x.(Int); ok {
-			return xint.Not(), nil
-		}
-	case syntax.NOT:
+	// The NOT operator is not customizable.
+	if op == syntax.NOT {
 		return !x.Truth(), nil
 	}
+
+	// Int, Float, and user-defined types
+	if x, ok := x.(HasUnary); ok {
+		// (nil, nil) => unhandled
+		y, err := x.Unary(op)
+		if y != nil || err != nil {
+			return y, err
+		}
+	}
+
 	return nil, fmt.Errorf("unknown unary op: %s %s", op, x.Type())
 }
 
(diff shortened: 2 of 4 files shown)


starlark/eval.go
```diff
@@ -478,26 +478,18 @@ func setIndex(fr *Frame, x, y, z Value) error {
...
-		if xint, ok := x.(Int); ok {
-			return xint.Not(), nil
-		}
-	case syntax.NOT:
+	if op == syntax.NOT {
 		return !x.Truth(), nil
 	}
+
+	// Int, Float, and user-defined types
+	if x, ok := x.(HasUnary); ok {
```
jayconrod: No change needed in this PR, but does it make sense to move logic from `Binary` to `HasBinary` implementations for builtin types? I can't remember if there was a good reason to have everything in `Binary` or if `HasBinary` simply came later.

internal/compile/compile.go
```diff
@@ -266,8 +266,10 @@ var stackEffect = [...]int8{
 	SLICE:       -3,
 	STAR:        -1,
 	TRUE:        +1,
+	UMINUS:      0,
```
jayconrod: Just to confirm, this is just for readability, right? It's not strictly necessary to set zero values here?

internal/compile/compile.go
```diff
@@ -266,8 +266,10 @@ var stackEffect = [...]int8{
...
 	UNIVERSAL:   +1,
 	UNPACK:      variableStackEffect,
+	UPLUS:       0,
```
josharian: UTILDE?

starlark/eval.go
```diff
@@ -478,26 +478,18 @@ func setIndex(fr *Frame, x, y, z Value) error {
...
+		y, err := x.Unary(op)
+		if y != nil || err != nil {
```
josharian: Should this be `y == nil`?

starlark/int.go
```diff
@@ -74,6 +76,19 @@ func init() {
...
+func (i Int) Unary(op syntax.Token) (Value, error) {
+	switch op {
+	case syntax.MINUS:
+		return zero.Sub(i), nil
+	case syntax.PLUS:
+		return i, nil
+	case syntax.TILDE:
+		return i.Not(), nil
+	}
+	return nil, nil
```
josharian: `panic("unreachable")` instead?

starlark/value.go
```diff
@@ -34,6 +34,7 @@
 //      HasSetField     -- value has settable fields x.f
 //      HasSetIndex     -- value supports element update using x[i]=y
 //      HasSetKey       -- value supports map update using x[k]=v
+//      HasUnary        -- value defines unary operations such as + and -
```
josharian: May as well exhaustively list all unary operations; you're almost there. ;)

starlark/eval.go
josharian: Oh, I see later that `nil, nil` has a special meaning. This deserves a comment here, because it definitely reads like a typo.

starlark/value.go
josharian: Also, note that `not` is not customizable, but instead works off truthiness.

starlark/eval.go
alandonovan: It seems tempting to do so, but it can't be done in languages without multimethod dispatch. This is a classic problem in object-oriented languages called the "binary method problem"; see for example
https://lib.dr.iastate.edu/cgi/viewcontent.cgi?article=1032&context=cs_techreports

internal/compile/compile.go
alandonovan: Yes, just for completeness.

internal/compile/compile.go
alandonovan: TILDE is always unary, so it's just called TILDE.

starlark/eval.go
alandonovan: Done.

PR #90, opened 2019-01-02
starlark: expose NewFrame operation

alandonovan:
This allows clients to synthesize stack frames for non-Starlark (e.g. Go) function calls.
(See wip-stargo branch.)

### DIFF
--- a/starlark/eval.go
+++ b/starlark/eval.go
@@ -122,6 +122,14 @@ type Frame struct {
 	locals   []Value         // local variables, for debugger
 }
 
+// NewFrame returns a new frame with the specified parent and callable.
+//
+// It may be used to synthesize stack frames for error messages.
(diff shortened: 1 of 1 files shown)


starlark/eval.go
```diff
@@ -119,6 +119,11 @@ type Frame struct {
 	locals   []Value         // local variables, for debugger
 }
 
+// NewFrame returns a new frame with the specified parent and callable.
```
jayconrod: Maybe explain more about when this would be used? Sounds like there is a specific purpose from the PR description.

alandonovan: Done: 
```
// It may be used to synthesize stack frames for error messages.
// Few clients should need to use this function.
```
Basically, these are not the droids you are looking for.

PR #91, opened 2019-01-02
starlark: add 'module' type, and use it for assert module

alandonovan:
...instead of struct, which is more suited to dynamic values.

This type will be used for Stargo (see wip-stargo branch).
For example, this load statement

  load("go", http="net/http", json="encoding/json")

would load two modules.

Aside: Starlark's load statement is arguably its weakest feature.
Python and Go both have an 'import' statement that loads a collection
of names at once, such as "import os"; later uses are qualified, such
as os.mkdir. This is what we want for assert, for Stargo, and often in
Bazel, where it is tedious to enumerate every name when importing a
module with a broad API. There is room for improvement in the spec.

### DIFF
--- a/starlarktest/assert.star
+++ b/starlarktest/assert.star
@@ -1,45 +1,45 @@
-
 # Predeclared built-ins for this module:
 #
 # error(msg): report an error in Go's test framework without halting execution.
 # catch(f): evaluate f() and returns its evaluation error message, if any
 # matches(str, pattern): report whether str matches regular expression pattern.
-# struct: a constructor for a simple HasFields implementation.
+# module(**kwargs): a constructor for a module.
 # _freeze(x): freeze the value x and everything reachable from it.
 #
 # Clients may use these functions to define their own testing abstractions.
 
 def _eq(x, y):
-  if x != y:
-    error("%r != %r" % (x, y))
+    if x != y:
+        error("%r != %r" % (x, y))
 
 def _ne(x, y):
-  if x == y:
-    error("%r == %r" % (x, y))
+    if x == y:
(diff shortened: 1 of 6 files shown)


starlark/value.go
```diff
@@ -927,6 +928,49 @@ func (s *Set) Union(iter Iterator) (Value, error) {
 	return set, nil
 }
 
+// A Module is a named collection of values,
+// typically a suite of functions imported by a load statement.
+//
+// It differs from go.starlark.net/starlarkstruct.Struct primarily
+// in that its string representation does not enumerate its fields.
```
jayconrod: This doesn't seem like sufficient motivation to introduce a new type, especially a builtin type (`starlarkstruct` is not part of the spec or the main package). But I think I'm missing a lot of stargo context.

starlarktest/assert.star
```diff
@@ -1,45 +1,45 @@
...
+    if msg == None:
+        error("evaluation succeeded unexpectedly (want error matching %r)" % pattern)
+    elif not matches(pattern, msg):
+        error("regular expression (%s) did not match error (%s)" % (pattern, msg))
 
-freeze = _freeze # an exported global whose value is the built-in freeze function
+freeze = _freeze  # an exported global whose value is the built-in freeze function
 
-assert = struct(
+assert = module(
```
jayconrod: FYI, [unittest.bzl](https://github.com/bazelbuild/bazel-skylib/blob/6dd0b9cbd67478e20361933a994ca01fc1758f43/lib/unittest.bzl#L311) in Skylib is kind of similar. `struct` is used in each Skylib file to define a module (sometimes more than one module).

PR #94, opened 2019-01-02
resolve: allow if statement outside of function body, behind config flag

jazzdan:
It seemed like this was a feature that starlark-go would be interested in based on [this issue](https://github.com/bazelbuild/starlark/issues/17). Happy to close and move to discussion on an issue if that's not the case.


jazzdan:
whoops, this is covered by https://github.com/google/starlark-go/pull/77! Closing.

PR #115, opened 2019-01-15
starlark: disallow {+1} in format strings

### LINKED ISSUE #114: reject +0 as field identifier in string.format
```
$ starlark -c "'{+0}'.format(9)"
$ python3 -c "'{+0}'.format(9)"
  File "<string>", line 1, in <module>
KeyError: '+0'
$ python2 -c "'{+0}'.format(9)"
  File "<string>", line 1, in <module>
KeyError: '+0'
```

The spec says: `The *field name* may be either a decimal number or a keyword.` Decimal number isn't defined, although python rejects +0. starlark rejects other weird possible decimal numbers:

```
$ starlark -c "'{0.0}'.format(9)"
  cmdline:1: in <toplevel>
Error: attribute syntax x.y is not supported in replacement fields: 0.0
$ starlark -c "'{-0}'.format(9)"
  cmdline:1: in <toplevel>
Error: keyword -0 not found
```

So I think we should probably reject +0 (and +1 etc).

alandonovan:
+Test

Fixes #114

### DIFF
--- a/starlark/library.go
+++ b/starlark/library.go
@@ -1724,7 +1724,7 @@ func string_format(fnname string, recv_ Value, args Tuple, kwargs []Tuple) (Valu
 			}
 			arg = args[index]
 			index++
-		} else if num, err := strconv.Atoi(name); err == nil && !strings.HasPrefix(name, "-") {
+		} else if num, ok := decimal(name); ok {
 			// positional argument
 			if auto {
 				return nil, fmt.Errorf("cannot switch from automatic field numbering to manual field specification")
@@ -1778,6 +1778,22 @@ func string_format(fnname string, recv_ Value, args Tuple, kwargs []Tuple) (Valu
 		}
 	}
 	return String(buf.String()), nil
+}
+
+// decimal interprets s as a sequence of decimal digits.
+func decimal(s string) (x int, ok bool) {
+	n := len(s)
+	for i := 0; i < n; i++ {
+		digit := s[i] - '0'
+		if digit > 9 {
+			return 0, false
+		}
+		x = x*10 + int(digit)
+		if x < 0 {
+			return 0, false // underflow
+		}
+	}
+	return x, true
 }
 
 // https://github.com/google/starlark-go/blob/master/doc/spec.md#string·index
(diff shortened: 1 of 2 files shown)


starlark/library.go
```diff
@@ -1780,6 +1780,18 @@ func string_format(fnname string, recv_ Value, args Tuple, kwargs []Tuple) (Valu
...
 }
 
+func decimal(s string) (x int, ok bool) {
+	n := len(s)
+	for i := 0; i < n; i++ {
+		digit := s[i] - '0'
+		if digit > 9 {
+			return 0, false // not in [0-9]
+		}
+		x = x*10 + int(digit)
```
josharian:
```go
if x < 0 {
  return 0, false // overflow
}
```

starlark/library.go
```diff
@@ -1780,6 +1780,18 @@ func string_format(fnname string, recv_ Value, args Tuple, kwargs []Tuple) (Valu
...
+	}
+	return x, 0 < n && n < 10
```
josharian:
```go
return x, true
```

starlark/testdata/string.star
```diff
@@ -174,6 +174,8 @@ assert.eq("a{x}b{y}c{}".format(1, x=2, y=3), "a2b3c1")
 assert.fails(lambda: "a{z}b".format(x=1), "keyword z not found")
 assert.fails(lambda: "{-1}".format(1), "keyword -1 not found")
 assert.fails(lambda: "{-0}".format(1), "keyword -0 not found")
+assert.fails(lambda: "{+0}".format(1), "keyword \\+0 not found")
+assert.fails(lambda: "{+1}".format(1), "keyword \\+1 not found") # starlark-go/issues/114
```
josharian:
```python
assert.eq("{0000000000001}".format(0,1), "1")
```

(This matches Python 2/3 behavior, and makes explicit to future readers of the code that this behavior is intended.)

starlark/library.go
alandonovan: The final condition (0 < n && n < 10) is false if there was any possibility of underflow.

alandonovan: Done.

starlark/library.go
alandonovan: Done.

starlark/testdata/string.star
alandonovan: Good point. This much better than a length check.

alandonovan: I added this one too:
assert.eq("{012}".format(*range(100)), "12") # decimal, despite leading zeros

PR #122, opened 2019-01-23
resolve: disallow augmented assignments at toplevel

adonovan:
...unless -globalreassign is set.

Formerly, the Go implementation permitted x+=y on the basis
that we couldn't statically tell whether x was a list (in which
case x+=y means x.extend(y), which is not a rebinding), or some
other type, in which case it is a forbidden rebinding.
This change makes it match the Java implementation.
See https://github.com/bazelbuild/starlark/issues/20#issuecomment-456647994

Most non-Bazel-like clients of Starlark set the -globalreassign flag.


alandonovan:
Done (as #125)

PR #123, opened 2019-01-23
resolve: enable bitwise (& | ~ ^ << >>) operators always

adonovan:
See https://github.com/bazelbuild/starlark/issues/20#issuecomment-456647994


adonovan:
@laurentlb

PR #132, opened 2019-02-01
Make default DataFile work with GOPATH with multiple entries.

vadimsht:
It's not a big deal if this pull request is rejected, since DataFile is overridable and I can override it in my projects instead (it uses multiple GOPATHs).

### DIFF
--- a/syntax/parse_test.go
+++ b/syntax/parse_test.go
@@ -8,9 +8,7 @@ import (
 	"bufio"
 	"bytes"
 	"fmt"
-	"go/build"
 	"io/ioutil"
-	"path/filepath"
 	"reflect"
 	"strings"
 	"testing"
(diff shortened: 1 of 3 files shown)


syntax/parse_test.go
```diff
@@ -481,14 +479,8 @@ File
...
 
-// dataFile is the same as starlarktest.DataFile.
-// We make a copy to avoid a dependency cycle.
-var dataFile = func(pkgdir, filename string) string {
-	return filepath.Join(build.Default.GOPATH, "src/go.starlark.net", pkgdir, filename)
-}
-
 func BenchmarkParse(b *testing.B) {
-	filename := dataFile("syntax", "testdata/scan.star")
+	filename := starlarktest.DataFile("syntax", "testdata/scan.star")
```
vadimsht: It is already used by this file in TestParseErrors above.

josharian:
I don't understand why this routine is necessary at all. By default, `go test` runs executables with the cwd set to its package path, so you should be able to simply load the relative path `testdata/something.star` and have it work.

alandonovan:
go is not the only build system.

vadimsht:
I can explain why I'm sending this: my project uses starlarktest package, including assert.star it has. Starlark-go code is "vendored" into parallel GOPATH (i.e. all my sources are in /.../go and all third party code, including starlark-go, are in automatically managed /.../go/.vendor). This results in `build.Default.GOPATH` being literally `/.../go;/.../go/.vendor` which confused `DataFile`.

mcuadros:
Since the introduction of the go modules, now this is even more complex since is really hard found the path where is the source code. I believe that the easiest way to solve this is embed the source code of `assert.star` in Go.

PR #138, opened 2019-02-05
resolve: report likely identifier misspelling

alandonovan:
Move spell.go to internal/spell package to enable re-use.

Unfortunately the resolver API provides no way for it to enumerate
predeclared and universe, so these candidates are missing.


alandonovan:
Ping. :)

alandonovan:
No worries---thanks for all the reviews!

PR #139, opened 2019-02-06
starlarkstruct: remove to_json/to_proto methods

alandonovan:
These methods are a (mis)feature of Bazel and don't belong here.

### DIFF
--- a/starlarkstruct/struct.go
+++ b/starlarkstruct/struct.go
@@ -5,9 +5,6 @@
 // Package starlarkstruct defines the Starlark types 'struct' and
 // 'module', both optional language extensions.
 //
-// TODO(adonovan): remove all the Bazelisms (e.g. to_json, to_proto), by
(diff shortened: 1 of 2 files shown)


pcj:
For my education (this is what I love best about open-source, learning from other people)...  Why the disdain for these methods?  Is it that they are format-specific and belong in a supporting library?  Is there another design reason?

alandonovan:
> Why the disdain for these methods? Is it that they are format-specific and belong in a supporting library?

Several reasons:
1) they create ambiguities in the struct name space, because you can't rely on user-chosen field names being distinct from built-in methods.
2) JSON and especially proto are dependencies that deserve packages of their own; they should not come for free when you use struct.
3) JSON and proto are capable of handling all kinds of values, but the current methods make them available only to struct values.
4) The conversion from structs to json/proto is poorly documented.

PR #141, opened 2019-02-06
starlark: make built-ins require True/False for bool params

alandonovan:
Previously they would accept any value and use it truth-value.
Python permits any bool or int, more by historical accident than design.

This change may cause existing programs to have dynamic errors!

Updates bazelbuild/starlark#30

### DIFF
--- a/doc/spec.md
+++ b/doc/spec.md
@@ -1066,6 +1066,12 @@ in the environment of a specific module.
 Except where noted, built-in functions accept only positional arguments.
 The parameter names serve merely as documentation.
 
+Most built-in functions that have a Boolean parameter require its
+argument to be `True` or `False`. Unlike `if` statements, other values
+are not implicitly converted to their truth value and instead cause a
+dynamic error.
+
+
 ## Name binding and variables
 
 After a Starlark file is parsed, but before its execution begins, the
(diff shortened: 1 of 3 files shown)


josharian:
The changes look fine, but let's wait for the conversation at https://github.com/bazelbuild/starlark/issues/30 to reach a conclusion.

alandonovan:
Removed support for 0/1: the motivating example (Bazel's package function) was an outlier.

PTAL
@laurentlb

doc/spec.md
```diff
@@ -2846,12 +2852,12 @@ application-specific dialect) without breaking existing programs.
...
 ### any
 
-`any(x)` returns `True` if any element of the iterable sequence x is true.
+`any(x)` returns `True` if any element of the iterable sequence x has a truth value of true.
 If the iterable is empty, it returns `False`.
 
 ### all
 
-`all(x)` returns `False` if any element of the iterable sequence x is false.
+`all(x)` returns `False` if any element of the iterable sequence x has a truth-value of false.
```
josharian: "truth value" above, "truth-value" here. Either is fine, but pick one. :)

starlark/library.go
```diff
@@ -166,7 +166,8 @@ func builtinAttrNames(methods map[string]builtinMethod) []string {
 // If the variable is a bool, int, string, *List, *Dict, Callable,
 // Iterable, or user-defined implementation of Value,
 // UnpackArgs performs the appropriate type check.
-// (An int uses the AsInt32 check.)
+// A bool permits True, False, 0, or 1.
```
josharian: Update.

doc/spec.md
alandonovan: Done, in https://github.com/google/starlark-go/pull/142.

starlark/library.go
alandonovan: ditto

PR #142, opened 2019-02-07
belatedly apply review suggestions from #141

alandonovan:


josharian:
Thanks. I missed you by like 30 seconds. :)

PR #149, opened 2019-02-15
Support required keyword-only parameters

adonovan:
In def f(a, b=1, *, d, e=2, f), the parameters d and f
are keyword-only (because they come after the * or *args parameter),
but have no default value, so they are mandatory.

Also, improve the error messages.

Updates bazelbuild/starlark#23

### DIFF
--- a/internal/compile/compile.go
+++ b/internal/compile/compile.go
@@ -37,7 +37,7 @@ import (
 const debug = false // TODO(adonovan): use a bitmap of options; and regexp to match files
 
 // Increment this to force recompilation of saved bytecode files.
-const Version = 6
+const Version = 7
 
 type Opcode uint8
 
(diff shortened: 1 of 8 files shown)


alandonovan:
@laurentlb

internal/compile/compile.go
```diff
@@ -113,7 +114,7 @@ const (
 	CONSTANT    //                - CONSTANT<constant>  value
 	MAKETUPLE   //        x1 ... xn MAKETUPLE<n>        tuple
 	MAKELIST    //        x1 ... xn MAKELIST<n>         list
-	MAKEFUNC    //      args kwargs MAKEFUNC<func>      fn
+	MAKEFUNC    //  deflts freevars MAKEFUNC<func>      fn
```
jayconrod: Change to `defaults`. `deflts` is a little too terse.

jayconrod:
A note on the CLA thing: when you squash-and-merge a PR, GitHub sets the Author and Committer fields of the squashed commit to the PR owner and the PR merger, respectively. So currently, at the tip of `master`:

```
go.starlark.net $ git show --pretty=full HEAD | head
commit 52153852d546514d9bf61e444fe5d829f3835476
Author: Alan Donovan <alan@alandonovan.net>
Commit: alandonovan <adonovan@google.com>

    Support keyword-only function parameters (#143)
```

This probably isn't important, but I think it's worth pointing out that the difference is observable after merge.

internal/compile/compile.go
alandonovan: Done.

alandonovan:
Branching to #151 (using google.com account) for submit.

PR #151, opened 2019-02-15
Support required keyword-only parameters

alandonovan:
In def f(a, b=1, *, d, e=2, f), the parameters d and f
are keyword-only (because they come after the * or *args parameter),
but have no default value, so they are mandatory.

Also:
- improve the error messages
- spec update
- avoid need for "defined" bitset in setArgs

Updates bazelbuild/starlark#23


alandonovan:
Reviewed as #149  (wrong GitHub account)

PR #155, opened 2019-02-17
Small int as int64

afking:
Represents small ints as int64 values promoting to big.Int types when needed. Have added some simple benchmarking. Based off of https://github.com/robpike/ivy int/bigint types.

```
Benchmark/bench_bigint-8                   10000            187182 ns/op
Benchmark/bench_int-8                      10000            180208 ns/op

Benchmark/bench_bigint-8                    5000            261473 ns/op
Benchmark/bench_int-8                      10000            128741 ns/op
```

### DIFF
--- a/starlark/int.go
+++ b/starlark/int.go
@@ -8,36 +8,44 @@ import (
 	"fmt"
 	"math"
 	"math/big"
+	"strconv"
 
 	"go.starlark.net/syntax"
 )
 
 // Int is the type of a Starlark int.
-type Int struct{ bigint *big.Int }
+type Int struct {
+	// big != nil <=> value is not representable as int32.
+	// We use only the signed 32 bit range of small to ensure
+	// that small+small and small*small do not overflow.
+
+	small int64    // minint32 <= small <= maxint32
+	big   *big.Int // big.BitLen() >= 32
+}
+
+// newBig allocates a new big.Int.
+func newBig(x int64) *big.Int {
+	if 0 <= x && int64(big.Word(x)) == x {
+		// x is guaranteed to fit into a single big.Word.
+		// Most starlark ints are small,
+		// but math/big assumes that since you've chosen to use math/big,
+		// your big.Ints will probably grow, so it over-allocates.
+		// Avoid that over-allocation by manually constructing a single-word slice.
+		// See https://golang.org/cl/150999, which will hopefully land in Go 1.13.
+		return new(big.Int).SetBits([]big.Word{big.Word(x)})
+	}
+	return big.NewInt(x)
+}
 
 // MakeInt returns a Starlark int for the specified signed integer.
 func MakeInt(x int) Int { return MakeInt64(int64(x)) }
 
 // MakeInt64 returns a Starlark int for the specified int64.
 func MakeInt64(x int64) Int {
-	if 0 <= x {
-		if x < int64(len(smallint)) {
-			if !smallintok {
-				panic("MakeInt64 used before initialization")
-			}
-			return Int{&smallint[x]}
-		}
-		if int64(big.Word(x)) == x {
-			// x is guaranteed to fit into a single big.Word.
-			// Most starlark ints are small,
-			// but math/big assumes that since you've chosen to use math/big,
-			// your big.Ints will probably grow, so it over-allocates.
-			// Avoid that over-allocation by manually constructing a single-word slice.
-			// See https://golang.org/cl/150999, which will hopefully land in Go 1.13.
-			return Int{new(big.Int).SetBits([]big.Word{big.Word(x)})}
-		}
+	if math.MinInt32 <= x && x <= math.MaxInt32 {
+		return Int{small: x}
 	}
-	return Int{big.NewInt(x)}
+	return Int{big: newBig(x)}
 }
 
 // MakeUint returns a Starlark int for the specified unsigned integer.
@@ -45,36 +53,31 @@ func MakeUint(x uint) Int { return MakeUint64(uint64(x)) }
 
 // MakeUint64 returns a Starlark int for the specified uint64.
 func MakeUint64(x uint64) Int {
-	if x < uint64(len(smallint)) {
-		if !smallintok {
-			panic("MakeUint64 used before initialization")
-		}
-		return Int{&smallint[x]}
+	if x <= math.MaxInt32 {
+		return Int{small: int64(x)}
 	}
 	if uint64(big.Word(x)) == x {
-		// See comment in MakeInt64 for an explanation of this optimization.
-		return Int{new(big.Int).SetBits([]big.Word{big.Word(x)})}
+		// See comment in newBig for an explanation of this optimization.
+		return Int{big: new(big.Int).SetBits([]big.Word{big.Word(x)})}
 	}
-	return Int{new(big.Int).SetUint64(x)}
+	return Int{big: new(big.Int).SetUint64(x)}
+}
+
+// MakeBigInt returns a Starlark int for the specified big.Int.
+// The caller must not subsequently modify x.
+func MakeBigInt(x *big.Int) Int {
+	if n := x.BitLen(); n < 32 || n == 32 && x.Int64() == math.MinInt32 {
+		return Int{small: x.Int64()}
+	}
+	return Int{big: x}
 }
 
 var (
-	smallint   [256]big.Int
-	smallintok bool
-	zero, one  Int
+	zero, one = Int{small: 0}, Int{small: 1}
+	oneBig    = newBig(1)
 
 	_ HasUnary = Int{}
 )
-
-func init() {
-	for i := range smallint {
-		smallint[i].SetInt64(int64(i))
-	}
-	smallintok = true
-
-	zero = MakeInt64(0)
-	one = MakeInt64(1)
-}
 
 // Unary implements the operations +int, -int, and ~int.
 func (i Int) Unary(op syntax.Token) (Value, error) {
@@ -92,21 +95,39 @@ func (i Int) Unary(op syntax.Token) (Value, error) {
 // Int64 returns the value as an int64.
 // If it is not exactly representable the result is undefined and ok is false.
 func (i Int) Int64() (_ int64, ok bool) {
-	x, acc := bigintToInt64(i.bigint)
-	if acc != big.Exact {
-		return // inexact
+	if i.big != nil {
+		x, acc := bigintToInt64(i.big)
+		if acc != big.Exact {
+			return // inexact
+		}
+		return x, true
 	}
-	return x, true
+	return i.small, true
+}
+
+// BigInt returns the value as a big.Int.
+// The returned variable must not be modified by the client.
+func (i Int) BigInt() *big.Int {
+	if i.big != nil {
+		return i.big
+	}
+	return newBig(i.small)
 }
 
 // Uint64 returns the value as a uint64.
 // If it is not exactly representable the result is undefined and ok is false.
 func (i Int) Uint64() (_ uint64, ok bool) {
-	x, acc := bigintToUint64(i.bigint)
-	if acc != big.Exact {
+	if i.big != nil {
+		x, acc := bigintToUint64(i.big)
+		if acc != big.Exact {
+			return // inexact
+		}
+		return x, true
+	}
+	if i.small < 0 {
 		return // inexact
 	}
-	return x, true
+	return uint64(i.small), true
 }
 
 // The math/big API should provide this function.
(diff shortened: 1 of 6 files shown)


starlark/int.go
```diff
@@ -142,75 +138,161 @@ var (
...
 	return 12582917 * uint32(lo+3), nil
 }
-func (x Int) CompareSameType(op syntax.Token, y Value, depth int) (bool, error) {
-	return threeway(op, x.bigint.Cmp(y.(Int).bigint)), nil
+func (x Int) CompareSameType(op syntax.Token, v Value, depth int) (bool, error) {
+	y := v.(Int)
+	if x.bigint != nil && y.bigint != nil {
+		return threeway(op, x.big().Cmp(y.big())), nil
+	}
+	return threeway(op, int(x.val-y.val)), nil
```
alandonovan: This looks wrong. If one but not both are big, then one of the operands of the subtraction is undefined.

starlark/int.go
```diff
@@ -142,75 +138,161 @@ var (
...
-func (x Int) Sub(y Int) Int  { return Int{new(big.Int).Sub(x.bigint, y.bigint)} }
-func (x Int) Mul(y Int) Int  { return Int{new(big.Int).Mul(x.bigint, y.bigint)} }
-func (x Int) Or(y Int) Int   { return Int{new(big.Int).Or(x.bigint, y.bigint)} }
-func (x Int) And(y Int) Int  { return Int{new(big.Int).And(x.bigint, y.bigint)} }
-func (x Int) Xor(y Int) Int  { return Int{new(big.Int).Xor(x.bigint, y.bigint)} }
-func (x Int) Not() Int       { return Int{new(big.Int).Not(x.bigint)} }
-func (x Int) Lsh(y uint) Int { return Int{new(big.Int).Lsh(x.bigint, y)} }
-func (x Int) Rsh(y uint) Int { return Int{new(big.Int).Rsh(x.bigint, y)} }
+	if i.bigint != nil {
+		// TODO(adonovan): opt: handle common values without allocation.
```
alandonovan: You can remove this TODO, since you've done it.

starlark/int.go
```diff
@@ -142,75 +138,161 @@ var (
...
+		f, _ := new(big.Float).SetInt(i.bigint).Float64()
+		return Float(f)
+	}
+	return Float(float64(i.val))
```
alandonovan: return Float(i.val)

starlark/int.go
```diff
@@ -142,75 +138,161 @@ var (
...
+}
+
+func (x Int) Sign() int {
+	switch {
+	case x.bigint != nil:
+		return x.bigint.Sign()
+	case x.val > 0:
```
alandonovan: This version is less branchy:

if x.bigint != nil {
   return x.bigint.Sign()
}
return signum64(x.val)

(See existing signum function.)

starlark/int.go
```diff
@@ -142,75 +138,161 @@ var (
...
+	return strconv.FormatInt(i.val, base)
+}
+func (i Int) String() string { return i.Text(10) }
 func (i Int) Type() string   { return "int" }
 func (i Int) Freeze()        {} // immutable
 func (i Int) Truth() Bool    { return i.Sign() != 0 }
 func (i Int) Hash() (uint32, error) {
 	var lo big.Word
-	if i.bigint.Sign() != 0 {
+	if i.bigint != nil && i.bigint.Sign() != 0 {
```
alandonovan: if bigint != nil, sign cannot be zero.

starlark/int.go
```diff
@@ -142,75 +138,161 @@ var (
...
+}
+func (x Int) Mul(y Int) Int {
+	if x.bigint != nil || y.bigint != nil {
+		return Int{bigint: new(big.Int).Mul(x.big(), y.big())}.maybeSmall()
+	}
+	return Int{val: x.val * y.val}.maybeBig()
+}
+func (x Int) Or(y Int) Int {
+	if x.bigint != nil || y.bigint != nil {
+		return Int{bigint: new(big.Int).Or(x.big(), y.big())}
```
alandonovan: Can you prove that x|y never makes a small from two bigs?

starlark/int.go
```diff
@@ -142,75 +138,161 @@ var (
...
+	}
+	return Int{val: x.val | y.val}
+}
+func (x Int) And(y Int) Int {
+	if x.bigint != nil || y.bigint != nil {
+		return Int{bigint: new(big.Int).And(x.big(), y.big())}
```
alandonovan: This looks incorrect. x&y can definitely make a small from two bigs (or one of each).
 1 << 100 & 0 == 0.

starlark/int.go
```diff
@@ -142,75 +138,161 @@ var (
...
+	}
+	return Int{val: x.val & y.val}
+}
+func (x Int) Xor(y Int) Int {
+	if x.bigint != nil || y.bigint != nil {
+		return Int{bigint: new(big.Int).Xor(x.big(), y.big())}
```
alandonovan: Same here. x^x == 0 for any x.

starlark/int.go
```diff
@@ -142,75 +138,161 @@ var (
...
+	}
+	return Int{val: x.val ^ y.val}
+}
+func (x Int) Not() Int {
+	if x.bigint != nil {
+		return Int{bigint: new(big.Int).Not(x.bigint)}
```
alandonovan: Same here, I think.

starlark/int.go
```diff
@@ -142,75 +138,161 @@ var (
...
-	if (x.bigint.Sign() < 0) != (y.bigint.Sign() < 0) && rem.Sign() != 0 {
-		rem.Add(&rem, y.bigint)
+	quo.QuoRem(xb, yb, &rem)
+	if (xb.Sign() < 0) != (yb.Sign() < 0) && rem.Sign() != 0 {
+		rem.Add(&rem, yb)
+	}
+	return Int{bigint: &rem}.maybeSmall()
+}
+
+func (i Int) rational() *big.Rat {
```
alandonovan: Move rational nearer where it is needed.

starlark/int.go
```diff
@@ -142,75 +138,161 @@ var (
...
 	i, ok := x.(Int)
 	if !ok {
 		return 0, fmt.Errorf("got %s, want int", x.Type())
 	}
-	if i.bigint.BitLen() <= 32 {
-		v := i.bigint.Int64()
-		if v >= math.MinInt32 && v <= math.MaxInt32 {
-			return int(v), nil
+	if i.bigint != nil {
+		if i.bigint.BitLen() <= 32 {
```
alandonovan: Isn't this is always false?

starlark/testdata/benchmark.star
```diff
@@ -20,3 +20,14 @@ range1000 = range(1000)
 def bench_builtin_method():
   for _ in range1000:
     emptydict.get(None)
+
+def bench_int():
+  a = 0
+  for _ in range1000:
+    a += 1
+
+maxint64 = 9223372036854775807 # = 2^63
```
alandonovan: maxint64 = 1 << 63

starlark/int.go
```diff
@@ -142,75 +138,161 @@ var (
...
+}
+
+func (i Int) rational() *big.Rat {
+	if i.bigint != nil {
+		return new(big.Rat).SetInt(i.bigint)
+	}
+	return new(big.Rat).SetInt64(i.val)
+}
+
+func (i Int) maybeBig() Int {
```
alandonovan: I think maybeBig wants to be a function from int64 to Int, rather than a function that takes an Int whose representation invariant is violated and returns a sound Int. Similarly maybeSmall should be a function from *big.Int to Int. In other words, make this the implementation of MakeInt64 and call that function instead of maybeBig. Similarly, add a new public function MakeBigInt, and call that instead of maybeSmall.

afking:
@alandonovan thanks! I've fixed up some of the issues but will add some more test cases.

starlark/int.go
```diff
@@ -8,74 +8,61 @@ import (
...
 	"math"
 	"math/big"
+	"strconv"
 
 	"go.starlark.net/syntax"
 )
 
 // Int is the type of a Starlark int.
-type Int struct{ bigint *big.Int }
+type Int struct {
```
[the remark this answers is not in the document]
alandonovan: Please use this wording:

type Int struct {
// big != nil <=> value is not representable as int32.
// We use only the signed 32 bit range of small to ensure
//  that small+small and small*small do not overflow.

small int64 // minint32 <= small <= maxint32
big *big.Int // big.BitLen() > 32
}

starlark/int.go
alandonovan: And since we're adding a constructor from big.Int, we should have an accessor big values too. Let's rename the toBig method (again) to BigInt for symmetry.

We should also document at the MakeBigInt and BigInt methods that the client most not modify the provided/returned big.Int variable.

afking: Added the MakeBigInt, BigInt methods and move toBig to newBig.

starlark/bench_test.go
```diff
@@ -11,11 +11,15 @@ import (
 	"strings"
 	"testing"
 
+	"go.starlark.net/resolve"
 	"go.starlark.net/starlark"
 	"go.starlark.net/starlarktest"
 )
 
 func Benchmark(b *testing.B) {
+	resolve.AllowNestedDef = true
```
alandonovan: The problem this solves was since fixed.

starlark/int.go
```diff
@@ -8,74 +8,66 @@ import (
...
-			// but math/big assumes that since you've chosen to use math/big,
-			// your big.Ints will probably grow, so it over-allocates.
-			// Avoid that over-allocation by manually constructing a single-word slice.
-			// See https://golang.org/cl/150999, which will hopefully land in Go 1.13.
-			return Int{new(big.Int).SetBits([]big.Word{big.Word(x)})}
-		}
-	}
-	return Int{big.NewInt(x)}
-}
+func MakeInt64(x int64) Int { return maybeBig(x) }
```
alandonovan: Let's inline maybeBig here, and call it as MakeInt64.

starlark/int.go
```diff
@@ -8,74 +8,66 @@ import (
...
-		return Int{new(big.Int).SetBits([]big.Word{big.Word(x)})}
+		// See comment in newBig for an explanation of this optimization.
+		return Int{big: new(big.Int).SetBits([]big.Word{big.Word(x)})}
 	}
-	return Int{new(big.Int).SetUint64(x)}
+	return Int{big: new(big.Int).SetUint64(x)}
 }
 
+// MakeBigInt returns a Starlark int for the specified big.Int.
+func MakeBigInt(x *big.Int) Int { return maybeSmall(x) }
```
alandonovan: And inline maybeSmall here.

starlark/int.go
```diff
@@ -92,21 +84,39 @@ func (i Int) Unary(op syntax.Token) (Value, error) {
...
+		if acc != big.Exact {
+			return // inexact
+		}
+		return x, true
+	}
+	return i.small, true
+}
+
+// BigInt returns the value as a big.Int.
+// The provided value is immutable and must not be modified by the client.
```
alandonovan: "The returned variable must not ..."

(we only hope it is immutable)

starlark/int.go
```diff
@@ -142,75 +152,166 @@ var (
...
+
+		var quo, rem big.Int
+		quo.QuoRem(xb, yb, &rem)
+		if (xb.Sign() < 0) != (yb.Sign() < 0) && rem.Sign() != 0 {
+			rem.Add(&rem, yb)
+		}
+		return maybeSmall(&rem)
+	}
+	rem := x.small % y.small
+	if (signum64(x.small) < 0) != (signum64(y.small) < 0) && signum64(rem) != 0 {
```
alandonovan: redundant parens, x 2

starlark/int.go
```diff
@@ -142,75 +152,166 @@ var (
...
+		rem += y.small
+	}
+	return Int{small: rem}
+}
+
+func maybeSmall(x *big.Int) Int {
+	if x.BitLen() < 32 {
```
alandonovan: <= 32

starlark/int.go
```diff
@@ -142,75 +152,166 @@ var (
...
+func maybeBig(x int64) Int {
+	if math.MinInt32 <= x && x <= math.MaxInt32 {
+		return Int{small: x}
 	}
-	return Int{&rem}
+	return Int{big: newBig(x)}
 }
 
-func (i Int) rational() *big.Rat { return new(big.Rat).SetInt(i.bigint) }
+func (i Int) rational() *big.Rat {
```
alandonovan: Let's inline this; it's called only once. Then we can eliminate the following nil check.

starlark/int_test.go
```diff
@@ -0,0 +1,54 @@
...
+// license that can be found in the LICENSE file.
+
+package starlark
+
+import (
+	"math"
+	"testing"
+)
+
+func TestIntOpts(t *testing.T) {
```
alandonovan: // TestIntOpts exercises integer arithmetic, especially at the boundaries.

starlark/int_test.go
```diff
@@ -0,0 +1,54 @@
...
+	f := maybeBig
+	for i, test := range []Int{
```
alandonovan: It would be nice to show (and assert) the expected result as a string:

{f(math.MaxInt32).Add(f(1)), "0x100000000"},
etc

starlark/int_test.go
```diff
@@ -0,0 +1,54 @@
...
+		// Mod
+		f(1 << 32).Mod(f(2)),
+		// Or
+		f(1 << 32).Or(f(0)),
+		f(math.MaxInt32).Or(f(0)),
+		f(math.MinInt32).Or(f(math.MinInt32)),
+		// Xor
+		f(math.MinInt32).Xor(f(-1)),
+		// Not
+		f(math.MinInt32).Not(),
```
alandonovan: I'm a bit surprised about this one. ~ -0x80000000 = 0x7fffffff. Both are int32 bits.

Perhaps this is a consequence of using < not <= (see above).

starlark/int_test.go
```diff
@@ -0,0 +1,54 @@
...
+		f(math.MaxInt32).Not(),
+		// Shift
+		f(1).Lsh(31), f(1).Lsh(32),
+		f(math.MaxInt32 + 1).Rsh(1),
+		f(math.MinInt32 * 2).Rsh(1), // Big math.MinInt32
+	} {
+		if test.big != nil && test.big.BitLen() < 32 {
```
alandonovan: <= again

starlark/int_test.go
```diff
@@ -0,0 +1,54 @@
...
+			t.Errorf("expected small, %d %s", i, test)
+			continue
```
alandonovan: if test.big != nil {
   if bitlen <= 32  { error  }
   if small != 0 { error }
   continue
}

starlark/int.go
```diff
@@ -8,74 +8,66 @@ import (
...
-		// See comment in MakeInt64 for an explanation of this optimization.
-		return Int{new(big.Int).SetBits([]big.Word{big.Word(x)})}
+		// See comment in newBig for an explanation of this optimization.
+		return Int{big: new(big.Int).SetBits([]big.Word{big.Word(x)})}
 	}
-	return Int{new(big.Int).SetUint64(x)}
+	return Int{big: new(big.Int).SetUint64(x)}
 }
 
+// MakeBigInt returns a Starlark int for the specified big.Int.
```
alandonovan: // The caller must not subsequently modify x.

starlark/int.go
afking: This has to be `<` to avoid overflow. `big.NewInt(math.MaxUint32).BitLen() == 32` which overflows int64 on multiplication.

starlark/int.go
afking: I think these are required, might have misunderstood.

starlark/int.go
afking: Its called twice here: https://github.com/google/starlark-go/blob/master/starlark/value.go#L1160 and https://github.com/google/starlark-go/blob/master/starlark/value.go#L1175

afking: Do floating comparisons need to be done with math big?

starlark/int_test.go
afking: I exposed big.Ints Format on Int and added %x printing, is there a better way to do that?

afking:
Thanks for taking the time for reviews. The motivation was for a use case of starlark as a user provided event handler of which our most common test case is counting.

starlark/int.go
```diff
@@ -8,74 +8,77 @@ import (
...
 
 // Int is the type of a Starlark int.
-type Int struct{ bigint *big.Int }
+type Int struct {
+	// big != nil <=> value is not representable as int32.
+	// We use only the signed 32 bit range of small to ensure
+	// that small+small and small*small do not overflow.
+
+	small int64    // minint32 <= small <= maxint32
+	big   *big.Int // big.BitLen() > 32
```
alandonovan: You're right that > 32 isn't correct, and the BitLen alone is not the right predicate to express the boundary between big and small.  Let's update it to:

// big.BitLen() >= 32 (note: MinInt32 is sole small value of BitLen 32)

Sorry for the bad directions.

starlark/int.go
```diff
@@ -8,74 +8,77 @@ import (
...
+		return Int{big: new(big.Int).SetBits([]big.Word{big.Word(x)})}
 	}
-	return Int{new(big.Int).SetUint64(x)}
+	return Int{big: new(big.Int).SetUint64(x)}
+}
+
+// MakeBigInt returns a Starlark int for the specified big.Int.
+// The caller must not subsequently modify x.
+func MakeBigInt(x *big.Int) Int {
+	if x.BitLen() < 32 {
```
alandonovan: This will have to be updated to:

n := x.BitLen()
if n < 32 || n == 32 && x.Int64() == math.MinInt32 {

starlark/int_test.go
```diff
@@ -0,0 +1,71 @@
...
+		{f(1).Lsh(32), "100000000"},
+		{f(math.MaxInt32 + 1).Rsh(1), "40000000"},
+		{f(math.MinInt32 * 2).Rsh(1), "-80000000"}, // Big math.MinInt32
+	} {
+		if got := fmt.Sprintf("%x", test.val); got != test.want {
+			t.Errorf("%d equals %s, want %s", i, got, test.want)
+		}
+
+		if test.val.big != nil {
+			if test.val.big.BitLen() < 32 {
```
alandonovan: n := test.val.big.BitLen()
if n < 32 || n == 32 && x.Int64() == math.MinInt32 {

Although at that point the test is testing the production code by duplicating it.

starlark/int.go
```diff
@@ -142,75 +163,158 @@ var (
...
+		return i.big.Text(base)
+	}
+	return strconv.FormatInt(i.small, base)
+}
+func (i Int) Format(s fmt.State, ch rune) {
+	if i.big != nil {
+		i.big.Format(s, ch)
+		return
+	}
+	newBig(i.small).Format(s, ch)
```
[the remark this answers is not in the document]
afking: I couldn't find one that was nicer then rebuilding the format string and passing it back to `fmt.Fprintf`.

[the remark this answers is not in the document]
alandonovan: That's all I could see too. Let's keep it as you have it.

starlark/int.go
afking: No problem, fixed the minInt32 edge case.

starlark/int.go
```diff
@@ -142,75 +163,158 @@ var (
 	maxint64 = new(big.Int).SetInt64(math.MaxInt64)
 )
 
-func (i Int) String() string { return i.bigint.String() }
+func (i Int) Text(base int) string {
```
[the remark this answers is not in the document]
afking: Moved Text back to String() and use fmt for formatting.

starlark/int_test.go
afking: I moved the check to a bigint range check.

alandonovan:
LGTM. Thanks again.

PR #161, opened 2019-02-24
Two comment tweaks

alandonovan:


josharian:
Off by one: three comment tweaks. :)

I wonder whether it’d be easier to understand if instead of talking of BitLen and minint32, you just refer to the range of Go’s built-in int32.

alandonovan:
Alright, PTAL.

Issue #170, opened 2019-03-07
starlark: dynamic error executing definition of a recursive nested def 


alandonovan:
Declaring a recursive nested function causes a dynamic error as execution of the def statement, while attempting to build its closure, attempts to load the value of the not-yet-defined local variable that will contain the function.

```
$ cat a.star
def a():
    def b():
        print(b)

print(a())
$ go run ./cmd/starlark/ -nesteddef a.star
Traceback (most recent call last):
  a.star:5: in <toplevel>
  a.star:2: in a
Error: local variable b referenced before assignment
```
How did I not notice this already? It's the classic 'letrec' problem. It's as if I never went to graduate school. :(

Recursion is not essential; the problem arises any time the outer local changes after the closure is created:
```
$ cat d.star 
def a():
    x = 1
    def b(): print(x)
    b()
    x = 2
    b()

a()

$ starlark -nesteddef d.star 
1
1
$ python  d.star 
1
2
```

The solution is, during resolution, to mark local variables such as x that are shared by more than one function (a and b), as "cell" variables, and allocate each one separately at function start. Parameters that are so marked must be spilled into cells in the function prologue. Cells must be accesses and updated using a different indirect opcode to regular locals. A Function's freevars tuple will consist only of references to a subset of its ancestor's cells. That means cells must be first-class Starlark values.

PR #171, opened 2019-03-08
syntax: expose Binding concept

alandonovan:
This refactoring exposes the concept of Binding, which was previously
internal to the resolver. Previously the Scope and Index fields of the
binding were copied into each Ident; now the relationship is indirect.

A Binding is the entity created by the first binding occurrence of a
name, and the binding ties together all uses of that name.

This change is a preparation for fixing issue #170, which requires
that we mark the bindings that need to be allocated indirectly in
closure cells. Currently, in the absence of an indirect relationship
there is no way to mark a binding so that all Idents are affected.

This change leads to some a minor simplification in lookupLexical,
which used to synthesize a new Ident, when logically it is creating
only a new binding.

Unfortunately, Binding and Scope must be declared in the syntax
package even though it logically belongs to the resolver,
because the (resolved) syntax trees refer to it.

(If you're familiar with the go/types package, syntax.Binding
corresponds to types.Var. go/types solves its analogous dependency
problem by putting resolver-derived facts in an external map, not the
syntax tree, which is expensive. So perhaps it's more accurate to
say Binding corresponds to the deprecated ast.Object.)

Change-Id: I5976a638fd62a249a485b37aec6a476650fe32b3

### DIFF
--- a/syntax/syntax.go
+++ b/syntax/syntax.go
@@ -71,8 +71,8 @@ type File struct {
 	Stmts []Stmt
 
 	// set by resolver:
-	Locals  []*Ident // this file's (comprehension-)local variables
-	Globals []*Ident // this file's global variables
+	Locals  []*Binding // this file's (comprehension-)local variables
+	Globals []*Binding // this file's global variables
 }
 
 func (x *File) Span() (start, end Position) {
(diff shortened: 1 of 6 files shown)


alandonovan:
FYI: the follow-up change to fix the bug is https://github.com/google/starlark-go/tree/letrec-170

syntax/syntax.go
```diff
@@ -6,6 +6,7 @@
 package syntax // import "go.starlark.net/syntax"
 
 // A Node is a node in a Starlark syntax tree.
+
```
jayconrod: Nit: why the blank line? This should probably still be a doc comment.

alandonovan: Why? Because of my inept fumbling with the keyboard. Fixed!

PR #172, opened 2019-03-08
starlark: capture free variables by reference

### LINKED ISSUE #170: starlark: dynamic error executing definition of a recursive nested def

alandonovan:
This change causes closures for nested functions to capture their
enclosing functions' variables by reference. Even though an inner
function cannot update outer variables, it must observe updates to
them made by the outer function.

A special case of this is a nested recursive function f:

   def outer():
       def f(): ...f()...

The def f statement constructs a closure which captures f, and then
binds the closure value to f. If the closure captures by value (as
before this change), the def statement will fail because f is
undefined (see issue #170). Now, the closure captures a reference
to f, so it is safe to execute before f has been assigned.

This is implemented as follows. During resolving, captured local
variables such as f are marked as as "cells". The compiler assumes and
guarantees that such locals are values of a special internal type
called 'cell', and it emits explicit instructions to load from and
store into the cell. At runtime, cells are created on entry to the function;
parameters may be "spilled" into cells as needed.
Each cell variable gets its own allocation to avoid spurious liveness.
A function's tuple of free variables contains only cells.

Fixes #170

### DIFF
--- a/starlark/interp.go
+++ b/starlark/interp.go
@@ -52,6 +52,12 @@ func (fn *Function) CallInternal(thread *Thread, args Tuple, kwargs []Tuple) (Va
 		defer fmt.Println("Leaving ", f.Name)
 	}
 
+	// Spill indicated locals to cells.
+	// Each cell is a separate alloc to avoid spurious liveness.
+	for _, index := range f.Cells {
+		locals[index] = &cell{locals[index]}
+	}
+
 	// TODO(adonovan): add static check that beneath this point
 	// - there is exactly one return statement
 	// - there is no redefinition of 'err'.
@@ -502,6 +508,12 @@ loop:
 			locals[arg] = stack[sp-1]
 			sp--
 
+		case compile.SETCELL:
+			x := stack[sp-2]
+			y := stack[sp-1]
+			sp -= 2
+			y.(*cell).v = x
+
 		case compile.SETGLOBAL:
 			fn.globals[arg] = stack[sp-1]
 			sp--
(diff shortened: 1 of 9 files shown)


starlark/interp.go
```diff
@@ -52,6 +52,12 @@ func (fn *Function) CallInternal(thread *Thread, args Tuple, kwargs []Tuple) (Va
 		defer fmt.Println("Leaving ", f.Name)
 	}
 
+	// Spill indicated locals to cells.
+	// Each cell is a separate alloc to avoid spurious liveness.
```
jayconrod: Thanks for keeping these separate. I've seen a couple bad memory leaks in V8 that were the consequence of allocating an array for captured variables instead of individual cells.

PR #178, opened 2019-03-21
resolve: load statements now create local bindings

alandonovan:
Previously, load statements created global bindings.
As an often-unwanted consequence, names loaded by one module
would leak into its module dictionary and be visible to others.

This change to the resolver makes load define file-local variables
using a similar approach to Go: a new lexical block (the file block)
between function/comprehension blocks and the globals.
The set of names in the file block and the globals must be disjoint.

This incompatible change is guarded by the LoadBindsGlobally flag,
which is false by default. Users who want the old behavior may set
it to true, but they should prepare for it to go away entirely.
After that point, the old behavior may be simulated thus:
	- load("...", "x")
	+ load("...", _x="x"); x = _x # re-export

### DIFF
--- a/doc/spec.md
+++ b/doc/spec.md
@@ -1111,7 +1111,7 @@ h = [2*i for i in a]
 The environment of a Starlark program is structured as a tree of
 _lexical blocks_, each of which may contain name bindings.
 The tree of blocks is parallel to the syntax tree.
-Blocks are of four kinds.
+Blocks are of five kinds.
 
 <!-- Avoid the term "built-in" block since that's also a type. -->
 At the root of the tree is the _predeclared_ block,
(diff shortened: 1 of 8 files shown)


alandonovan:
@laurentlb -- this is the Go implementation of https://github.com/bazelbuild/starlark/issues/37

laurentlb:
Thank you

doc/spec.md
```diff
@@ -1126,17 +1126,24 @@ These additional functions may have side effects on the application.
...
-Bindings inside either of these kinds of block are called _local_.
+Nested beneath the module block is the _file_ block,
+which contains bindings local to the current file.
+Only `load` statements create bindings in this block.
+The names bound by the file block and by the module block are disjoint:
+it is an error for a load statement to bind the name of a global,
+or for a top-level statement to assign to a name bound by a load statement.
+
+A file block contains a _function_ block for each top-level
+function, and a _comprehension_ block for each top-level comprehension.
```
jayconrod: It would be helpful to mention the other variables in the example here which aren't explicitly called out elsewhere, especially `i`.

alandonovan: Good idea; done.

doc/spec.md
```diff
@@ -1126,17 +1126,24 @@ These additional functions may have side effects on the application.
 Starlark programs cannot change the set of predeclared bindings
 or assign new values to them.
 
-Nested beneath the predeclared block is the _module_ block, which
-contains the bindings of the current file.
+Nested beneath the predeclared block is the _module_ block,
+which contains the bindings of the current module.
 Bindings in the module block (such as `a`, `b`, `c`, and `h` in the
-example) are called _global_.
+example) are called _global_ and may be visible to other modules.
```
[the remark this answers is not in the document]
alandonovan: Well spotted. PTAL.

PR #181, opened 2019-04-01
starlark: a simple pprof-compatible execution-time profiler

alandonovan:
See profile.go for design.

### DIFF
--- a/starlark/profile.go
+++ b/starlark/profile.go
@@ -1,0 +1,444 @@
+// Copyright 2019 The Bazel Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package starlark
+
+// This file defines a simple execution-time profiler for Starlark.
+// It measures the wall time spent executing Starlark code, and emits a
+// gzipped protocol message in pprof format (github.com/google/pprof).
+//
+// When profiling is enabled, the interpreter calls the profiler to
+// indicate the start and end of each "span" or time interval. A leaf
+// function (whether Go or Starlark) has a single span. A function that
+// calls another function has spans for each interval in which it is the
+// top of the stack. (A LOAD instruction also ends a span.)
+//
+// At the start of a span, the interpreter records the current time in
+// the thread's topmost frame. At the end of the span, it obtains the
+// time again and subtracts the span start time. The difference is added
+// to an accumulator variable in the thread. If the accumulator exceeds
+// some fixed quantum (10ms, say), the profiler records the current call
+// stack and sends it to the profiler goroutine, along with the number
+// of quanta, which are subtracted. For example, if the accumulator
+// holds 3ms and then a completed span adds 25ms to it, its value is 28ms,
+// which exceeeds 10ms. The profiler records a stack with the value 20ms
+// (2 quanta), and the accumulator is left with 8ms.
+//
+// The profiler goroutine converts the stacks into the pprof format and
+// emits a gzip-compressed protocol message to the designated output
+// file. We use a hand-written streaming proto encoder to avoid
+// dependencies on pprof and proto, and to avoid the need to
+// materialize the profile data structure in memory.
+//
+// A limitation of this profiler is that it measures wall time, which
+// does not necessarily correspond to CPU time. A CPU profiler requires
+// that only running (not runnable) threads are sampled; this is
+// commonly achieved by having the kernel deliver a (PROF) signal to an
+// arbitrary running thread, through setitimer(2). The CPU profiler in the
+// Go runtime uses this mechanism, but it is not possible for a Go
+// application to register a SIGPROF handler, nor is it possible for a
+// Go handler for some other signal to read the stack pointer of
+// the interrupted thread.
+//
+// Two caveats:
+// (1) it is tempting to send the leaf Frame directly to the profiler
+// goroutine instead of making a copy of the stack, since a Frame is a
+// spaghetti stack--a linked list. However, as soon as execution
+// resumes, the stack's Frame.pc values may be mutated, so Frames are
+// not safe to share with the asynchronous profiler goroutine.
+// (2) it is tempting to use Callables as keys in a map when tabulating
+// the pprof protocols's Function entities. However, we cannot assume
+// that Callables are valid map keys, and furthermore we must not
+// pin function values in memory indefinitely as this may cause lambda
+// values to keep their free variables live much longer than necessary.
+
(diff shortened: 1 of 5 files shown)


starlark/profile.go
```diff
@@ -0,0 +1,419 @@
...
+type profFrame struct {
+	fn  Callable        // don't hold this live for too long (prevents GC of lambdas)
+	pc  uint32          // program counter (Starlark frames only)
+	pos syntax.Position // position of pc within this frame
+}
+
+// profile is the profiler goroutine.
+// It runs until StopProfiler is called.
+func profile(w io.Writer) {
+	// Field numbers from pprof protocol.
```
jayconrod: Consider extracting the encoding functionality into a separate type with methods. Maybe `type profileEncoder struct {...}` which has an `io.Writer`, a `stringIndex map[string]int64` and other maps, and `str` and other methods. I think it's surprising this function has all the encoding logic and the sample reading loop.

starlark/profile.go
```diff
@@ -0,0 +1,419 @@
...
+	enc.bytes(Profile_sample_type, wallNanos.Bytes()) // type of samples
+	enc.int(Profile_period, quantum.Nanoseconds())    // magnitude of sampling period
+	enc.bytes(Profile_period_type, wallNanos.Bytes()) // dimension and unit of period
+
+	start := nanotime()
+
+	// Read profile events from the channel
+	// until it is closed by StopProfiler.
+	for e := range profiler.events {
+		sample := new(bytes.Buffer)
```
jayconrod: There are a lot of `bytes.Buffers` being allocated, both here and in the functions above. Can we reset and reuse these?

starlark/profile.go
```diff
@@ -0,0 +1,419 @@
...
+// and github.com/golang/protobuf.
+//
+// This also avoids the need to materialize a protocol message object
+// tree of unbounded size and serialize it all at the end.
+// The pprof format appears to have been designed to
+// permit streaming implementations such as this one.
+//
+// See https://developers.google.com/protocol-buffers/docs/encoding.
+type protoEncoder struct {
+	w   io.Writer // *bytes.Buffer or *gzip.Writer
```
jayconrod: I don't think it's safe to ignore errors if you allow `*gzip.Writer` here. I'm not sure how much buffering `*gzip.Writer` has internally, but if the underlying `io.Writer` is unbuffered, this could be making a lot of system calls.

Maybe instead, accumulate a number of profile events in a `*bytes.Buffer`, then flush that when it reaches some threshold (maybe somewhere in the range of 8-64 KB)? You would only need to check for errors on flush.

starlark/sys_unix.go
```diff
@@ -0,0 +1,12 @@
+// +build linux darwin
+
+package starlark
+
+import "syscall"
+
+func nanotime() int64 {
+	// gettimeofday (17ns) is almost 3x faster than time.Now (46ns).
+	var tv syscall.Timeval
+	syscall.Gettimeofday(&tv) // can't fail
```
jayconrod: This doesn't seem like the right clock because it could be changed while the process is running. I'm not sure what the canonical correct answer is. I would expect something based on `CLOCK_MONOTONIC`, but I'm not sure of the precision or latency there.

starlark/profile.go
```diff
@@ -0,0 +1,419 @@
...
+// It measures the wall time spent executing Starlark code, and emits a
+// gzipped protocol message in pprof format (github.com/google/pprof).
+//
+// When profiling is enabled, the interpreter calls the profiler to
+// indicate the start and end of each "span" or time interval. A leaf
+// function (whether Go or Starlark) has a single span. A function that
+// calls another function has spans for each interval in which it is the
+// top of the stack. (A LOAD instruction also ends a span.)
+//
+// At the start of a span, the interpreter records the current time in
```
jayconrod: It sounds like this will measure time spent inside Go code, but it can't record stacks inside Go.

Do you have a sense of what the overhead is for having the profiler on? Calling `gettimeofday` on every span seems like it would be expensive. How many times does that typically happen inside a quantum?

Since profiling is being driven by the interpreter, maybe there are other options available? The interpreter could count instructions executed and check the time after ~1000 instructions (and before/after every call to Go code). That could be checked on back-branch and call instructions to avoid branch in the main interpreter loop.

starlark/sys_generic.go
```diff
@@ -0,0 +1,8 @@
+// +build !linux,!darwin
```
jayconrod: Need a copyright notice for these files.

starlark/profile.go
alandonovan: I tried this, but I found the code less readable. First, it required either splitting the big const declaration into chunks, one per helper function, or moving it to the top level, which pollutes the package namespace. (I often find a bunch of lambdas closing over some shared state to be preferable to a struct with explicit fields, because it is less verbose and it keeps declarations local. In Go, functions nest, but methods don't.)

Second, we're not aiming for a clean abstraction of a "profile writer". Everything we do here is custom made to minimally serve the constraints imposed by the encoding, and the current code doesn't try to hide that. If we had a profileWriter type, its methods signatures would still reveal a mix of pprof-proto generic concepts and Starlark-specific ones. Separating the event loop and the profile writer into different functions wouldn't actually disentangle those concepts.

starlark/profile.go
alandonovan: Oops, I neglected your second question. The gzip.Writer compresses data in 240 byte chunks, so that's the size of each underlying write to the file descriptor. I added a bufio.Writer to write in 4KB chunks, though I suspect the speed improvement is intangible in practice.

alandonovan:
PTAL

starlark/profile.go
jayconrod: If the profiler overhead is less than 10%, then this is probably overengineering, so feel free to disregard.

The counting instructions idea would reduce the number of calls to gettimeofday. You don't actually need to count instructions executed, since you're still measuring time. Just increment a counter when you enter a function or take a loop back branch. When the counter reaches some threshold, that's when you call gettimeofday.

alandonovan: The problem is that any function call could potentially take a long time, so we need to inspect the clock at least at every call return (and after a module load instruction) otherwise we're likely to sample the stack only after the most expensive frame has been popped. The vdso implementation of gettime is fast enough (19ns) that even redundantly calling it twice per span is not that slow. (Redundant because since a function call entails end-start in the caller and start-end in the callee, which is four clock ops even though two would suffice.)

Issue #186, opened 2019-04-04
starlark: profiler records incorrect Profile.time (Jan 1 1970)


alandonovan:
Most likely a consequence of using the MONOTONIC not REALTIME clock.

josharian:
I haven't had a chance to fully grok through the original PR or the subsequent one, but I did see a comment along the lines of "CLOCK_GETTIME is not available". But I believe it should be, in golang.org/x/sys. I added it recently for darwin (https://github.com/golang/sys/commit/3b5209105503162ded1863c307ac66fec31120dd) and it was there already for linux.

Please ignore this comment if it is unhelpful. :)

alandonovan:
Thanks. I've noted this in the comment.

PR #187, opened 2019-04-04
starlark/profile: use real time (not monotonic) clock for Profile.time

alandonovan:
Also, add integration test of profiler and 'go tool pprof'.

### DIFF
--- a/starlark/profile.go
+++ b/starlark/profile.go
@@ -327,11 +327,12 @@ func profile(w io.Writer) {
 	wnenc.int(ValueType_unit, str("nanoseconds"))
 
 	// informational fields of Profile
-	enc.bytes(Profile_sample_type, wallNanos.Bytes()) // type of samples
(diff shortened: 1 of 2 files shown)


starlark/profile.go
```diff
@@ -345,9 +346,7 @@ func profile(w io.Writer) {
 		enc.bytes(Profile_sample, sample.Bytes())
 	}
 
-	end := nanotime()
-	enc.int(Profile_time_nanos, start)         // start time of profile
-	enc.int(Profile_duration_nanos, end-start) // duration of profile
+	enc.int(Profile_duration_nanos, time.Since(start).Nanoseconds())
```
jayconrod: This fixes #186, right?

If the start time is the only issue, I think it makes sense to record the start time with `time.Now`, but call `nanotime` twice separately to measure the duration.

alandonovan: Good point. Done.

PR #193, opened 2019-04-19
starlark: remove deprecated Frame type

alandonovan:
Also, an optimization: use the slack portion of the
Thread.stack slice as a free-list of frames, to avoid
allocating a frame on most calls.

THIS IS A BREAKING API CHANGE.
It will be submitted on May 6.
Clients: please switch to the new CallStack API before then.


alandonovan:
Hard to tell; I found only one use of thread.TopFrame in GitHub global search (which I'm not sure I trust), and none of starlark.Frame.

PR #198, opened 2019-05-14
hash: allow only strings, and use a deterministic algorithm

alandonovan:
Previously, hash would allow any operand defined as hashable
by the spec, and would use (for strings at least) a fast
hardware implementation that varies across process restarts.

+ Test.

### DIFF
--- a/starlark/library.go
+++ b/starlark/library.go
@@ -19,6 +19,7 @@ import (
 	"strconv"
 	"strings"
 	"unicode"
+	"unicode/utf16"
 	"unicode/utf8"
 
 	"go.starlark.net/syntax"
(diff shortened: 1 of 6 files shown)


alandonovan:
@laurentlb -- this change brings the Go and Java impls closer, and fixes a bug in which the set of targets in a Bazel package would vary from one run to the next.

alandonovan:
A question: should Starlark in Go attempt to generate the exact same hashes as Starlark-in-Java, which hashes the sequence of UTF-16 codes using a polynomial accumulator? Transcoding valid UTF-8 to 16 is not hard, but it does seem to be indulging bad habits of writing code that depends on hashing for its observed results. On the other hand, that is exactly what some users do.

starlark/library.go
```diff
@@ -588,12 +588,20 @@ func hasattr(thread *Thread, _ *Builtin, args Tuple, kwargs []Tuple) (Value, err
...
 		return nil, err
 	}
-	h, err := x.Hash()
-	return MakeUint(uint(h)), err
+
+	// The Starlark spec requires that the hash function be
+	// deterministic across all runs (motivated by the need
+	// for reproducibility of builds). Thus we cannot call
+	// String.Hash, which may use the fastest implementation
+        // available, even if it varies across process restarts,
```
jayconrod: Indentation is a little off here.

alandonovan: Indentation fixed.

Implemented the algorithm specified in j.l.String docs.

PR #205, opened 2019-05-22
syntax: drop legacy (0644) octal literals

alandonovan:
Use 0o644 instead.

See also https://github.com/bazelbuild/bazel/issues/8059


alandonovan:
@laurentlb

PR #208, opened 2019-05-24
Implement fail() buit-in function

Quarz0:
fail() is already implemented as a global function in starlark's java and rust implementations and also used by the [test runners for assertions](https://github.com/bazelbuild/bazel/blob/master/src/test/starlark/starlark_test.py#L81).
https://docs.bazel.build/versions/master/skylark/lib/globals.html#fail

+@laurentlb

### DIFF
--- a/starlark/library.go
+++ b/starlark/library.go
@@ -44,6 +44,7 @@ func init() {
 		"bool":      NewBuiltin("bool", bool_),
 		"chr":       NewBuiltin("chr", chr),
 		"dict":      NewBuiltin("dict", dict),
+		"fail":      NewBuiltin("fail", fail),
 		"dir":       NewBuiltin("dir", dir),
 		"enumerate": NewBuiltin("enumerate", enumerate),
 		"float":     NewBuiltin("float", float), // requires resolve.AllowFloat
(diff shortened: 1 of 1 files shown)


starlark/library.go
```diff
@@ -888,6 +889,32 @@ func print(thread *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error
 	return None, nil
 }
 
+// TODO: add link to spec.md
+func fail(thread *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error) {
+	if len(kwargs) > 0 {
+		return nil, fmt.Errorf("fail does not accept keyword arguments")
```
alandonovan: If you use UnpackPositionalArgs, you don't need this check.

But given the behavior of fail I proposed, you don't need the Unpack functions. Just grab starlark.AsString(args[0]), and all of kwargs.

starlark/library.go
```diff
@@ -888,6 +889,32 @@ func print(thread *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error
...
+	}
+	if len(args) == 0 {
+		return None, fmt.Errorf("none")
```
alandonovan: All built-in functions should include their name in the error message. So this should be "fail: no arguments".

But UnpackPositionalArguments takes a min argument to require a minimum number of positional arguments. If you set it to 1, you don't need this check.

starlark/library.go
```diff
@@ -888,6 +889,32 @@ func print(thread *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error
...
+	}
+	attr := ""
+	if err := UnpackPositionalArgs("fail", args, kwargs, 0, &args[0], &attr); err != nil {
```
alandonovan: Use a new variable instead of overwriting args[0]. If you make that variable a string, you don't need the check on L909. For example:

var msg, attr string
if err := UnpackPositionalArgs("fail", args, kwargs, 1, &msg, &attr); err != nil {

starlark/library.go
```diff
@@ -888,6 +889,32 @@ func print(thread *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error
...
+		return nil, err
+	}
+	buf := new(strings.Builder)
+	if len(attr) > 0 {
+		buf.WriteString(fmt.Sprintf("attribute %s: ", attr))
```
alandonovan: fmt.Fprintf(buf, ...)

buf implements io.Writer, so you can Fprintf directly to it.

starlark/library.go
```diff
@@ -888,6 +889,32 @@ func print(thread *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error
...
+	}
+
+	if s, ok := AsString(args[0]); ok {
+		buf.WriteString(s)
+	} else {
+		writeValue(buf, args[0], nil)
+	}
+
+	return None, fmt.Errorf(buf.String())
```
alandonovan: Replace None with nil here. If you return an error you don't need to return value too.

It is not safe to call fmt.Errorf on an arbitrary string such as buf.String(), as it may contain % signs.  Instead, call fmt.Errorf("%s", buf).

starlark/library.go
Quarz0: My initial implementation was basically making an exact copy of fail() in Java which accepted 0 args, that's why I didn't set the min to 1.

starlark/library.go
Quarz0: Again, I included args[0] because java's implementation accepted the first arg as any type not just strings.

Quarz0:
Thanks, I'll file a spec change issue and come back to this PR later

PR #209, opened 2019-05-28
doc: print: drop mention of kwargs

alandonovan:
Updates bazelbuild/starlark#47


alandonovan:
@laurentlb bazelbuild needs the same change.

laurentlb:
cc @Quarz0 to send a PR for bazelbuild/starlark

PR #223, opened 2019-07-11
spec: add '0' as valid decimal_lit

mewmew:
Since each of octal_lit, hex_lit and binary_lit
specify a mandatory prefix (e.g. 0o, 0x and 0b),
we extend decimal_lit to allow for the decimal
literal zero.


mewmew:
Also included a fix of syntax highlighting in this PR.

Prior to 8216e79, data types were highlighted as follows:

2019-07-11-153440_1305x368_scrot

alandonovan:
Thanks!

PR #235, opened 2019-08-19
starlark,resolve,internal/compile: replace log.Fatal with panic

aarzilli:
I've replaced some calls to `log.Fatal` and `log.Fatalf` with panics. I believe those are all unreachable internal errors (in practice) so this is a somewhat inconsequential nitpick; however if this code were ever to execute, I believe a recoverable panic, carrying stacktrace information is preferrable to getting the calling process instantly killed.


aarzilli:
Done.

alandonovan:
Sorry, do you mind changing the CL description to say "so that we get a stack"? I don't agree that a library shouldn't kill its process. I think it's the correct way to handle unrecoverable errors.

aarzilli:
Done.

PR #238, opened 2019-08-21
syntax: add benchmarks for scanning and parsing

alandonovan:


syntax/parse.go
```diff
@@ -47,6 +47,23 @@ func Parse(filename string, src interface{}, mode Mode) (f *File, err error) {
 	return f, nil
 }
 
+// ScanAndDiscard tokenizes the input data and discards the tokens.
+// Parameters are as for Parse.
+// It exists only for internal benchmarking purposes.
```
jayconrod: It's unfortunate this needs to be part of the external API for benchmarking. Could the `"scan"` benchmark be moved to the `syntax` package instead?

PR #244, opened 2019-10-04
starlark: preserve the backtrace in load() errors

nicks:
The errors from a load() were dropping the stack trace on the floor. Hopefully this should be non-controversial.

I also considered using the Go 1.13 error wrapping API, but decided that there are probably too many starlark users on Go 1.12

### DIFF
--- a/starlark/eval.go
+++ b/starlark/eval.go
@@ -190,6 +190,7 @@ func (stack CallStack) String() string {
 type EvalError struct {
 	Msg       string
 	CallStack CallStack
+	cause     error
 }
 
 // A CallFrame represents the function name and current
@@ -210,6 +211,7 @@ func (thread *Thread) evalError(err error) *EvalError {
 	return &EvalError{
 		Msg:       err.Error(),
 		CallStack: thread.CallStack(),
+		cause:     err,
 	}
 }
 
@@ -220,6 +222,8 @@ func (e *EvalError) Error() string { return e.Msg }
 func (e *EvalError) Backtrace() string {
 	return fmt.Sprintf("%sError: %s", e.CallStack, e.Msg)
 }
+
+func (e *EvalError) Unwrap() error { return e.cause }
 
 // A Program is a compiled Starlark program.
 //
--- a/starlark/eval_test.go
+++ b/starlark/eval_test.go
@@ -35,6 +35,11 @@ func option(chunk, name string) bool {
 	return strings.Contains(chunk, "option:"+name)
 }
 
+// Wrapper is the type of errors with an Unwrap method; see https://golang.org/pkg/errors.
+type Wrapper interface {
+	Unwrap() error
+}
+
 func TestEvalExpr(t *testing.T) {
 	// This is mostly redundant with the new *.star tests.
 	// TODO(adonovan): move checks into *.star files and
@@ -481,19 +486,19 @@ func TestInt(t *testing.T) {
 	}
 }
 
-func TestBacktrace(t *testing.T) {
-	getBacktrace := func(err error) string {
-		switch err := err.(type) {
-		case *starlark.EvalError:
-			return err.Backtrace()
-		case nil:
-			t.Fatalf("ExecFile succeeded unexpectedly")
-		default:
-			t.Fatalf("ExecFile failed with %v, wanted *EvalError", err)
-		}
-		panic("unreachable")
+func backtrace(t *testing.T, err error) string {
+	switch err := err.(type) {
+	case *starlark.EvalError:
+		return err.Backtrace()
+	case nil:
+		t.Fatalf("ExecFile succeeded unexpectedly")
+	default:
+		t.Fatalf("ExecFile failed with %v, wanted *EvalError", err)
 	}
+	panic("unreachable")
+}
 
+func TestBacktrace(t *testing.T) {
 	// This test ensures continuity of the stack of active Starlark
 	// functions, including propagation through built-ins such as 'min'.
 	const src = `
@@ -514,7 +519,7 @@ i()
   crash.star:3:12: in g
   crash.star:2:19: in f
 Error: floored division by zero`
-	if got := getBacktrace(err); got != want {
+	if got := backtrace(t, err); got != want {
 		t.Errorf("error was %s, want %s", got, want)
 	}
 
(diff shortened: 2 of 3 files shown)


nicks:
hmmm...that makes sense. A couple ideas:

1) Would you accept a change to implement the Unwrap() interface, so that API callers can at least get the underlying error? This would make it more compatible with the Go 1.13 way of doing errors

2) Alternatively, we could check if the root of the EvalError call stack against the current call stack, and only do this if they match?  That way, Load() implementations that use the same thread would work ok, and we wouldn't break Load() implementations that use a different thread.

alandonovan:
Idea #1 sounds fine.

nicks:
great! 

I just pushed a change that uses xerrors.Wrappers. If we don't want starlark to depend on xerrors, happy to change it to use a different dependency (or create structs that implement Unwrap() manually)

starlark/eval.go
```diff
@@ -188,8 +188,9 @@ func (stack CallStack) String() string {
 // An EvalError is a Starlark evaluation error and
 // a copy of the thread's stack at the moment of the error.
 type EvalError struct {
-	Msg       string
-	CallStack CallStack
+	Msg        string
+	CallStack  CallStack
+	wrappedErr error
```
alandonovan: cause error

In hindsight it's unfortunate that Msg is an exposed field.

starlark/eval_test.go
```diff
@@ -540,12 +542,67 @@ Error: join: in list, want string, got int`,
...
+	const want = `Traceback (most recent call last):
+  root.star:2:1: in <toplevel>
+Error: cannot load crash.star: floored division by zero`
+	if got := getBacktrace(t, err); got != want {
+		t.Errorf("error was %s, want %s", got, want)
+	}
+
+	unwrapEvalError := func(err error) *starlark.EvalError {
+		var result *starlark.EvalError
+		for true {
```
alandonovan: for {

starlark/eval_test.go
```diff
@@ -540,12 +542,67 @@ Error: join: in list, want string, got int`,
...
+			evalErr, isEvalErr := err.(*starlark.EvalError)
```
alandonovan: if evalErr, ok := err.(*starlark.EvalError); ok {
    result = evalErr
}

nicks:
all comments addressed! ptal?

starlark/interp.go
```diff
@@ -503,7 +505,7 @@ loop:
 			dict, err2 := thread.Load(thread, module)
 			thread.beginProfSpan()
 			if err2 != nil {
-				err = fmt.Errorf("cannot load %s: %v", module, err2)
+				err = xerrors.Errorf("cannot load %s: %w", module, err2)
```
alandonovan: I'm a little uneasy about this function, partly because I don't know exactly what it does, and partly because it adds an unnecessary dependency. IIUC, this code formats the error message and retains the underlying error, exposed through a Unwrap. It should be possible to construct an EvalError that does that (now that it has 'cause').

The other import of xerrors in the test can be eliminated by simply duplicating the relevant interface type declaration.

Can we replace this code with a new EvalError

nicks: ya, i was bothered by the xerrors dependency too.

I added a new loadError that implements the same interface.

wrapping this as an EvalError isn't quite right, because (as you pointed out at the start of this review), EvalError has a backtrace and that backtrace would be wrong here (i.e., it makes the tests fail)

starlark/eval_test.go
```diff
@@ -35,6 +35,12 @@ func option(chunk, name string) bool {
 	return strings.Contains(chunk, "option:"+name)
 }
 
+// Copy of the xerrors.Wrapper interface
```
alandonovan: // Wrapper is the type of errors with an Unwrap method; see https://golang.org/pkg/errors.

starlark/eval_test.go
```diff
@@ -481,19 +487,19 @@ func TestInt(t *testing.T) {
...
-		switch err := err.(type) {
-		case *starlark.EvalError:
-			return err.Backtrace()
-		case nil:
-			t.Fatalf("ExecFile succeeded unexpectedly")
-		default:
-			t.Fatalf("ExecFile failed with %v, wanted *EvalError", err)
-		}
-		panic("unreachable")
+func getBacktrace(t *testing.T, err error) string {
```
alandonovan: func backtrace

starlark/eval_test.go
```diff
@@ -540,12 +546,66 @@ Error: join: in list, want string, got int`,
...
+	}
+
+	unwrapEvalError := func(err error) *starlark.EvalError {
+		var result *starlark.EvalError
+		for {
+			if evalErr, ok := err.(*starlark.EvalError); ok {
+				result = evalErr
+			}
+
+			wrapper, isWrapper := err.(Wrapper)
```
alandonovan: // TODO: use errors.Unwrap when go >=1.13 is everywhere.

starlark/interp.go
```diff
@@ -590,6 +593,21 @@ loop:
 	return result, err
 }
 
+type loadError struct {
```
alandonovan: Is this necessary? I think EvalError should suffice.

nicks: wrapping this as an EvalError isn't quite right, because (as you pointed out at the start of this review) EvalError has a backtrace. That backtrace would be wrong here, and makes the existing test TestThreadLoad_ParallelCycle fail

starlark/eval_test.go
nicks: done

starlark/eval_test.go
nicks: done

starlark/eval_test.go
nicks: done

starlark/interp.go
alandonovan: Right, but I meant its CallStack field would be nil.  Actually, probably best not to, since the starlark.Call checks that all returned errors are EvalError, and it would fail to re-wrap the load error in that case.

OK, let's keep a distinct type, but let's declare it thus:
```
type wrappedError struct {
    msg string
    cause err
}
```

nicks: done!

PR #254, opened 2019-12-02
.travis.yml: test that module is tidy

jayconrod:
Removed "sudo: false". It has no effect.

Added explicit test script.

* 'go test -mod=readonly ./...' will report an error if go.mod or
  go.sum is incomplete, since no requirements may be added with
  -mod=readonly.
* The rest of the script checks that go.mod and go.sum do not change
  after running 'go mod tidy'. This verifies that there are no
  redundant requirements or sums to remove.

Changed minimum Go version to 1.11.x. This is the oldest version
with support for modules. Older versions will still work in GOPATH
mode, but versions lower than 1.12 are unsupported by the Go, so
there's little point in adding additional test code for them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ go:
     - "1.11.x"
     - "master"
 
+env:
+    - "GO111MODULE=on"
+
 script:
     - "go test -mod=readonly ./..."
     - "cp go.mod go.mod.orig"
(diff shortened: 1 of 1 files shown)


alandonovan:
LGTM I guess?

jayconrod:
> LGTM I guess?

Seems like the best approval I can hope for with a change to a file with a loosely documented YAML schema.

(Incidentally, https://config.travis-ci.com/ and https://config.travis-ci.com/ref/language/go were helpful.)

Sorry for the trial and error. Not easy to test it locally.

PR #257, opened 2020-01-25
Move Travis CI badge to the top of file [skip ci]

mbrukman:
Travis CI is more of a status badge, not an aspect of documentation, and
is often kept at the top of the project's README, together with other
badges, such as Go Report Card, etc., which don't belong in the
"documentation" section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,8 @@
 
 # Starlark in Go
 
+[![Travis CI](https://travis-ci.org/google/starlark-go.svg)](https://travis-ci.org/google/starlark-go)
+
 This is the home of the _Starlark in Go_ project.
 Starlark in Go is an interpreter for Starlark, implemented in Go.
 Starlark was formerly known as Skylark.
@@ -42,8 +44,6 @@ languages and compilers.
 * Mailing list: [starlark-go](https://groups.google.com/forum/#!forum/starlark-go)
 
 * Issue tracker: [https://github.com/google/starlark-go/issues](https://github.com/google/starlark-go/issues)
-
-* Travis CI: [![Travis CI](https://travis-ci.org/google/starlark-go.svg) https://travis-ci.org/google/starlark-go](https://travis-ci.org/google/starlark-go)
 
 ### Getting started
 


mbrukman:
FWIW, in the next PR, I was going to add two more badges to the top of the file, so it would look as follows:

[Travis CI](https://travis-ci.org/google/starlark-go) [Go Report Card](https://goreportcard.com/report/github.com/google/starlark-go) [Go docs](https://godoc.org/go.starlark.net)

I can add those additional badges to this PR, if you'd like, assuming you're OK with this direction.

alandonovan:
Sure, sounds good to me, so long as the "go report card" widget doesn't force us to engage with all the trivial findings of golint.

mbrukman:
Yes, there are many Go Report Card findings about not using underscores, comments for exported names, and consistency in receiver names, e.g.,

> Line 190: warning: receiver name x should be consistent with previous receiver name i for Int (golint)

Do you think these are worth fixing?

If you don't care about any of those, maybe it doesn't make sense to add the Go Report Card badge. Or just add it with image-only, but without the link so it's easy to tell if it drops from A+ to A (at which point it's something to look into), but otherwise ignore? Up to you; let me know what you prefer.

The GoDoc reference badge seems useful and uncontroversial and shouldn't cause any additional effort, so that seems like a net win. Thoughts?

alandonovan:
On Mon, 27 Jan 2020 at 10:15, Misha Brukman <notifications@github.com>
wrote:

> Yes, there are many Go Report Card findings about not using underscores,
> comments for exported names, and consistency in receiver names, e.g.,
>
> Line 190: warning: receiver name x should be consistent with previous
> receiver name i for Int (golint)
>
> Do you think these are worth fixing?
>

No. Most golint findings are a waste of time.


If you don't care about any of those, maybe it doesn't make sense to add
> the Go Report Card badge. Or just add it with image-only, but without the
> link so it's easy to tell if it drops from A+ to A (at which point it's
> something to look into), but otherwise ignore? Up to you; let me know what
> you prefer.
>
> The GoDoc reference badge seems useful and uncontroversial and shouldn't
> cause any additional effort, so that seems like a net win. Thoughts?
>
I have no strong feelings yet about Go Report Card, but the moment we start
to game the metric by making a pointless (or harmful) change to the source
just to pacify the static analysis tool is the day I delete the badge. :)

PR #258, opened 2020-01-25
Fix ineffective assignments

mbrukman:
* `return nil, err` after assigning to `err` (rather than ignoring it)
* removed initialization of a variable just before it's re-assigned

Both of these issues were found in the Go Repord Card:
https://goreportcard.com/report/github.com/google/starlark-go#ineffassign

### DIFF
--- a/syntax/scan.go
+++ b/syntax/scan.go
@@ -276,6 +276,7 @@ func readSource(filename string, src interface{}) ([]byte, error) {
 		data, err := ioutil.ReadAll(src)
 		if err != nil {
 			err = &os.PathError{Op: "read", Path: filename, Err: err}
+			return nil, err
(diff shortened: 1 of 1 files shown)


syntax/scan.go
```diff
@@ -276,6 +276,7 @@ func readSource(filename string, src interface{}) ([]byte, error) {
 		data, err := ioutil.ReadAll(src)
 		if err != nil {
 			err = &os.PathError{Op: "read", Path: filename, Err: err}
+			return nil, err
```
alandonovan: Good catch. I suspect no-one ever passes a Reader here which is why it went unnoticed.

alandonovan:
Thanks!

PR #259, opened 2020-01-25
Also run tests with Go 1.12.x and 1.13.x

mbrukman:
The current setup only runs tests with Go 1.11.x and the latest source
code on the `master` branch, but is not running the more recent releases
1.12.x and 1.13.x which may be widely in use. This change includes both
of those versions as well.


alandonovan:
Thanks!

PR #265, opened 2020-03-25
syntax: strict string escapes

alandonovan:
This change causes Starlark, like Go, to reject backslashes that
are not part of an escape sequence. Previously they were treated
literally, so \ ( would encode a two-character string, and much
code relied on this, especially for regular expressions.

This may break some programs, but the fix is simple:
double each errant backslashes.

Python does not yet enforce this behavior, but since 3.6
has emitted a deprecation warning for it.

Also, document string escapes.

This is Google issue b/34519173.

Change-Id: I5c9609a4e28d58593e9d6918757bca2cfd838d51


alandonovan:
@laurentlb

doc/spec.md
```diff
@@ -321,6 +321,132 @@ hex_digit     = '0' … '9' | 'A' … 'F' | 'a' … 'f' .
...
+escaping:
+
+```python
+"a\nb"		# "a\nb"  = 'a' + '\n' + 'b'
+r"a\nb"		# "a\\nb" = 'a' + '\' + '\n' + 'b'
+
+"a\
+b"		# "ab"
+r"a\
+b"		# "a\\nb"
```
jayconrod: `"a\\\nb"` I think

doc/spec.md
```diff
@@ -321,6 +321,132 @@ hex_digit     = '0' … '9' | 'A' … 'F' | 'a' … 'f' .
...
+other than an escaped quotation mark (which denotes a literal
+quotation mark), or an escaped newline (which denotes a backslash
+followed by a newline). This form of quotation is typically used when
+writing strings that contain many quotation marks or backslashes (such
+as regular expressions or shell commands) to reduce the burden of
+escaping:
+
+```python
+"a\nb"		# "a\nb"  = 'a' + '\n' + 'b'
+r"a\nb"		# "a\\nb" = 'a' + '\' + '\n' + 'b'
```
jayconrod: `'\\'`

Issue #267, opened 2020-04-03
Support for type-specific unpack logic


landism:
The project I work on defines a number of starlark builtins with parameter types not quite supported by `starlark.UnpackArgs`, e.g., we want an argument of type `Union[None|str]`, or `Union[str|List[str]]`.

Right now, we're unpacking these args into a variable of type `starlark.Value` and then doing post-unpack validation / casting, which seems like a lot of distracting noise in the builtin definitions.

A solution that comes to mind is to add a new interface `starlark.Unpackable` with the method `Unpack(starlark.Value, ptr interface{})`, and have `unpackOneArg`'s switch statement call that if it exists.

This would allow us to have UnpackArgs handle arg validation for custom types as it already does for default types.

I'm happy to write this, but wanted to check on interest or alternatives before doing so.

Thanks!

alandonovan:
Sounds reasonable. Care to send a sketch of your idea so we can discuss it in more detail?

PR #272, opened 2020-05-19
Support for type-specific unpack logic

nicks:
Fixes https://github.com/google/starlark-go/issues/267

This bit me recently, so I took the liberty of proposing something very minimalist that borrows from the json package's Unmarshaler

### DIFF
--- a/starlark/unpack.go
+++ b/starlark/unpack.go
@@ -10,6 +10,12 @@ import (
 	"strings"
 )
 
+// An Unpacker defines custom argument unpacking behavior.
+// See UnpackArgs.
+type Unpacker interface {
+	Unpack(v Value) error
+}
+
 // UnpackArgs unpacks the positional and keyword arguments into the
 // supplied parameter variables.  pairs is an alternating list of names
 // and pointers to variables.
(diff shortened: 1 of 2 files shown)


starlark/unpack.go
```diff
@@ -10,6 +10,12 @@ import (
 	"strings"
 )
 
+// Unpacker defines an interface for custom types that implement
+// their own unpacking logic.
```
alandonovan, line 14: // An Unpacker defines custom argument unpacking behavior.
// See UnpackArgs.

starlark/unpack.go
```diff
@@ -10,6 +10,12 @@ import (
...
+type Unpacker interface {
+	Unpack(v Value) error
+}
+
```
alandonovan, line 18: Update the UnpackArgs comment:

```
// it and all following parameters are optional.
//
// If the variable implements Unpacker, its Unpack argument
// is called with the argument value, allowing an application
// to define its own argument validation and conversion.
//
// If the variable implements Value...
```

(Why won't GitHub let me comment on faraway lines? Grrr.)

nicks: updated!

starlark/unpack.go
nicks: updated!

starlark/unpack.go
```diff
@@ -172,6 +182,11 @@ func unpackOneArg(v Value, ptr interface{}) error {
 		}
 		*ptr = it
 	default:
+		unpacker, ok := ptr.(Unpacker)
+		if ok {
```
alandonovan, line 186: This could be a switch case, so let's do that:
```
case Unpacker:
   return ptr.Unpack(v)
```

Please insert it as the first case. Order doesn't matter, since the cases are all currently disjoint, but if we add new cases I think this one should take priority.

Sorry for not pointing this out before.

nicks: good point, done

Issue #278, opened 2020-06-12
scanner should allow \" within both '...' and "..." quotations


nicks:
Repro steps:

test.sky
```
x = '\"'
```

Run it with:

```
$ starlark test.sky
../test.sky:1:5: invalid escape sequence \"
```

I understand the argument why this *shouldn't* be a valid string escape, per
https://github.com/google/starlark-go/pull/265

but this seems inconsistent with Python 3.8, which treats this as valid, and doesn't emit any warnings:
python3 -Wall test.sky

laurentlb:
The Java implementation explicitly supports it:
https://cs.opensource.google/bazel/bazel/+/034b09e810de42baac8e1a431eab6acd91b2c8a1:src/main/java/com/google/devtools/build/lib/syntax/Lexer.java;l=315,318

The Python 3 documentation mentions a deprecation for the unrecognized escape sequences (like Starlark). But the documentation explicitly lists `\"` and `\'` as valid:
https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals

Even if it's surprising to have those useless escapes, I think starlark-go should allow them.

alandonovan:
Yes, as you point out this is inconsistent with Python 3.8---though presumably Python 3.9 or later will catch up. This was the explicit intent of #265: 

https://github.com/google/starlark-go/pull/265/files#diff-fe578779dc15d59765499b7151c7c926R123:
```
// In Starlark, like Go, a backslash must escape something.
// then it appears as literal text in the string.
// (Python still treats unnecessary backslashes literally,
// but since 3.6 has emitted a deprecation warning.)
```

https://github.com/google/starlark-go/pull/265/files#diff-a37bdfbd0d2f3627958003f87033871eR449:
```
It is an error for a backslash to appear within a string literal other
than as part of one of the escapes described above.
```

The motivation for strict escapes is that you can never add a new interpretation for say `\k` if we already permit `\k` to appear in strings. This is the reason Python is trying to eliminate this feature.

@laurentlb why do you think starlark-go should add back this feature of which Python is trying to rid itself? Reading starlark-java, it appears to be moving in the same direction:
https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/syntax/FileOptions.java#L48

laurentlb:
The bug is specifically about backslash+quote (I think we all agree that `\k` should be rejected).

Here's Python behavior:

```
$ python -Wall
[...]

PR #279, opened 2020-06-15
syntax: permit \" and \' escapes in both "..." and '...' literals

alandonovan:
Clarify spec on this issue.

Fixes https://github.com/google/starlark-go/issues/278

Also, fix semantic merge conflict in testdata/json.star tests that
relied on "\u" being interpreted as "\\u", causing tests to fail.


alandonovan:
@laurentlb @nicks

PR #280, opened 2020-06-16
int: reduce allocation by representing small ints as pointers

alandonovan:


starlark/int_generic.go
```diff
@@ -0,0 +1,33 @@
+//+build !linux,!darwin
```
rsc, line 1: This pair of lines should be one line:

// +build !linux,!darwin !amd64,!arm64,!mips64x,!ppc64x

starlark/int_posix64.go
```diff
@@ -0,0 +1,55 @@
+//+build linux darwin
+//+build amd64 arm64 mips64x ppc64x
+
+package starlark
+
+// Optimized Int implementation for 64-bit machines running POSIX (for mmap).
+// It reserves a portion of the address space to represent int32 values.
```
jayconrod, line 7: Consider expanding this:

```
It reserves a 4GB range of the virtual address space using mmap and represents int32 values
as addresses within that range. This disambiguates int32 values from big.Int pointers.
```

starlark/int_posix64.go
```diff
@@ -0,0 +1,55 @@
...
+// Precondition: x cannot be represented as int32.
+func makeBigInt(x *big.Int) Int { return Int{intImpl(x)} }
+
+// smallints is the base address of a 2^32 byte memory region.
+// Pointers to addresses in this region represent int32 values.
+// We assume smallints is not at the very top of the address space.
+var smallints = reserveAddresses(1 << 32)
+
+func reserveAddresses(len int) uintptr {
+	b, err := unix.Mmap(-1, 0, len, unix.PROT_READ, unix.MAP_PRIVATE|unix.MAP_ANONYMOUS)
```
jayconrod, line 50: Would it be possible to use something similar to NaN boxing instead? That would let you avoid the `mmap`.

amd64 processors only support 48-bit virtual addresses. They're sign-extended to 64-bits. There's a big range in the middle that could be used to store other values. I believe arm64 processors have a similar limitation. I don't know about mips.

starlark/int_posix64.go
alandonovan: Done.

starlark/int_posix64.go
alandonovan: Isn't NaN-boxing the trick by which JS implementations embed 53-bit ints, 48-bit pointers, and other tagged value types into each of the many kinds of NaN? We're trying to hide numbers inside pointers, but Nan-boxing goes in the opposite direction: hiding pointers inside numbers, which is problematic for garbage collection. (big.Ints would falsely appear unreachable.)

The use of mmap is not such a problem: it is relatively portable (it's available as VirtualAlloc on Windows), and the current approach doesn't depend on features of the hardware other than word size.

starlark/int_generic.go
alandonovan: Good catch. I always forget whether space means AND or OR.

starlark/int_posix64.go
jayconrod: I just meant NaN boxing as an analogy: using the dead bits of a large type (in this case `unsafe.Pointer`) to store a smaller numeric value (`int32`).

Fair enough on `mmap` though. I'm still stuck in the dark ages of 32-bit where virtual address space was a scarce resource. :)

rsc: If you were willing to understand what "valid" pointers are on the host system, you could take a space from the "invalid" pointers by fiat, without asking mmap for address range. On amd64 the valid pointers start with 17 0s or 17 1s; if you pick a different prefix, like "01", then you could store up to 62-bit numbers in the remaining bits. But on non-amd64 that specific trick wouldn't work. On some 64-bit systems Go supports (I forget which one) the entire 64-bit range is technically valid. The mmap trick is more portable for sure.

alandonovan: I'd like to stick with the mmap version for now because it seems less fragile overall, but I will remember these ideas for when I implement my next proper compiler to object code. :)

alandonovan:
> I'm still stuck in the dark ages of 32-bit where virtual address space was a scarce resource. :)

It turns out there's quite a lot of spare address space just sitting there awaiting one's creativity. This change only squandered 1 part in 2^32 of it. :)

benhoyt:
@alandonovan This is fun! Out of interest, do you have performance numbers on how much this helped on the targeted platforms? (at least for the microbenchmarks you wrote)

alandonovan:
> do you have performance numbers on how much this helped

Take a look at the commit message: https://github.com/google/starlark-go/pull/280/commits/29194bd0bceafc55570d60844d1410a2e1065ec0

benhoyt:
Aha, of course -- I was being too PR-centric. Thanks!

PR #282, opened 2020-06-19
type Int struct{ impl intImpl } should type Int struct{ in…

anyktx:
type Int struct{ impl intImpl } raise bug at starlark/int_generic.go:22.
should be type Int struct{ intImpl }

### DIFF
--- a/starlark/int.go
+++ b/starlark/int.go
@@ -16,7 +16,7 @@ import (
 // Int is the type of a Starlark int.
 //
 // The zero value is not a legal value; use MakeInt(0).
-type Int struct{ impl intImpl }
+type Int struct{ intImpl }
 
 // --- high-level accessors ---
 


starlark/int.go
```diff
@@ -16,7 +16,7 @@ import (
 // Int is the type of a Starlark int.
 //
 // The zero value is not a legal value; use MakeInt(0).
-type Int struct{ impl intImpl }
+type Int struct{ intImpl }
```
alandonovan, line 19: Thanks for pointing out the problem. Would you mind fixing it by instead making this change to int_generic.go:
```
-       return i.small_, i.big_
+       return i.impl.small_, i.impl.big_
```
I would prefer not to use embedding unnecessarily.

You'll need to sign Google's CLA before we can commit. Or if you like I can make the change directly.

alandonovan:
Update: I rolled the fix into https://github.com/google/starlark-go/pull/283, which fixes the tests for 32-bit platforms (e.g. 386), so this change is no longer needed.

Thanks again for pointing it out.

PR #287, opened 2020-07-02
starlark: use syscall.Mmap to avoid golang.org/x/syscall dependency

alandonovan:


starlark/int_posix64.go
```diff
@@ -51,7 +50,9 @@ func makeBigInt(x *big.Int) Int { return Int{intImpl(x)} }
 var smallints = reserveAddresses(1 << 32)
 
 func reserveAddresses(len int) uintptr {
-	b, err := unix.Mmap(-1, 0, len, unix.PROT_READ, unix.MAP_PRIVATE|unix.MAP_ANONYMOUS)
+	// Use syscall to avoid golang.org/x/sys/unix dependency.
+	const MAP_ANON = 0x1000
```
jayconrod, line 54: `go get` is failing with `bad file descriptor`. Are `MAP_ANON` and `MAP_ANONYMOUS` always the same, and do they have the same value on supported platforms? From https://pkg.go.dev/golang.org/x/sys/unix?tab=doc#pkg-constants (accessing from macOS), it looks like they are both `0x20`.

alandonovan: Well spotted---I thought I had checked on various platforms but I seem to have messed up.  The values do differ. I've added a switch on GOOS. It's ugly but at least if it's every wrong we'll know immediately.

MAP_ANON is a deprecated alias for MAP_ANONYMOUS on Linux, but on Darwin there is only MAP_ANON. It's 0x1000 on all the BSDs, 0x20 on Linux, and 0x100 on Solaris.

Does the OS of your web browser affect pkg.go.dev? I would have thought it always shows Linux.

Fixed now.

jayconrod: I assumed pkg.go.dev showed a platform-specific view, but now that I look at it, I don't think that's actually the case. It links to definitions in Linux-specific files at least. I might be remembering a brainstorming session from long ago.

alandonovan:
> you just need to run go mod tidy to update go.mod and go.sum. golang.org/x/sys is no longer a direct dependency.

Good point, thanks.

PR #289, opened 2020-07-02
starlark: add examples of UnpackArgs

### LINKED ISSUE #111: doc: show examples of UnpackArgs
Example: all arguments must be named:

```
if len(args) > 0 {
    return nil, fmt.Errorf("%s: unexpected positional arguments")
}
if err := UnpackArgs(b, args, kwargs, ...); err != nil {
    return nil, err
}
```

Example: equivalent to f(x, y=1, *, z=2):
```
if len(args) > 2 {
    return nil, fmt.Errorf("%s: unexpected positional arguments")
}
var x, y, z int = 0, 1, 2
if err := UnpackArgs(b, args, kwargs, "x", &x, "y?", &y, "z", &z); err != nil {
    return nil, err
}
```

alandonovan:
Fixes #111

### DIFF
--- a/starlark/unpack.go
+++ b/starlark/unpack.go
@@ -34,10 +34,54 @@ type Unpacker interface {
 // If the variable implements Value, UnpackArgs may call
 // its Type() method while constructing the error message.
 //
-// Beware: an optional *List, *Dict, Callable, Iterable, or Value variable that is
-// not assigned is not a valid Starlark Value, so the caller must
-// explicitly handle such cases by interpreting nil as None or some
-// computed default.
+// Examples:
+//
+//      var (
+//          a Value
+//          b = MakeInt(42)
+//          c Value = starlark.None
+//      )
+//
+//      // 1. mixed parameters, like def f(a, b=42, c=None).
+//      err := UnpackArgs("f", args, kwargs, "a", &a, "b?", &b, "c?", &c)
+//
+//      // 2. keyword parameters only, like def f(*, a, b, c=None).
+//      if len(args) > 0 {
+//              return fmt.Errorf("f: unexpected positional arguments")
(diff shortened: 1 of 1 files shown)


starlark/unpack.go
```diff
@@ -34,10 +34,54 @@ type Unpacker interface {
...
+//      // 1. mixed parameters, like def f(a, b=42, c=None).
+//      err := UnpackArgs("f", args, kwargs, "a", &a, "b?", &b, "c?", &c)
+//
+//      // 2. keyword parameters only, like def f(*, a, b, c=None).
+//      if len(args) > 0 {
+//              return fmt.Errorf("f: unexpected positional arguments")
+//      }
+//      err := UnpackArgs("f", args, kwargs, "a", &a, "b?", &b, "c?")
+//
+//      // 3. positional parameters only, like def f(a, b=42, c=None, /) in Python 3.8.
```
jayconrod, line 54: The `/` after `c=None` is new to me. Is that new in Python 3? Is it supported here?

alandonovan: It's PEP570, new in Python 3.8. (Slash is the opposite of star, geddit?) The parameters are (a, ... /, b, ... *, c, ...), where the a's are positional-only, the b's are positional-or-keyword, and the c's are keyword only. The slash is omitted if there are no a's.

It's not a bad feature as it's symmetric and it helps to reduce inadvertent exposure of parameter names, but I'm not sure Python's function calling convention really needs more stuff. Starlark doesn't currently support it.

alandonovan:
I did initially express them as runnable examples, but I wasn't very happy with the amount of additional code surrounding the one line of interest, especially as I added three or four examples.

PR #291, opened 2020-07-05
starlark: disable int optimization in iOS, which doesn't support mmap

### LINKED ISSUE #290: Fatalf on darwin arm64 iOS
root cause:

https://github.com/google/starlark-go/blob/50ca820fafb940caeda09b96864b9215f8b84c04/starlark/int_posix64.go#L55-L57

when targeting on darwin/arm64, it case `mmap: cannot allocate memory`

alandonovan:
Fixes #290

### DIFF
--- a/starlark/int_posix64.go
+++ b/starlark/int_posix64.go
@@ -1,5 +1,5 @@
 //+build linux darwin
-//+build amd64 arm64 mips64x ppc64x
+//+build amd64 arm64,!darwin mips64x ppc64x
 
 package starlark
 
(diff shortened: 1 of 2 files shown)


alandonovan:
@googlebot I fixed it.

starlark/int_posix64.go
```diff
@@ -10,6 +10,10 @@ package starlark
 // values be represented as an unsafe.Pointer, so that Int-to-Value
 // interface conversion need not allocate.
 
+// Although iOS (arm64,darwin) claims to be a POSIX-compliant,
+// it limits each process to about 700MB of virtual address space,
+// which defeats the optimization.
```
jayconrod, line 15: Consider adding:

```
// TODO(golang.org/issue/38485): darwin,arm64 may refer to macOS in the future.
// Update this when there are distinct GOOS values for macOS, iOS, and other Apple
// operating systems on arm64.
```

alandonovan: Nice. Done.

PR #295, opened 2020-07-22
fix: `mmap: bad file descriptor` on android arm64

Kr328:
Fix https://github.com/google/starlark-go/issues/290#issuecomment-660674313


Kr328:
@googlebot I signed it!

PR #297, opened 2020-07-23
fix: mmap failure on linux after the android fix

### LINKED ISSUE #296: `mmap: bad file descriptor` on Linux x86_64
To reproduce:
```
go get go.starlark.net/starlark
go test -v go.starlark.net/starlark/...
```
Output:
```
2020/07/23 14:13:23 mmap: bad file descriptor
FAIL    go.starlark.net/starlark        0.002s
FAIL
```

Tested and reproducible in

- Arch Linux latest / Go 1.14.6
- Debian sid / Go 1.14.4
- Fedora 32 / Go 1.14.3

felixonmars:
Fixes #296


felixonmars:
The offending change was introduced in #295

alandonovan:
Thanks for the fix, and sorry for not catching that in review---I've been working in C++ this week, and had already become reaccustomed to implicit fall-through.

Issue #252, opened 2019-11-14
Is there any way to stop a running script with a timeout?


Dreamacro:
In my use case, I receive a `function` from the user. And I passthrough params and execute it. But I need a way to stop the function running with a timeout (e.g. 2000ms).

can starlark-go provide a function like CallContext?
```go
thread := &starlark.Thread{
    Name:  "example",
}

globals, _ := starlark.ExecFile(thread, "test.star", f, nil)

mainFn := globals["main"]
dict := starlark.StringDict{
    "param": starlark.String("test"),
}

m := starlarkstruct.FromStringDict(starlarkstruct.Default, dict)
arg := starlark.Tuple{m}

ctx, cancel := context.WithTimeout(context.background(), time.Second)
defer cancel()
starlark.CallContext(ctx, thread, mainFn, arg, nil)
```

alandonovan:
Currently there is no way to stop a function in the middle of Starlark execution, though it is a feature that has been requested before and may be feasible to implement without a major performance penalty. Is the function you wish to interrupt executing a Starlark loop, or is it calling into a built-in function that takes a long time?  If the latter, then you can easily make that built-in function honor the cancellation of a context carried within the thread. For example:

```
thread := &starlark.Thread{Name:  "example"}
ctx, cancel := context.WithTimeout(context.background(), time.Second)
defer cancel()
thread.SetLocal("context", ctx)
starlark.ExecFile(...)

...

func slowBuiltin(thread *Thread, fn *Builtin, args Tuple, kwargs []Tuple) (Value, error) {
    if (thread.Local("context").(context.Context).Err() != nil) { ... give up ... }
    ...
}
```
```

Dreamacro:
My use cases are more like a lightweight FaaS. The user defines a function that I can't control and I call it many times. So when it runs too long, I need to stop/kill it. Is there any difficulty that doesn't implement yet?

alandonovan:
Oh, if you can't trust the user's function then you definitely should not run it in the same address space as your code, as it can trivial consume all available memory and there's nothing we can hope to do about that.

Let the operating system help you. Fork and exec a subprocess in a cgroup or setrlimit container that bounds the memory, CPU, and running time.

maxmcd:
@alandonovan do you think it would be a large effort to implement rough cpu/memory metering and limits? I understand if this isn't the goal of the project and that it wouldn't have a home in this codebase, but is there something about the implementation that would make this difficult?

[...]

Issue #160, opened 2019-02-24
Question: Prevent long-running scripts


bradrydzewski:
Hi there. Is it possible to constrain execution time and resources? I am interested in embedding starlark in a server application that processes user-defined scripts and was interested if there were steps that could be take to mitigate long-running or malicious code.

For example to prevent something like this:

```
def loop():
  x = '';
  for i in range(1, 100000000):
    x = x + 'asdf'

loop()
```

alandonovan:
> Is it possible to constrain execution time and resources?

Not within the same process. It is trivial for a Starlark program to consume all time and memory, and to cause stack overflow. Even without recursion enabled, it's possible to get the interpreter stuck in an endless loop, for example by by creating a cyclic value such as a list that contains itself and then applying built-in recursive operators to it. The str(x) and x==y operators try to defend against this, but the defense is imperfect, and there are likely other operators, including application-defined ones, that are not so defensive.

There are also likely to be ways in which the author of a Starlark script can cause a panic in the interpreter, or worse, a fatal error. I regard all of these as bugs, but they may be hard to find and hard or even impossible to fix. For example, imposing constraints on memory would essentially require a completely separate memory allocator and garbage collector for the Starlark heap, which would complicate the API and implementation and make it much less useful as an embeddable language.

In short, if you're designing a networked server that accepts Starlark scripts from clients, you are trusting the health of your server to those clients. So this might be a reasonable design for dynamically updatable configuration (e.g. a request filter) where the scripts are and the server code are written by the same author, but not for any scenario with potentially hostile clients.

The best way to defend against all these problems is to use a separate POSIX process for the interpreter. Of course, that has overheads (no sharing of memory) but the interpreter starts very quickly, and you can easily bound the time (by having the parent kill it after a timeout) and memory (using ulimit) of the process, and if it crashes or fatals you can simply clean up and move on. 

This is a great question----I should add this to a list of FAQs.

tv42:
Related and/or duplicate: canceling execution via a Context (or other mechanism):
https://github.com/google/starlark-go/issues/236

alandonovan:
FWIW, we are likely to add support for per-thread limits on virtual instruction counts, so an application can bound the number of abstract computation steps done by a program. The (unspecified) measure won't exactly correlate with CPU time, but it has the virtue of being deterministic and reproducible.

PR #298, opened 2020-07-24
starlark: support thread cancellation and limits on computation

alandonovan:
This change adds two related features:
(1) asynchronous cancellation of Starlark threads (Thread.Cancel), and
(2) recording and limiting of the number of abstract computation steps.

Fixes issue #252
Fixes issue #160

### DIFF
--- a/starlark/eval.go
+++ b/starlark/eval.go
@@ -13,9 +13,11 @@ import (
 	"math/big"
 	"sort"
 	"strings"
+	"sync/atomic"
 	"time"
 	"unicode"
 	"unicode/utf8"
+	"unsafe"
 
 	"go.starlark.net/internal/compile"
 	"go.starlark.net/internal/spell"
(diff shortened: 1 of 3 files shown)


starlark/eval.go
```diff
@@ -46,6 +48,21 @@ type Thread struct {
...
+	// this thread. It is incremented by the interpreter. If after
+	// being incremented, its value is zero, the interpreter calls
+	// thread.Cancel("too many steps").
+	//
+	// Steps may be used as a measure the approximate cost of
+	// Starlark execution, by computing the difference in its value
+	// before and after a computation. It may also be used to limit
+	// a computation to k steps by setting its initial value to
+	// 1<<64 - k.
+	Steps uint64
```
jayconrod, line 61: No change needed necessarily, just thinking about alternatives:

When I worked on V8, they had a counter which was set to some value, then decremented at function returns and loop back branches. If it reached zero, the profiler was notified, which could trigger optimization.

I think it's a little more intuitive to have a decrementing counter, since you can initially set it to `k` rather than `1<<64 - k`. But then measuring how many steps were taken is harder, so perhaps there's no advantage either way.

In any case, I wonder if it would be better to provide methods to get / set this, rather than exporting the counter itself. There are a lot of uses and possible representations for a counter like this, and it could easily change in the future.

starlark/interp.go
```diff
@@ -85,6 +87,15 @@ func (fn *Function) CallInternal(thread *Thread, args Tuple, kwargs []Tuple) (Va
 	code := f.Code
 loop:
 	for {
+		thread.Steps++
+		if thread.Steps == 0 {
+			thread.Cancel("too many steps")
+		}
+		if reason := atomic.LoadPointer((*unsafe.Pointer)(unsafe.Pointer(&thread.cancelReason))); reason != nil {
```
jayconrod, line 94: The atomic load, compare, and branch on every iteration seems expensive.

`Cancel` could set `Steps` to `^0`, but I suppose an atomic load would still be needed.

That could be optimized by keeping a local step count and updating the thread's step count on certain instructions (on builtin, return, or back branch) or when the local step count crosses some constant threshold.

No need to optimize that now, but it may require changing the contract of `Steps` in the future. Better to keep it abstract or vague.

starlark/eval.go
```diff
@@ -46,6 +48,21 @@ type Thread struct {
...
+	// Steps is a count of abstract computation steps executed by
+	// this thread. It is incremented by the interpreter. If after
+	// being incremented, its value is zero, the interpreter calls
+	// thread.Cancel("too many steps").
+	//
+	// Steps may be used as a measure the approximate cost of
+	// Starlark execution, by computing the difference in its value
+	// before and after a computation. It may also be used to limit
+	// a computation to k steps by setting its initial value to
+	// 1<<64 - k.
```
jayconrod, line 60: Consider adding a note that the number of steps taken by a given computation is an implementation detail and may change.

starlark/interp.go
alandonovan: I measured it on x86 and it wasn't noticeable (sorry I didn't keep the data). The address is always in L1 cache, the branch is extremely predictable, and on Intel at least, every load is an atomic load. (The atomic load is much cheaper than selecting on the Done channel of a context: even though fundamentally that too boils down to an atomic load, compare, and branch, the function call overheads are significant.)

I considered using a local variable but it complicates the logic, and I suspect it could increase register pressure in the rest of the bytecode interpreter.

PR #299, opened 2020-07-31
spec: clarify scope rules for comprehensions

alandonovan:
- first loop operand is resolved in enclosing environment
- all loop vars are bound before any expression is resolved

See https://github.com/bazelbuild/starlark/issues/84


alandonovan:
@laurentlb

PR #301, opened 2020-08-04
Add parse test for genrule with docstring

pcj:
This (draft) PR is not intended to be merged.

Rather, I'm trying to illustrate what I am doing wrong here.  I was under the impression that a docstring at the beginning of a rule was permitted by the language.  However, parsing a rule-like expression statement with a docstring causes a parse failure.  Is this not supported by the language?


pcj:
Closing this as it's almost certainly working as intended; I suppose a docstring inside a rule like this makes no sense; from a syntactic perspective it appears as a string literal argument to the call expression, followed by a binary expression.

PR #305, opened 2020-09-26
Add special case error for slice index, which overflows int

### LINKED ISSUE #169: starlark: confusing error for giant range parameter
```
$ starlark -c "print(list(range(1)[::4503599627370496]))"
  cmdline:1: in <toplevel>
Error: got int for slice step, want int
```

I presume it is trying to tell me that 4503599627370496 doesn't fit in a uint32.

tdakkota:
Fixes #169.

### DIFF
--- a/starlark/eval.go
+++ b/starlark/eval.go
@@ -1163,7 +1163,7 @@ func slice(x, lo, hi, step_ Value) (Value, error) {
 		var err error
 		step, err = AsInt32(step_)
 		if err != nil {
-			return nil, fmt.Errorf("got %s for slice step, want int", step_.Type())
+			return nil, fmt.Errorf("invalid slice step: %s", err)
 		}
 		if step == 0 {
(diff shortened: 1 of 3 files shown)


alandonovan (commented):
AsInt32 already returns an error that says "out of range" if that is the cause (and reports the actual type if it was not int), so the correct fix here is to use the error it returns. Let's change this to:
```
return nil, fmt.Errorf("invalid slice step: %s", err)
```
I just audited and found two more occurrences of the same problem in 'chr', 'asIndex', and 'slice'. Could you change them to:
```
       return nil, fmt.Errorf("chr: %s", err)
```
and:
```
        return fmt.Errorf("invalid index: %s", err)      
```
respectively?

Thanks.

tdakkota:
Done.
Also, I changed `%s` to `%w`.

alandonovan (approved):
Thank you!

PR #313, opened 2020-11-02
starlark: bring floating-point into spec compliance

alandonovan:
This change makes go.starlark.net's floating-point implementation
match the proposed spec (see https://github.com/bazelbuild/starlark/pull/119),
and thus much more closely match the behavior of the Java implementation.

The major changes are:
- Float values are totally ordered; NaN compares greater than +Inf.
- The string form of a finite float value always contains an exponent
  or a decimal point, so they are self-evidently not int values.
- Operations that would cause a large integer to become rounded to
  an infinite float are now an error.

The resolve.AllowFloat boolean, and the corresponding -float command-line
flag, now have no effect. Floating point support is always enabled.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,5 +6,5 @@ require (
 	github.com/chzyer/logex v1.1.10 // indirect
 	github.com/chzyer/readline v0.0.0-20180603132655-2972be24d48e
 	github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1 // indirect
-	golang.org/x/sys v0.0.0-20200625212154-ddb9806d33ae // indirect
+	golang.org/x/sys v0.0.0-20200803210538-64077c9b5642 // indirect
 )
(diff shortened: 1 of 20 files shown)


go.mod
```diff
@@ -3,8 +3,37 @@ module go.starlark.net
 go 1.13
 
 require (
+	github.com/NebulousLabs/go-upnp v0.0.0-20181203152547-b32978b8ccbf // indirect
```
jayconrod, line 6: Lot of new dependencies. Run `go mod tidy`?

jayconrod (approved):
Looks good except for extras in `go.mod` and `go.sum`. Sorry for the delay.

go.mod
alandonovan: Done.

alandonovan: Thanks!

PR #321, opened 2020-11-18
starlark: use portable syscall wrapper for mmap

### LINKED ISSUE #320: Cannot cross-compile for Solaris
When trying to build a project that depends on starlark-go with `GOOS=solaris`, compiler throws the following error:

```
/go/pkg/mod/go.starlark.net@v0.0.0-20201014215153-dff0ae5b4820/starlark/int_posix64.go:70:12: undefined: syscall.Mmap
```

Is there a reason why `golang.org/x/sys` (which would otherwise provide the `Mmap()` implementation) was ditched in favor to a in-house implementation in https://github.com/google/starlark-go/pull/287?

alandonovan:
Also, add test to ensure that new dependencies (such as golang.org/x/sys/unix)
are not added casually.

Fixes #320

### DIFF
--- a/starlark/eval_test.go
+++ b/starlark/eval_test.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"fmt"
 	"math"
+	"os/exec"
 	"path/filepath"
 	"sort"
 	"strings"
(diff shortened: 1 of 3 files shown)


starlark/eval_test.go
```diff
@@ -890,3 +891,25 @@ func TestExecutionSteps(t *testing.T) {
...
 	}
 }
+
+// TestDeps fails if the interpreter proper (not the REPL, etc) sprouts new external dependencies.
+// We may expand the list of permitted dependencies, but should do so deliberately, not casually.
+func TestDeps(t *testing.T) {
+	cmd := exec.Command("go", "list", "-deps", ".")
+	out, err := cmd.Output()
+	if err != nil {
+		t.Fatalf("'go list starlark' failed: %s", err)
```
jayconrod, line 901: Maybe `Skipf` here instead? If the test is cross-compiled and run in an environment without the Go toolchain, it probably shouldn't fail because of this.

jayconrod (approved):

starlark/eval_test.go
alandonovan: Done---although now I worry about the check failing silently.

PR #325, opened 2020-12-03
ci: Add support for go 1.15.x to .travis.yml

pims:
Also drops testing against go 1.12.x

It might be worth adding support for Github Actions down the road to reduce the chance of a missed "check" on PRs in the future.

For now, testing with a recent go version would be nice.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,9 @@ language: go
 go_import_path: go.starlark.net
 
 go:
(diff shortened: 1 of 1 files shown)


pims:
@googlebot I signed it!

alandonovan (approved):

alandonovan:
Thanks!

PR #327, opened 2020-12-04
time: add time module

b5:
Signed-off-by: b5 <sparkle_pony_2000@qri.io>

A few notes:
* I've added this to `lib/time` based on @alandonovan's feedback from #326
* we're waiting on a chat with @srebhan about the naming of the constructors vs. parsing functions before this is final
* We use a [tool](https://github.com/b5/outline) to extract documentation from this package in `doc.go` into other formats. I can remove this to conform to other packages, but if this info can stay & be maintained, I'd be deeply appreciated. Keeping documentation to the code itself has made our lives much easier.
* If we merge this I'll take this back to our contributors on the starlib project & chat about relying on this package upstream. @alandonovan in the  past you'd mentioned interest in other packages from the project, now might be the time to go shopping 😉


b5:
@googlebot I signed it!

lib/time/time.go
```diff
@@ -0,0 +1,454 @@
...
+	dur, ok := x.Int64()
+	if !ok {
+		return starlark.None, fmt.Errorf("invalid sleep value")
+	}
+
+	gotime.Sleep(gotime.Duration(dur))
+	return starlark.None, nil
+}
+
+func time(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
```
alandonovan, line 98: parseTime / parse_time

lib/time/time.go
```diff
@@ -0,0 +1,454 @@
...
+	var (
+		x, location starlark.String
+		format      = starlark.String(gotime.RFC3339)
```
alandonovan, line 101: format = gotime.RFC3339

(UnpackArgs accepts *string)

lib/time/time.go
```diff
@@ -0,0 +1,454 @@
...
+		return nil, err
+	}
+	t, err := gotime.ParseInLocation(format.String(), x.String(), loc)
+	if err != nil {
+		return nil, err
+	}
+	return Time(t), nil
+}
+
+func fromtimestamp(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
```
alandonovan, line 126: "timestamp "is insufficiently precise: it's a 32-bit number of seconds since 1970, aka a UNIX timestamp. 

from_unixtime ?

lib/time/time.go
```diff
@@ -0,0 +1,454 @@
...
+	t := gotime.Unix(i, 0)
+
+	return Time(t), nil
+}
+
+func now(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	return Time(NowFunc()), nil
+}
+
+// Duration is a starlark representation of a duration
```
alandonovan, line 148: Period at end of sentence (throughout).

lib/time/time.go
```diff
@@ -0,0 +1,454 @@
...
+type Duration gotime.Duration
+
+// String implements the Stringer interface
+func (d Duration) String() string { return gotime.Duration(d).String() }
+
+// Type returns a short string describing the value's type.
+func (d Duration) Type() string { return "duration" }
```
alandonovan, line 155: "time.duration"

lib/time/time.go
```diff
@@ -0,0 +1,454 @@
...
+
+// Freeze renders Duration immutable. required by starlark.Value interface
+// because duration is already immutable this is a no-op
+func (d Duration) Freeze() {}
+
+// Hash returns a function of x such that Equals(x, y) => Hash(x) == Hash(y)
+// required by starlark.Value interface
+func (d Duration) Hash() (uint32, error) { return hashString(d.String()), nil }
```
alandonovan, line 163: String.Hash is expensive; call Int.Hash.

lib/time/time.go
```diff
@@ -0,0 +1,454 @@
...
+
+// Truth returns the truth value of an object required by starlark.Value interface
+func (d Duration) Truth() starlark.Bool { return d > 0 }
```
alandonovan, line 166: d != 0.  It's a signed quantity (though I'm not sure exactly what a negative duration means).

lib/time/time.go
```diff
@@ -0,0 +1,454 @@
...
+}
+
+var durationMethods = map[string]builtinMethod{
+	"hours":       durationhours,
+	"minutes":     durationminutes,
+	"seconds":     durationseconds,
+	"nanoseconds": durationnanoseconds,
+}
+
+func durationhours(fnname string, recV starlark.Value, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
```
alandonovan, line 227: durationHours etc

lib/time/time.go
```diff
@@ -0,0 +1,454 @@
...
+	default:
+		return nil, nil
+	}
+
+	switch op {
+	case syntax.PLUS:
+		return Duration(x + y), nil
+	case syntax.MINUS:
+		return Duration(x - y), nil
+	case syntax.SLASH:
```
alandonovan, line 208: Also define SLASHSLASH.

lib/time/time.go
```diff
@@ -0,0 +1,454 @@
...
+	return builtinAttr(d, name, durationMethods)
+}
+
+// AttrNames lists available dot expression strings. required by starlark.HasAttrs interface
+func (d Duration) AttrNames() []string { return builtinAttrNames(durationMethods) }
+
+// Binary implements binary operators, which satisfies the starlark.HasBinary interface
+func (d Duration) Binary(op syntax.Token, yV starlark.Value, side starlark.Side) (starlark.Value, error) {
+	x := gotime.Duration(d)
+	var y gotime.Duration
```
alandonovan, line 179: It's much clearer to switch on the side and op before the operand types.

lib/time/time.go
```diff
@@ -0,0 +1,454 @@
...
+}
+
+// Time is a starlark representation of a point in time
+type Time gotime.Time
+
+// String implements the Stringer interface
+func (t Time) String() string { return gotime.Time(t).String() }
+
+// Type returns a short string describing the value's type.
+func (t Time) Type() string { return "time" }
```
alandonovan, line 254: time.time

lib/time/time.go
```diff
@@ -0,0 +1,454 @@
...
+
+// Freeze renders time immutable. required by starlark.Value interface
+// because Time is already immutable this is a no-op
+func (t Time) Freeze() {}
+
+// Hash returns a function of x such that Equals(x, y) => Hash(x) == Hash(y)
+// required by starlark.Value interface
+func (t Time) Hash() (uint32, error) { return hashString(t.String()), nil }
```
alandonovan, line 262: int hash

lib/time/time.go
```diff
@@ -0,0 +1,454 @@
...
+		}
+	case Time:
+		y := gotime.Time(yV.(Time))
+		switch op {
+		// time - time = duration
+		case syntax.MINUS:
+			if side == starlark.Left {
+				return Duration(x.Sub(y)), nil
+			}
+			return Duration(y.Sub(x)), nil
```
alandonovan, line 309: Use if { return E1 } else { return E2 }  when the two E expressions have parallel structure.

(It's fine to omit the second else if the first case is an early return.)

lib/time/time.go
```diff
@@ -0,0 +1,454 @@
...
+	"nanosecond": timenanosecond,
+	"unix":       timeunix,
+	"unix_nano":  timeunixnano,
+
+	"in_location": timein,
+	"format":      timeformat,
+}
+
+// TODO - consider using a higher order function to generate these
+func timeyear(fnname string, recV starlark.Value, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
```
alandonovan, line 333: timeYear

lib/time/time.go
```diff
@@ -0,0 +1,454 @@
...
+func builtinAttrNames(methods map[string]builtinMethod) []string {
+	names := make([]string, 0, len(methods))
+	for name := range methods {
+		names = append(names, name)
+	}
+	sort.Strings(names)
+	return names
+}
+
+// hashString computes the FNV hash of s.
```
alandonovan, line 426: Not needed.

alandonovan (commented):
Just a first pass. Thanks for contributing!

srebhan:
@b5: Thanks for your quick reaction! I have `duration()` and `time()` as "constructors" and use `parse_{time,duration}()` as parsing function. I think having the "class-name" as constructors is a matter of least-surprise and the the name for the parsing functions are obvious. However, I see that this would crush your current API...
Is there anything I can do to speed this up?

@alandonovan: I'm supporting this PR and will close mine once this one is merged.

lib/time/time.go
```diff
@@ -4,40 +4,44 @@ import (
...
 const ModuleName = "time"
 
 // Module time is a Starlark module of time-related functions.
 var Module = &starlarkstruct.Module{
 	Name: ModuleName,
 	Members: starlark.StringDict{
-		"duration":      starlark.NewBuiltin("duration", duration),
-		"location":      starlark.NewBuiltin("location", location),
+		"duration":      starlark.NewBuiltin("duration", parseDuration),
+		"location":      starlark.NewBuiltin("location", parseLocation),
```
srebhan, line 23: Shouldn't this be `"parseDuration":` and `"parseLocation":`?

lib/time/time.go
```diff
@@ -4,40 +4,44 @@ import (
...
 		"now":           starlark.NewBuiltin("now", now),
 		"struct":        starlark.NewBuiltin("struct", starlarkstruct.Make),
-		"time":          starlark.NewBuiltin("time", time),
+		"time":          starlark.NewBuiltin("time", parseTime),
```
srebhan, line 26: Same here `"parseTime":`?

lib/time/time.go
```diff
@@ -315,87 +333,87 @@ func (t Time) Binary(op syntax.Token, yV starlark.Value, side starlark.Side) (st
...
+	"second":     timeSecond,
+	"nanosecond": timeNanosecond,
+	"unix":       timeUnix,
+	"unix_nano":  timeUnixNano,
 
-	"in_location": timein,
-	"format":      timeformat,
+	"in_location": timeIn,
+	"format":      timeFormat,
 }
```
srebhan: Just a question probably to @adonovan: Shouldn't these be attributes rather than builtins?

lib/time/time.go
```diff
@@ -0,0 +1,462 @@
...
+	}
+
+	time.Sleep(time.Duration(dur))
+	return starlark.None, nil
+}
+
+func parseTime(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var (
+		x, location starlark.String
+		format      = starlark.String(time.RFC3339)
```
alandonovan, line 102: format = time.RFC3339

(UnpackArgs accepts *string)

lib/time/time.go
```diff
@@ -0,0 +1,462 @@
...
+	)
+	if err := starlark.UnpackArgs("time", args, kwargs, "x", &x, "format?", &format, "location", &location); err != nil {
+		return nil, err
+	}
+
+	if location == "" {
+		t, err := time.Parse(format.String(), x.String())
```
alandonovan, line 109: format.String() looks incorrect if format has type starlark.String as it returns a string literal, as if by repr(x).

lib/time/time.go
```diff
@@ -0,0 +1,462 @@
...
+		return starlark.None, fmt.Errorf("invalid sleep value")
+	}
+
+	time.Sleep(time.Duration(dur))
+	return starlark.None, nil
+}
+
+func parseTime(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var (
+		x, location starlark.String
```
alandonovan, line 101: var x, location string

lib/time/time.go
```diff
@@ -0,0 +1,462 @@
...
+
+func fromUnixTime(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var (
+		x starlark.Int
+	)
+	if err := starlark.UnpackArgs("time", args, kwargs, "x", &x); err != nil {
+		return nil, err
+	}
+
+	i, err := strconv.ParseInt(x.String(), 10, 64)
```
alandonovan, line 135: x is already an int; there's no need to format and parse it.

var x int
if err := starlark.UnpackArgs("time", args, kwargs, "x", &x); err != nil { ... }
return Time(time.Unix(x, 0)), nil

lib/time/time.go
```diff
@@ -0,0 +1,462 @@
...
+// Type returns a short string describing the value's type.
+func (d Duration) Type() string { return "time.duration" }
+
+// Freeze renders Duration immutable. required by starlark.Value interface
+// because duration is already immutable this is a no-op.
+func (d Duration) Freeze() {}
+
+// Hash returns a function of x such that Equals(x, y) => Hash(x) == Hash(y)
+// required by starlark.Value interface.
+func (d Duration) Hash() (uint32, error) { return starlark.MakeInt64(int64(d)).Hash() }
```
alandonovan, line 164: There's no need to allocate an Int here. You can simply combine the low and high uint32s of the int64 using xor or similar.

lib/time/time.go
```diff
@@ -0,0 +1,462 @@
...
+func (d Duration) Attr(name string) (starlark.Value, error) {
+	return builtinAttr(d, name, durationMethods)
+}
+
+// AttrNames lists available dot expression strings. required by
+// starlark.HasAttrs interface.
+func (d Duration) AttrNames() []string { return builtinAttrNames(durationMethods) }
+
+// Binary implements binary operators, which satisfies the starlark.HasBinary
+// interface.
```
alandonovan, line 181: A summary of the supported operations here would help, as it's hard to tell from the implementation

```
int * duration
duration * int
etc
```

lib/time/time.go
```diff
@@ -0,0 +1,462 @@
...
+func (d Duration) Binary(op syntax.Token, yV starlark.Value, _ starlark.Side) (starlark.Value, error) {
+	x := time.Duration(d)
+	var y time.Duration
+	switch yV.(type) {
+	case starlark.Int:
+		i, ok := yV.(starlark.Int).Int64()
+		if !ok {
+			return nil, fmt.Errorf("duration binary operation: couldn't parse int")
```
alandonovan, line 189: "int value out of range (want signed 64-bit value)"

lib/time/time.go
```diff
@@ -0,0 +1,462 @@
...
+func (d Duration) Hash() (uint32, error) { return starlark.MakeInt64(int64(d)).Hash() }
+
+// Truth returns the truth value of an object required by starlark.Value
+// interface.
+func (d Duration) Truth() starlark.Bool { return d != 0 }
+
+// Attr gets a value for a string attribute, implementing dot expression support
+// in starklark. required by starlark.HasAttrs interface.
+func (d Duration) Attr(name string) (starlark.Value, error) {
+	return builtinAttr(d, name, durationMethods)
```
alandonovan, line 173: Methods are appropriate when there is potential computation or failure, but for the component accessors of a duration or time, fields are more appropriate.

lib/time/time.go
```diff
@@ -0,0 +1,462 @@
...
+
+func durationnanoseconds(fnname string, recV starlark.Value, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	recv := time.Duration(recV.(Duration))
+	return starlark.MakeInt64(recv.Nanoseconds()), nil
+}
+
+// Time is a starlark representation of a point in time.
+type Time time.Time
+
+// String implements the Stringer interface.
```
alandonovan, line 260: Describe what the function does, not what interface it implements; specifically, the format of the resulting string.

lib/time/time.go
```diff
@@ -0,0 +1,462 @@
...
+func (t Time) String() string { return time.Time(t).String() }
+
+// Type returns a short string describing the value's type.
```
alandonovan, line 263: // Type returns "time.time".

lib/time/time.go
```diff
@@ -0,0 +1,462 @@
...
+func (t Time) Type() string { return "time.time" }
+
+// Freeze renders time immutable. required by starlark.Value interface
+// because Time is already immutable this is a no-op.
+func (t Time) Freeze() {}
+
+// Hash returns a function of x such that Equals(x, y) => Hash(x) == Hash(y)
+// required by starlark.Value interface.
+func (t Time) Hash() (uint32, error) {
+	return starlark.MakeInt64(time.Time(t).UnixNano()).Hash()
```
alandonovan, line 273: Ditto: no need to allocate an Int.

lib/time/time.go
```diff
@@ -0,0 +1,462 @@
...
+		"nanosecond":  durInt(time.Nanosecond),
+		"microsecond": durInt(time.Microsecond),
+		"millisecond": durInt(time.Millisecond),
+		"second":      durInt(time.Second),
+		"minute":      durInt(time.Minute),
+		"hour":        durInt(time.Hour),
+	},
+}
+
+func durInt(d time.Duration) starlark.Int {
```
alandonovan, line 41: Not needed.

lib/time/time.go
```diff
@@ -0,0 +1,462 @@
...
+		return nil, err
+	}
+	return Time(t), nil
+}
+
+func fromUnixTime(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var (
+		x starlark.Int
+	)
+	if err := starlark.UnpackArgs("time", args, kwargs, "x", &x); err != nil {
```
alandonovan, line 131: We should use UnpackPositionalArgs for all these functions unless there is a compelling reason to expose the parameter names.

lib/time/time.go
```diff
@@ -0,0 +1,462 @@
...
+	t := time.Unix(i, 0)
+
+	return Time(t), nil
+}
+
+func now(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	return Time(NowFunc()), nil
+}
+
+// Duration is a starlark representation of a duration.
```
alandonovan, line 149: Starlark (capital S)

This is the place---not the top of this file---to document the data type: what it represents, what operations it supports, and what Starlark attributes it has.

lib/time/time.go
```diff
@@ -0,0 +1,462 @@
...
+// AttrNames lists available dot expression strings. required by
+// starlark.HasAttrs interface.
+func (d Duration) AttrNames() []string { return builtinAttrNames(durationMethods) }
+
+// Binary implements binary operators, which satisfies the starlark.HasBinary
+// interface.
+func (d Duration) Binary(op syntax.Token, yV starlark.Value, _ starlark.Side) (starlark.Value, error) {
+	x := time.Duration(d)
+	var y time.Duration
+	switch yV.(type) {
```
alandonovan, line 185: It is much clearer to switch on op first, then y.(type) in each case.

srebhan:
@b5 any news on this one? I'm eagerly waiting for time support! ;-P

b5:
hi all, 

I still have a little more cleanup to do here, but wanted to get this set of changes up for review. I've made a bunch of adjustments to the Starlark API based on review feedback.

@srebhan, what do you think the signature for the `time.time` constructor should be? I've currently got it wrapping golang's `time.Date`, but could also see removing `time.from_timestamp` and using that instead.

srebhan:
@b5 I had `time([year[, month[, day[, hour[, minute[, second[, nanosecond[, location]]]]]]]])` to mimic the python API a bit, but I think a `kwargs` approach also works. Whatever you prefer to be honest! :-)

lib/time/doc.go
```diff
@@ -0,0 +1,56 @@
...
+  outline: time
+    time defines time primitives for starlark
+    path: time
+    functions:
+      duration(string) duration
+        parse a duration
+      location(string) location
+        parse a location
+      time(string, format=..., location=...) time
+        parse a time
```
srebhan, line 13: I think this needs an update.

lib/time/doc.go
```diff
@@ -0,0 +1,56 @@
...
+      fromtimestamptime(int) time
+        parse a Unix timestamp
+      now() time
+        implementations would be able to make this a constant
+      zero() time
+        a constant
```
srebhan, line 19: This should be the result of `time.time()` IMO.

lib/time/doc.go
```diff
@@ -0,0 +1,56 @@
...
+          duration < duration = booleans
+      time
+        functions:
+          year() int
+          month() int
+          day() int
+          hour() int
+          minute() int
+          second() int
+          nanosecond() int
```
srebhan, line 41: Those are attributes now I guess.

lib/time/time.go
```diff
@@ -0,0 +1,548 @@
...
+	Members: starlark.StringDict{
+		"duration":       starlark.NewBuiltin("duration", newDuration),
+		"parse_duration": starlark.NewBuiltin("parse_duration", parseDuration),
+		"parse_location": starlark.NewBuiltin("parse_location", parseLocation),
+		"now":            starlark.NewBuiltin("now", now),
+		"struct":         starlark.NewBuiltin("struct", starlarkstruct.Make),
+		"time":           starlark.NewBuiltin("time", newTime),
+		"parse_time":     starlark.NewBuiltin("time", parseTime),
+		"sleep":          starlark.NewBuiltin("sleep", sleep),
+		"from_timestamp": starlark.NewBuiltin("fromtimestamp", fromUnixTime),
```
srebhan, line 29: Maybe you should name the called function `fromTimestamp` instead of `fromUnixTime` to keep the naming pattern.

lib/time/time.go
```diff
@@ -0,0 +1,548 @@
...
+	if err := starlark.UnpackArgs("location", args, kwargs, "s", &s); err != nil {
+		return nil, err
+	}
+	loc, err := time.LoadLocation(s)
+	if err != nil {
+		return nil, err
+	}
+
+	return starlark.String(loc.String()), nil
+}
```
srebhan, line 72: I don't get the use of this function... You put in a location string and get that string out? Why do you need this?

lib/time/time.go
```diff
@@ -0,0 +1,548 @@
...
+// applications that require their Starlark scripts to be fully deterministic.
+var SleepFunc = time.Sleep
+
+func sleep(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var dur Duration
+	if err := starlark.UnpackPositionalArgs("sleep", args, kwargs, 1, &dur); err != nil {
+		return starlark.None, err
+	}
+
+	SleepFunc(time.Duration(dur))
```
srebhan, line 85: You need to check `SleepFunc` being not nil otherwise we might segfault here.

lib/time/time.go
```diff
@@ -0,0 +1,548 @@
...
+func fromUnixTime(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var x int
+	if err := starlark.UnpackPositionalArgs("from_timestamp", args, kwargs, 1, &x); err != nil {
+		return nil, err
+	}
+	return Time(time.Unix(int64(x), 0)), nil
+}
+
+func now(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	return Time(NowFunc()), nil
```
srebhan, line 126: We need to check `NowFunc` for not being nil to avoid a segfault here.

lib/time/time.go
```diff
@@ -0,0 +1,548 @@
...
+			return Duration(x - time.Duration(y)), nil
+		case starlark.Int:
+			i, ok := y.Int64()
+			if !ok {
+				return nil, fmt.Errorf("int value out of range (want signed 64-bit value)")
+			}
+			return Duration(x - time.Duration(i)), nil
+		case Time:
+			// duration - time = duration
+			return nil, nil
```
srebhan, line 256: You should remove this part as the default return is `nil, nil` in case the operation is not supported.

lib/time/time.go
```diff
@@ -0,0 +1,548 @@
...
+		}
+
+	case syntax.SLASH:
+		switch y := yV.(type) {
+		case Duration:
+			if int64(y) == 0 {
+				return nil, fmt.Errorf("%s division by zero", d.Type())
+			}
+			return d / y, nil
```
srebhan, line 265: Isn't this a self reference? I guess you want a floating-point division here and return a Starlark float!?

lib/time/time.go
```diff
@@ -0,0 +1,548 @@
...
+		case Time:
+			// y := time.Time(yV.(Time))
+			// switch op {
+			// case syntax.PLUS:
+			// 	// duration + time = time
+			// 	return Time(y.Add(x)), nil
+			// case syntax.MINUS:
+			// 	// duration - time = duration
+			// 	return nil, nil
+			// }
```
srebhan, line 284: This is not a valid operation IMO.

lib/time/time.go
```diff
@@ -0,0 +1,548 @@
...
+		}
```
srebhan, line 342: Again I think you can only operate `duration + time` everything else makes not really sense...

lib/time/time.go
```diff
@@ -0,0 +1,548 @@
...
+		year, month, day, hour, min, sec, nsec int
+		loc                                    string
+	)
+	if err := starlark.UnpackArgs("time", args, kwargs, "year", &year, "month", &month, "day", &day, "hour", &hour, "min", &min, "sec", &sec, "nsec", &nsec, "loc", &loc); err != nil {
+		return nil, err
+	}
+	location, err := time.LoadLocation(loc)
+	if err != nil {
+		return nil, err
+	}
```
srebhan, line 364: Can we please make the location optional in the similar to parsing the time!? We probably want to default to local time (`time.Local`) or UTC (`time.UTC`).

lib/time/time.go
```diff
@@ -0,0 +1,548 @@
...
+
+// Time is a starlark representation of a point in time.
+type Time time.Time
+
+func newTime(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var (
+		year, month, day, hour, min, sec, nsec int
+		loc                                    string
+	)
+	if err := starlark.UnpackArgs("time", args, kwargs, "year", &year, "month", &month, "day", &day, "hour", &hour, "min", &min, "sec", &sec, "nsec", &nsec, "loc", &loc); err != nil {
```
srebhan, line 358: Maybe make all args optional to allow for a "null-time".
```suggestion
	if err := starlark.UnpackArgs("time", args, kwargs, "year?", &year, "month?", &month, "day?", &day, "hour?", &hour, "minute?", &min, "second?", &sec, "nanosecond?", &nsec, "location?", &loc); err != nil {
```

lib/time/time.go
```diff
@@ -0,0 +1,548 @@
...
+	return nil, nil
+}
+
+var timeMethods = map[string]builtinMethod{
+	"in_location": timeIn,
+	"format":      timeFormat,
+}
+
+func timeFormat(fnname string, recV starlark.Value, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var x starlark.String
```
srebhan, line 483: IIRC you can directly use string here...

lib/time/time.go
```diff
@@ -0,0 +1,548 @@
...
+		return cmp <= 0
+	case syntax.LT:
+		return cmp < 0
+	case syntax.GE:
+		return cmp >= 0
+	case syntax.GT:
+		return cmp > 0
+	}
+	panic(op)
+}
```
srebhan, line 548: You use this function exactly once in `CompareSameType()`, so maybe you should move the implementation just there.

lib/time/time.go
```diff
@@ -0,0 +1,548 @@
...
+func (t Time) CompareSameType(op syntax.Token, yV starlark.Value, depth int) (bool, error) {
+	x := time.Time(t)
+	y := time.Time(yV.(Time))
+	cp := 0
+	if x.Before(y) {
+		cp = -1
+	} else if x.After(y) {
+		cp = 1
+	}
+	return threeway(op, cp), nil
```
srebhan, line 441: I prefer doing the comparison directly here. But this is just my two cent...
```suggestion
	switch op {
￼	case syntax.EQL:
￼		return x.Equal(y), nil
￼	case syntax.NEQ:
￼		return !x.Equal(y), nil
￼	case syntax.LT:
￼		return x.Before(y), nil
￼	case syntax.LE:
￼		return !x.After(y), nil
￼	case syntax.GT:
￼		return x.After(y), nil
￼	case syntax.GE:
￼		return !x.Before(y), nil
￼	}

	return false, errors.New("operation not supported")
￼￼```

File: lib/time/time.go (hunk 38 above)
Comment (b5) [reply lib/time/time.go:72]: this validates system clock time zone locations, users can use it to both confirm the timezone in question is supported by the OS. In the future This could be expanded to return a starlark "location" type that supports timezone conversion

File: lib/time/time.go (hunk 39 above)
Comment (b5) [reply lib/time/time.go:85]: Personally I think that if users set `SleepFunc` to nil and try to call `SleepFunc` the program should panic.

File: lib/time/doc.go [hunk 50]
```diff
@@ -0,0 +1,56 @@
...
+
+    types:
+      duration
+        fields:
+          hours float
+          minutes float
+          nanoseconds int
+          seconds float
+        operators:
+          duration - time = duration
```
srebhan, line 29: IMO this is not really a sensible operation.

lib/time/doc.go
srebhan: Please mark as resolved.

lib/time/doc.go
srebhan: Please mark as resolved.

lib/time/time.go
srebhan: Please mark as resolved.

lib/time/time.go
srebhan: Can we then please name the function `validate_location()` or similar and return a bool!? Otherwise it is very confusing.

lib/time/time.go
srebhan: Well I don't like panicking on this level. Imagine you expose setting the sleep function to the user script (for whatever reason). Now your whole program will explode without any way for the programmer to intervene. IMO we should either throw an error here or use the default sleep function in presence of `SleepFunc` being `nil`.

lib/time/time.go
```diff
@@ -0,0 +1,515 @@
...
+	SleepFunc(time.Duration(dur))
+	return starlark.None, nil
+}
+
+func parseTime(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var (
+		x, location string
+		format      = time.RFC3339
+	)
+	if err := starlark.UnpackArgs("time", args, kwargs, "x", &x, "format?", &format, "location", &location); err != nil {
```
srebhan, line 94: I think this should be
```suggestion
	if err := starlark.UnpackArgs("time", args, kwargs, "x", &x, "format?", &format, "location?", &location); err != nil {
```
to also make `location` optional, shouldn't it?

lib/time/time.go
alandonovan: I agree with @b5 that it's fine to panic here, as it's an application programming error.

lib/time/time.go
```diff
@@ -0,0 +1,515 @@
...
+	}
+
+	return fmt.Errorf("cannot convert %s to %s", v.Type(), d.Type())
+}
+
+// String implements the Stringer interface.
+func (d Duration) String() string { return time.Duration(d).String() }
+
+// Type returns a short string describing the value's type.
+func (d Duration) Type() string { return "time.duration" }
```
srebhan, line 173: Just in case someone changes the `ModuleName` we should do
```suggestion
func (d Duration) Type() string { return ModuleName+".duration" }
```

lib/time/time.go
```diff
@@ -0,0 +1,515 @@
...
+}
+
+// Binary implements binary operators, which satisfies the starlark.HasBinary
+// interface. operators:
+//    duration + duration = duration
+//    duration + int = duration
+//    duration + time = time
+//    duration - duration = duration
+//    duration - int = duration
+//    duration - time = duration
```
srebhan, line 223: I don't think this is a sensible operation.

lib/time/time.go
```diff
@@ -0,0 +1,515 @@
...
+		case Duration:
+			return Duration(x - time.Duration(y)), nil
+		case starlark.Int:
+			i, ok := y.Int64()
+			if !ok {
+				return nil, fmt.Errorf("int value out of range (want signed 64-bit value)")
+			}
+			return Duration(x - time.Duration(i)), nil
+		case Time:
+			// duration - time = duration
```
srebhan, line 255: Please remove.

lib/time/time.go
```diff
@@ -0,0 +1,515 @@
...
+				return nil, fmt.Errorf("%s division by zero", d.Type())
+			}
+			return Duration(d / Duration(i)), nil
+		}
+
+		// if int64(y) == 0 {
+		// 	return nil, fmt.Errorf("%s division by zero", d.Type())
+		// }
+		// return Duration(x / y), nil
+
```
srebhan, line 280: Remove.

lib/time/time.go
```diff
@@ -0,0 +1,515 @@
...
+			if int64(y) == 0 {
+				return nil, fmt.Errorf("%s division by zero", d.Type())
+			}
+			return d / y, nil
+		case starlark.Int:
+			i, ok := yV.(starlark.Int).Int64()
+			if !ok {
+				return nil, fmt.Errorf("int value out of range (want signed 64-bit value)")
+			}
+			return d / Duration(i), nil
```
srebhan, line 293: Check for division by zero (`i==0`).

lib/time/time.go
```diff
@@ -0,0 +1,515 @@
...
+		}
+
+		// if int64(y) == 0 {
+		// 	return nil, fmt.Errorf("time.duration floored divison by zero")
+		// }
+		// return Duration(x / y), nil
+
```
srebhan, line 300: Remove.

lib/time/time.go
```diff
@@ -0,0 +1,515 @@
...
+		return nil, err
+	}
+	return Time(time.Date(year, time.Month(month), day, hour, min, sec, nsec, location)), nil
+}
+
+// String implements the Stringer interface.
+func (t Time) String() string { return time.Time(t).String() }
+
+// Type returns "time.time".
+func (t Time) Type() string { return "time.time" }
```
srebhan, line 339: Same as for duration
```suggestion
func (t Time) Type() string { return ModuleName+".time" }
```

lib/time/time.go
```diff
@@ -0,0 +1,515 @@
...
+	x := time.Time(t)
+
+	switch op {
+	case syntax.PLUS:
+		switch y := yV.(type) {
+		case Duration:
+			return Time(x.Add(time.Duration(y))), nil
+		case Time:
+			return nil, fmt.Errorf("cannot add %s to %s", t.Type(), yV.Type())
+		}
```
srebhan, line 426: Not sure if we need to return this error here or `nil, nil` by removing this part...

srebhan:
@b5 thanks for all your work! Walked through the code once again, but IMO we are very close. However I've no say here. ;-)

lib/time/time.go
```diff
@@ -0,0 +1,548 @@
...
+}
+
+func now(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	return Time(NowFunc()), nil
+}
+
+// Duration is a Starlark representation of a duration.
+type Duration time.Duration
+
+func newDuration(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
```
alandonovan, line 132: This function looks unnecessary. It merely asserts that its argument is a Duration, and returns it unchanged.

lib/time/time.go
```diff
@@ -0,0 +1,548 @@
...
+		dur, err := time.ParseDuration(string(x))
+		if err != nil {
+			return err
+		}
+
+		*d = Duration(dur)
+		return nil
+	}
+
+	return fmt.Errorf("cannot convert %s to %s", v.Type(), d.Type())
```
alandonovan, line 166: "got %s, want a duration, string, or int"

lib/time/time.go
```diff
@@ -0,0 +1,548 @@
...
+// because duration is already immutable this is a no-op.
+func (d Duration) Freeze() {}
+
+// Hash returns a function of x such that Equals(x, y) => Hash(x) == Hash(y)
+// required by starlark.Value interface.
+func (d Duration) Hash() (uint32, error) {
+	return starlark.MakeInt64(int64(d)).Hash()
+}
+
+// Truth returns the truth value of an object required by starlark.Value
```
alandonovan, line 185: // Truth reports whether the duration is non-zero.

lib/time/time.go
alandonovan: It's fine to keep it separate, as it crops up a lot and this makes clear that it's a simply copy.

(The design of CompareSameType is overcomplicated; see  3b7e02ecde8b80808c06c8284ce9f85c101fcebd for context.)

lib/time/time.go
```diff
@@ -0,0 +1,548 @@
...
+//    time - time = duration
+func (t Time) Binary(op syntax.Token, yV starlark.Value, side starlark.Side) (starlark.Value, error) {
+	x := time.Time(t)
+
+	switch op {
+	case syntax.PLUS:
+		switch y := yV.(type) {
+		case Duration:
+			return Time(x.Add(time.Duration(y))), nil
+		case Time:
```
alandonovan, line 424: Delete this case: it's the default behavior (return nil, nil)

lib/time/coverage.txt
```diff
@@ -0,0 +1,149 @@
+mode: set
```
alandonovan, line 1: You needn't commit this file.

lib/time/doc.go
alandonovan: I agree.

lib/time/doc.go
```diff
@@ -0,0 +1,56 @@
+/*Package time defines time primitives for starlark, based heavily on the time
+package from the go standard library.
+
+  outline: time
```
alandonovan, line 4: What is this notation here?

Can we brush up this comment so that it conforms to godoc norms? Ideally with a fixed-width table summarizing the operations, and then a more expansive prose explanation below.

lib/time/testdata/test.star
```diff
@@ -0,0 +1,37 @@
+load('time', 'time')
+load('assert.star', 'assert')
+
+assert.eq(time.parse_time("2011-04-22T13:33:48Z"), time.parse_time("2011-04-22T13:33:48Z"))
+assert.eq(time.zero, time.parse_time("0001-01-01T00:00:00Z"))
+assert.true(time.parse_time("2010-04-22T13:33:48Z") < time.parse_time("2011-04-22T13:33:48Z"))
```
alandonovan, line 6: assert.lt(x, y) gives a better error.

lib/time/testdata/test.star
```diff
@@ -0,0 +1,37 @@
...
+assert.true(time.parse_time("2011-04-22T13:33:48Z") == time.parse_time("2011-04-22T13:33:48Z"))
```
alandonovan, line 7: This one is redundant wrt L4 I think.

lib/time/testdata/test.star
```diff
@@ -0,0 +1,37 @@
...
+assert.true(time.parse_time("2012-04-22T13:33:48Z") > time.parse_time("2011-04-22T13:33:48Z"))
+
+assert.eq(time.parse_time("2020-06-26T17:38:36Z"), time.from_timestamp(1593193116))
+
+# zero
+assert.eq(time.zero.format("Mon Jan 2 15:04:05 -0700 MST 2006"), "Mon Jan 1 00:00:00 +0000 UTC 0001")
```
alandonovan, line 13: Do we need time.zero? It makes sense in Go where a Time struct may be initialized, but that's not the case in Starlark. I would prefer that we eliminate it, and avoid having to reveal anything about the representation.

lib/time/testdata/test.star
```diff
@@ -0,0 +1,37 @@
...
+assert.eq(time.parse_duration("0s") + time.parse_duration("3m35s"), time.parse_duration("3m35s"))
+
+d2 = time.parse_duration("10h")
+assert.eq(10.0, d2.hours)
+assert.eq(10*60.0, d2.minutes)
+assert.eq(10*60*60.0, d2.seconds)
+assert.eq(10*60*60*1000000000, d2.nanoseconds)
+
+
+time.sleep(time.second)
```
alandonovan, line 37: This will make the tests slow. Sleep for a millisecond and then assert that time.now observes at least that much elapsed time.

lib/time/time.go
```diff
@@ -0,0 +1,548 @@
...
+	case syntax.PLUS:
+		switch y := yV.(type) {
+		case Duration:
+			return Duration(x + time.Duration(y)), nil
+		case starlark.Int:
+			i, ok := y.Int64()
+			if !ok {
+				return nil, fmt.Errorf("int value out of range (want signed 64-bit value)")
+			}
+			return Duration(x + time.Duration(i)), nil
```
alandonovan, line 239: I don't think we should allow this operation. Duration * int makes sense, but Duration + int does not. (It does in Go, because Go's time.Duration exposes its representation as an int64 of nanoseconds.) Users can always write `duration + 1 * time.nanosecond`, which is clear and unambiguous.

lib/time/time.go
alandonovan: Agreed.

lib/time/time.go
```diff
@@ -0,0 +1,548 @@
...
+// in starklark. required by starlark.HasAttrs interface.
+func (d Duration) Attr(name string) (starlark.Value, error) {
+	switch name {
+	case "hours":
+		return starlark.Float(time.Duration(d).Hours()), nil
+	case "minutes":
+		return starlark.Float(time.Duration(d).Minutes()), nil
+	case "seconds":
+		return starlark.Float(time.Duration(d).Seconds()), nil
+	case "nanoseconds":
```
alandonovan, line 199: Go has milliseconds and microseconds too (both int64). Any reason we should not?

lib/time/time.go
srebhan: Ok, but this means we never should expose setting `SleepFunc` to the interpreter and thus to the user. To be honest I would make `SleepFunc == nil`  the default and then run `time.Sleep` to circumvent all of this, but it's not my decision. :-)

lib/time/time.go
srebhan: Hmmm if I understand the code correctly it calls the `Unpack()` function of `Duration` below to construct a new duration. All the type handling/conversion/construction etc. is done below, isn't it?
What is the correct way in your opinion to construct a new duration?

lib/time/time.go
srebhan: Ok but then please make this an infrastructure and only ask the new types to compute a comparison between `a` and `b` by returning `-1`, `0` or `+1` indicting if `a<b`, `a==b` or `a>b` and do the `threeway` logic elsewhere? :-)

b5:
@srebhan I know you'd like to get this merged sooner rather than later, and I've got a number of other projects on the go at the moment. I've invited you as a collaborator on our fork, would you mind taking over this branch & making the necessary changes to get this up to speed? If not I'll get these changes from review made, but I can't guarantee they'll be done quickly.

srebhan:
@b5 I talked to my management and they were clear that my mandate only covers contributions to this repository. :-( So I think getting permission to contribute to your repo will take more time than it saves...
This being said, I'm allowed to make suggestion in the comments and you can accept them. Would this be an option?
Alternatively, we might ask @adonovan what is absolutely necessary to get this merged and then I will fix the remaining issues with an additional PR!?

b5:
ah that makes sense. I'll do what I can to make changes from review as soon as I can. Given that we can't directly collaborate, I think it'll be easier to just add another hour or two to clean this up & see if we can't get it merged. I'll try to get find that hour in the next day or two

srebhan:
Thanks a bunch!

lib/time/time.go
b5: great catch!

lib/time/time.go
b5: @alandonovan I have the same question as @srebhan, to me this is a constructor that leverages the `Unpack` interface. Would love advice on an alternative

essobedo:
Hi @b5,  @srebhan asked me to help you on this, please check my contribution to it hoping it is useful https://github.com/qri-io/starlark-go/pull/1

essobedo:
@googlebot I fixed it.

essobedo:
@googlebot I consent.

essobedo:
@alandonovan I think I went through all your remarks but the one about the `doc.go`, could you please clarify a little bit more what you have in mind? If you could check the rest also, it would be great. Thank you in advance 🙏

essobedo:
@alandonovan What do you suggest for `lib/time/doc.go`? Maybe you would rather prefer that we remove it?

srebhan (approved):
The code looks good to me, but `time/doc.go` does not match the code (e.g. year, month,... functions). So if @alandonovan decides we should keep it, it should be fixed.

b5:
I'd vote to remove it. This PR shouldn't introduce a change to documentation conventions. It originated from a different codebase, and we can still maintain this form of documentation within that codebase

essobedo:
@b5 I've just removed it

essobedo:
FYI I renamed the package to `starlarktime` to remain consistent with other modules like the json module

adonovan:
Sorry for radio silence: I'm on vacation at the moment but will attend to
this PR fully on Monday.

starlark/testdata/time.star
```diff
@@ -0,0 +1,119 @@
...
+assert.eq(d * 1000, time.parse_duration("16m40s"))
+assert.fails(lambda: d2 // 0, "division by zero")
+
+
+before = time.now()
+time.sleep(10 * time.millisecond)
+assert.true((time.now() - before).nanoseconds >= time.parse_duration("10ms").nanoseconds)
+
+# time(year=..., month=..., day=..., hour=..., minute=..., second=..., nanosecond=..., location=...)
+t1 = time.time(2009, 6, 12, 12, 6, 10, 99, "US/Eastern")
```
alandonovan, line 79: I suggest we get rid of the positional argument constructor: it seems like a recipe for mistakes, whereas `year=2009, month=6` etc is self-documenting.

starlark/testdata/time.star
```diff
@@ -0,0 +1,119 @@
...
+assert.eq(12, t1.hour)
+assert.eq(6, t1.minute)
+assert.eq(10, t1.second)
+assert.eq(99, t1.nanosecond)
+assert.eq(1244822770, t1.unix)
+assert.eq(1244822770000000099, t1.unix_nano)
+
+# time == time = boolean
+# time != time = boolean
+assert.eq(time.parse_time("2011-04-22T13:33:48Z"), time.parse_time("2011-04-22T13:33:48Z"))
```
alandonovan, line 96: Define some time and duration constants, and refer to them in the code below; it will make the block easier to read.

e.g. d10h = time.parse_duration("10h")

starlarktime/time.go
```diff
@@ -0,0 +1,507 @@
...
+	"fmt"
+	"sort"
+	"time"
+
+	"go.starlark.net/starlark"
+	"go.starlark.net/starlarkstruct"
+	"go.starlark.net/syntax"
+)
+
+// ModuleName defines the expected name for this Module when used in the
```
alandonovan, line 14: I'm not sure we need this. Users can say Module.Name, or just "time".

starlarktime/time.go
```diff
@@ -0,0 +1,507 @@
...
+	}
+	loc, err := time.LoadLocation(s)
+	if err != nil {
+		return nil, err
+	}
+
+	return starlark.String(loc.String()), nil
+}
+
+// SleepFunc is a function that pauses the current goroutine for at least d.
```
alandonovan, line 71: "puts the calling goroutine to sleep for the specified duration"

starlarktime/time.go
```diff
@@ -0,0 +1,507 @@
...
+// Intentionally exported so that it can be overridden, for example by
+// applications that require their Starlark scripts to be fully deterministic.
+var SleepFunc = time.Sleep
+
+func sleep(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var dur Duration
+	if err := starlark.UnpackPositionalArgs("sleep", args, kwargs, 1, &dur); err != nil {
+		return starlark.None, err
```
alandonovan, line 79: return nil, err

starlarktime/time.go
```diff
@@ -0,0 +1,507 @@
...
+// assert at compile time that Duration implements Unpacker.
+var _ starlark.Unpacker = (*Duration)(nil)
+
+// Unpack is a custom argument unpacker
+func (d *Duration) Unpack(v starlark.Value) error {
+	switch x := v.(type) {
+	case Duration:
+		*d = x
+		return nil
+	case starlark.Int:
```
alandonovan, line 138: I don't think we want to implicitly allow ints to be converted to durations in function calls.

starlarktime/time.go
```diff
@@ -0,0 +1,507 @@
...
+// interface. operators:
+//    duration + duration = duration
+//    duration + time = time
+//    duration - duration = duration
+//    duration / duration = float
+//    duration / int = duration
+//    duration // duration = int
+//    duration // int = duration
+//    duration * int = duration
+func (d Duration) Binary(op syntax.Token, yV starlark.Value, _ starlark.Side) (starlark.Value, error) {
```
alandonovan, line 242: You can name this parameter y, and then use `switch y := y.(type)` below. Not only does this avoid the need to invent another name, it prevents inadvertent use of the outer y of the wider type.

starlarktime/time.go
```diff
@@ -0,0 +1,507 @@
...
+		}
+
+	case syntax.SLASH:
+		switch y := yV.(type) {
+		case Duration:
+			if int64(y) == 0 {
+				return nil, fmt.Errorf("%s division by zero", d.Type())
+			}
+			return starlark.Float(float64(x.Nanoseconds()) / float64(time.Duration(y).Nanoseconds())), nil
+		case starlark.Int:
```
alandonovan, line 267: I think (not sure) that you need to inspect the 'side' parameter here to ensure that "int / duration" doesn't spuriously work. Ditto SLASHSLASH.

starlarktime/time.go
```diff
@@ -0,0 +1,507 @@
...
+				return nil, fmt.Errorf("int value out of range (want signed 64-bit value)")
+			}
+			return d * Duration(i), nil
+		}
+	}
+
+	return nil, nil
+}
+
+// Time is a starlark representation of a point in time.
```
alandonovan, line 310: "moment in time"

starlarktime/time.go
```diff
@@ -0,0 +1,507 @@
...
+type Time time.Time
+
+func newTime(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var (
+		year, month, day, hour, min, sec, nsec int
+		loc                                    string
+	)
+	if err := starlark.UnpackArgs("time", args, kwargs, "year?", &year, "month?", &month, "day?", &day, "hour?", &hour, "minute?", &min, "second?", &sec, "nanosecond?", &nsec, "location?", &loc); err != nil {
```
alandonovan, line 318: Put a newline after each pair of arguments.

starlark/testdata/time.star
```diff
@@ -0,0 +1,119 @@
+# Tests of time module.
+
+load('assert.star', 'assert')
+load('time.star', 'time')
+
+assert.eq(time.parse_time("2020-06-26T17:38:36Z"), time.from_timestamp(1593193116))
```
alandonovan, line 6: Let's add some tests of the data types' basic operations:
```
type(t)
str(t)
{t: None} # hashable
```
Ditto for duration.

starlarktime/time.go
```diff
@@ -0,0 +1,507 @@
...
+		"unix_nano",
+	)
+}
+
+// CompareSameType implements comparison of two Time values. required by
+// starlark.Comparable interface.
+func (t Time) CompareSameType(op syntax.Token, yV starlark.Value, depth int) (bool, error) {
+	x := time.Time(t)
+	y := time.Time(yV.(Time))
+	cp := 0
```
alandonovan, line 396: cmp :=

starlarktime/time.go
```diff
@@ -0,0 +1,507 @@
...
+		case Duration:
+			return Time(x.Add(time.Duration(y))), nil
+		}
+	case syntax.MINUS:
+		switch y := yV.(type) {
+		case Duration:
+			return Time(x.Add(time.Duration(-y))), nil
+		case Time:
+			// time - time = duration
+			if side == starlark.Left {
```
alandonovan, line 425: This condition is (I think) always true, because the left operand is always tested for HasBinary first, so the "else" case is unreachable. Simplify.

starlarktime/time.go
```diff
@@ -0,0 +1,507 @@
...
+func builtinAttrNames(methods map[string]builtinMethod) []string {
+	names := make([]string, 0, len(methods))
+	for name := range methods {
+		names = append(names, name)
+	}
+	sort.Strings(names)
+	return names
+}
+
+// threeway interprets a three-way comparison value cmp (-1, 0, +1)
```
alandonovan, line 489: FYI: in hindsight, the API design that required `threeway` was a mistake. Originally it was intended to support floating point NaNs according to IEEE 754, which requires that x < NaN, Nan < x and NaN == x are all false, so you can't represent the result of a "float op float" comparison using a three-valued result (smaller, bigger, equal): one must also know the query operator. But we changed things recently so that floats are totally ordered (NaN > +Inf), otherwise you can't sort floats, so CompareSameType could in principle be simplified to:

```
CompareSameType(v Value, depth int) (signum int, err error)
```

starlarktime/time.go
```diff
@@ -0,0 +1,507 @@
+package starlarktime
```
alandonovan, line 1: What happened to the doc comment?

lib/time/time.go
alandonovan: Starlark's error-handling theory is not well developed. It doesn't have exceptions, and I don't think we really want to add them, but this does mean that callers need to "look before they leap", or need a mode in which they succeed but return an error value (e.g. None or perhaps a string) instead of failing.

lib/time/time.go
alandonovan: It's fine to insist that the value be non-nil, and simply panic if someone violates that. Users: don't do that.

starlark/testdata/time.star
alandonovan: (I realize I said the opposite for the other functions, but they all have 1 parameter with a nonce name like "x".)

lib/time/time.go
srebhan: @alandonovan right, but `parseLocation` suggests that we somehow convert the given location to something else. Instead we simply return the same string in case of success or `nil` otherwise. Furthermore, the returned error is totally redundant with checking the string.... My suggestion is to simplify this to return a bool indicating if this is a valid location. So instead of 
```golang
l, err := parseLocation(..., loc)
if err != nil {...}
```
you can call
```golang
if isValidLocation(..., loc) {...}
```

starlarktime/time.go
essobedo: @alandonovan it has been removed since it was not in a standard format and there is no other module with this kind of documentation

essobedo:
@alandonovan Remarks applied, please check again

ssoroka:
Any update on this? :)

essobedo:
@b5 I cannot resolve conversations, could you please check if it is possible to give me the necessary rights?

lib/time/time.go
essobedo: FYI `NowFunc` cannot return `nil` because it is actually `time.Now`

essobedo:
@adonovan @alandonovan We really need this feature for https://github.com/influxdata/telegraf, so it is important for us to finish it and have it in Startlark. Can you still do the review or shall I ask someone else? and if so, who is the new project lead now please? (cc @ssoroka @srebhan)

lib/time/time.go
b5: It can be nil if users override the package variable, setting `time.NowFunc == nil`. In discussion above myself & Alan have agreed it's the package consumer's responsibility to not do that

essobedo: @b5 thanks for the info, if I get it right, it is acceptable to let if fail then right (leave it as it is now)?

b5: yep! I think it should be left as is

lib/time/time.go
b5: Agreed. I think we should add this change

lib/time/time.go
b5: ditto here.

lib/time/time.go
adonovan: Yes, we should do that. (Or perhaps define a new, simpler CompareSameType2 interface and make the interpreter test it first.) I've filed https://github.com/google/starlark-go/issues/356.

lib/time/time.go
```diff
@@ -0,0 +1,493 @@
...
+		"nanoseconds",
+	}
+}
+
+// CompareSameType implements comparison of two Duration values. required by
+// starlark.Comparable interface.
+func (d Duration) CompareSameType(op syntax.Token, v starlark.Value, depth int) (bool, error) {
+	x := time.Duration(d)
+	y := time.Duration(v.(Duration))
+	switch op {
```
adonovan, line 180: return threeway(op, x, y), nil

adonovan (commented):
Sorry for the delay. I'll do what I can to help you get this committed today.

lib/time/time.go
```diff
@@ -0,0 +1,493 @@
...
+				return nil, fmt.Errorf("unsupported operation")
+			}
+			i, ok := y.Int64()
+			if !ok {
+				return nil, fmt.Errorf("int value out of range (want signed 64-bit value)")
+			}
+			if int64(i) == 0 {
+				return nil, fmt.Errorf("%s division by zero", d.Type())
+			}
+			return Duration(x.Nanoseconds() / i), nil
```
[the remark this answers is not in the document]
essobedo: @adonovan you meant `d / Duration(i)`, right?

lib/time/time.go
essobedo: @adonovan I'm not sure, I get what you expect here. You would to see the content of the switch block moved into a new private function?

adonovan: Yes, and that function already exists below: it's `threeway`.

lib/time/time.go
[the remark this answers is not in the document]
adonovan: Yes, quite right.

starlark/testdata/time.star
```diff
@@ -0,0 +1,119 @@
...
+# duration // duration = int
+assert.eq(d2 // time.parse_duration("16m"), 37)
+assert.fails(lambda: d2 // time.parse_duration("0"), "division by zero")
+# duration * int = duration
+assert.eq(d * 1000, time.parse_duration("16m40s"))
+assert.fails(lambda: d2 // 0, "division by zero")
+
+
+before = time.now()
+time.sleep(10 * time.millisecond)
```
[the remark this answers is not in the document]
b5: Ah sorry, I missed the point where `time.sleep` was removed. We need a time.sleep for at least one use-case that's admittedly outside of the "configuration or query language" context: to add delay between HTTP requests. Removing `time.sleep` means that sadly we won't be able to make use of this package.

I understand that our use case is outside of the described goals of the Starlark project, but given that this PR is adding a time package, the notion of determinism is a bit... fluid, no?

I'd love it if we could have an API for time sleep that can be overridden, very similar to how we're handling `time.now`, but if that's going to derail getting this merged today, I'll move this conversation to a follow-up

lib/time/time.go
essobedo: @adonovan I probably miss something obvious but the current function has only 2 parameters while you are trying to call a function with 3 parameters. At best the second parameter could be somehow `x-y` but we could get an overflow.

starlark/testdata/time.star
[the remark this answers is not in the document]
adonovan: > I understand that our use case is outside of the described goals of the Starlark project, but given that this PR is adding a time package, the notion of determinism is a bit... fluid, no?

Well, all the other functions in the package are useful for computation on values obtained from (say) a database, where you want determinism, but not the ability to delay a request, which could easily lead to denial of service. Sleeping is an operation a thread, not a computation. Perhaps making it globally overridable, like time.Now, is a reasonable compromise.

lib/time/time.go
adonovan: Oops, you're right; I'm not really using my brain. What I mean was this, which is less concise that I had imagined:
```
	cmp := 0
	if x, y := d, v.(Duration); x < y {
		cmp = -1
	} else if x > y {
		cmp = 1
	}
	return threeway(op, cmp)
```

starlark/testdata/time.star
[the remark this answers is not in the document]
MonicaRJones: Our application cannot have a sleep function.   If the function is included, then we will `delete(time.Module.Members, "sleep")` in an init function.  I think it's better to allow applications to add the sleep function with  `time.Module.Members["sleep"] = starlark.NewBuiltin( ... )`.   That way, the application has full control over what the function actually does.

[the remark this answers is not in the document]
b5: ah ok, I see other's points and think it makes more sense to leave it out. Making it overridable from the go side as with `time.now` works, but given the intended use for starlark & the fact that others would need to write documentation / actively work against the existence of a `time.sleep` method, let's leave it out & we can add it back higher up the stack. 

the other upside is one less thing in the way of getting this mega-PR merged 😄

Thanks for chiming in @MonicaRJones

lib/time/time.go
essobedo: For consistency, I believe that it should rather remain `time.duration`, otherwise we would need to apply the same change for other modules too. @adonovan WDYT?

essobedo:
@adonovan Thx for the feedbacks, it is hard to follow it properly since I have no way to resolve the discussions but I think that I addressed your feedbacks, could you please check it again?

starlark/testdata/time.star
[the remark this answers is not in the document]
adonovan: Good idea: safer to leave it out for now and perhaps add it later, or let applications add it themselves (possibly using the slightly fishy but very convenient practice of modifying the module directly).

lib/time/time.go
```diff
@@ -0,0 +1,479 @@
...
+var NowFunc = time.Now
+
+// Parses the given duration string.
+// A duration string is a possibly signed sequence of
+// decimal numbers, each with optional fraction and a unit suffix,
+// such as "300ms", "-1.5h" or "2h45m".
+// Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
+func parseDuration(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var d Duration
+	err := starlark.UnpackPositionalArgs("duration", args, kwargs, 1, &d)
```
adonovan, line 45: "parse_duration"

adonovan (approved):
LGTM. Thanks for your patience!

lib/time/time.go
essobedo: @alandonovan indeed, fixed

b5:
quick think you to everyone who got this off the ground! @srebhan for motivating this in the first place, and @essobedo for putting in a huge amount of work to implement @adonovan's careful review 😄

srebhan:
@b5 I also want to thank you for your continuous effort in bringing this forward. I know it's not easy to find the time. Special thanks to @essobedo for investing so much and pushing this over the finishing line!

essobedo:
@adonovan Thanks for the impressive review, can you still somehow merge ? or do you at least know who can merge?

PR #328, opened 2020-12-04
resolver: make -nesteddef and -lambda always on

alandonovan:
See https://github.com/bazelbuild/starlark/pull/145
for spec changes.

Updates https://github.com/bazelbuild/starlark/issues/20


jayconrod (approved):
Looks good. Does this mean Bazel will support these as well?

alandonovan:
> Looks good. Does this mean Bazel will support these as well?

Yep, the plan is that the spec and all implementations support it.

PR #329, opened 2020-12-10
starlark: add ToInt helper function for unpacking values to Go ints

alandonovan:
Also, use it in Unpack* functions.
This is a minor breaking change: any call that assumed values
unpacked to 'int' were always in the int32 range will need to be modified
to handle possible int64 values.


jayconrod (approved):

alandonovan:
Thanks, Jay.

PR #330, opened 2020-12-10
starlark: add 'bytes' data type, for binary strings

alandonovan:
This change defines a 'bytes' data type, an immutable string of
bytes. In this Go implementation of Starlark, ordinary strings
are also strings of bytes, so the behavior of the two is very similar.
However, that is not required by the spec. Other implementations of
Starlark, notably in Java, may use strings of UTF-16 codes for the
ordinary string type, and thus need a distinct type for byte strings.


lib/proto/proto.go
```diff
@@ -1232,78 +1230,3 @@ func (x EnumValueDescriptor) CompareSameType(op syntax.Token, y_ starlark.Value,
...
 	}
 }
-
-// A Bytes is an immutable sequence of bytes.
-// It is comparable, iterable, indexable, and sliceable.
-//
-// (In go.starlark.net, text Strings are also byte strings,
-// but we shouldn't rely on that.
-// See https://github.com/bazelbuild/starlark/issues/112.)
-type Bytes string
```
jayconrod, line 1242: Could this be a type alias to the new type? Not sure if this is used yet.

jayconrod (approved):

lib/proto/proto.go
alandonovan: > Could this be a type alias to the new type?

No. syntax.Bytes is a part of the syntax package, whereas starlark.Bytes (a value) is part of the interpreter. Even if we wanted them to be the same, the dependency is in the wrong direction.

> Not sure if this is used yet.

It's the type of the "payload" of syntax.Literal when TokenKind==BYTES. Having just said that, I wonder if we can use string since the TokenKind should give us the information need to discriminate. [checks...] Ah, the reason is that compile.Program.Constants[i] needs to discriminate based on type alone. I suppose it could define a private Bytes type. Done.

starlark/testdata/bytes.star
```diff
@@ -0,0 +1,112 @@
...
+assert.eq(list(hello.elems()),
+          [b"h", b"e", b"l", b"l", b"o", b",", b" ",
+           b"\xE4", b"\xB8", b"\x96", b"\xE7", b"\x95", b"\x8C"])
+assert.eq(list(goodbye.elems()),
+          [b"g", b"o", b"o", b"d", b"b", b"y", b"e"])
+assert.eq(list(empty.elems()), [])
+
+# TODO(adonovan);
+# - more methods: find, index, split, etc.
+# - constructor from int values: bytes([1, 2, 3])?
```
alandonovan, line 98: Jay, do you have any thoughts on what should be the syntax to construct a bytes from a list of numeric byte values, or the various ord/chr operations to convert numbers to/from bytes, UTF-k codes, and Unicode code points?

jayconrod: I think Starlark should match Python 3 wherever possible. I don't know Python 3 that well though.

For the `bytes` constructor, I think that means accepting a string, `bytes`, or sequence of integers in the range [0, 256). We probably shouldn't support the `encoding` or `errors` parameters for now.

`ord` seems correct: we should only accept `bytes` of length 1.

`chr` I think should convert a code point to however many UTF-k codes are needed? `bytes(chr(i))` would convert to UTF-8 I think? I'm probably missing some cases though.

Would `str(bytes(chr(i))` be equivalent to `chr(i)`?

alandonovan: Ah, I didn't know that bytes([int]) worked; good. I've implemented that. I also made the string.elems iterable view indexable, which addresses some of the gaps.

The only missing case now is how to create a 1-element string, or a string from a list of int values, analogous to bytes([int]). Any ideas?

Take a look at this updated comment for a summary of the various operations.

> Would str(bytes(chr(i)) be equivalent to chr(i)?

Correct.

lib/proto/proto.go
jayconrod: Ah, I meant to ask if this could be a type alias for `starlark.Bytes`, just to avoid a breaking change to users of `lib/proto`. Not sure if anyone is using this type though, since `lib/proto` landed recently.

starlark/testdata/bytes.star
```diff
@@ -0,0 +1,131 @@
...
+# - string.elems()[i] returns substring of a single element (UTF-k code).
+# - int(string) - parse string as decimal (or other) numeric literal.
+#
+# number to string:
+# - chr(int) returns string, UTF-k encoding of Unicode code point (like Python).
+#   Redundant with '%c' % int (which Python2 calls 'unichr'.)
+# - TODO(adonovan): how to create a 1-element string? (analogous to bytes([int]))
+# - bytes(chr(int)) returns byte string containing UTF-8 encoding of one code point.
+# - bytes([int]) returns 1-byte string (with regrettable list allocation).
+# - string(int) - format number as decimal.
```
jayconrod, line 131: Should be `str(int)` instead?

starlark/testdata/bytes.star
```diff
@@ -0,0 +1,131 @@
...
+assert.eq(type(hello.elems()), "bytes.elems")
+assert.eq(str(hello.elems()), "b\"hello, 世界\".elems()")
+assert.eq(list(hello.elems()),
+          [b"h", b"e", b"l", b"l", b"o", b",", b" ",
+           b"\xE4", b"\xB8", b"\x96", b"\xE7", b"\x95", b"\x8C"])
+assert.eq(list(goodbye.elems()),
+          [b"g", b"o", b"o", b"d", b"b", b"y", b"e"])
+assert.eq(list(empty.elems()), [])
+
+# TODO(adonovan);
```
jayconrod, line 106: (New thread: GitHub won't let me comment on an old thread after a force-push)

> The only missing case now is how to create a 1-element string, or a string from a list of int values, analogous to bytes([int]). Any ideas?

> Take a look at this updated comment for a summary of the various operations.

For a 1-element string from an integer code point, `chr` should work. That can produce multiple UTF-k elements, depending on the implementation.

For a string from a list of int code points, I don't know of a way to do that in Python. `str([1, 2, 3])` should return literally "[1, 2, 3]"`. It looks like Python `bytes` has a `decode` method. Maybe we should implement that, only supporting `"utf-8"` (or no argument)?

starlark/testdata/bytes.star
```diff
@@ -0,0 +1,131 @@
...
+# - more methods: find, index, split, etc.
+# - hash(bytes)?
+#
+# Summary of string operations (put this in spec).
+#
+# string to number:
+# - ord(bytes)    returns numeric byte value of bytes[0] (requires len=1).
+# - ord(bytes[i]) returns numeric byte value of bytes[i].
```
jayconrod, line 114: `ord(bytes[i])` is `ord(int)`; I don't think that should be valid.

alandonovan: Ah, indexing a Python3 byte string returns a numeric byte value, not a bytes of length 1 (like text strings). I've changed Starlark/Go to match.

It occurs to me that ord(string) is not very useful on its own, because it requires the string to encode exactly one code point, yet the language provides no way to ensure that (other than iterating over string.codepoints()). Perhaps ord(string) should return the _first_ code point? Or it could have an optional index=int parameter for the start offset, which avoids the need to allocate a substring by using ord(string[i:]).

starlark/testdata/bytes.star
alandonovan: > For a 1-element string from an integer code point, chr should work. That can produce multiple UTF-k elements, depending on the implementation.

Right, chr(int) produces a 1-code point string of multiple UTF-k codes (elements). But how would one create a string of exactly one element?

> For a string from a list of int code points

That's easy enough, if inefficient: "".join(["%c" % cp for cp in codepoints]).

starlark/testdata/bytes.star
alandonovan: Done.

starlark/testdata/bytes.star
```diff
@@ -0,0 +1,131 @@
+# Tests of 'bytes' (immutable byte strings).
+
+load("assert.star", "assert")
+
+# bytes(string) -- UTF-k to UTF-8 transcoding with U+FFFD replacement
+hello = bytes("hello, 世界")
+goodbye = bytes("goodbye")
+empty = bytes("")
+nonprinting = bytes("\t\n\u200D")  # TAB, NEWLINE, DEL, ZERO_WIDTH_JOINER
```
brandjon, line 9: Missing DEL?

starlark/testdata/bytes.star
```diff
@@ -0,0 +1,131 @@
...
+
+# bytes(iterable of int) -- construct from numeric byte values
+assert.eq(bytes([65, 66, 67]), b"ABC")
+assert.eq(bytes((65, 66, 67)), b"ABC")
+assert.eq(bytes([0xf0, 0x9f, 0x98, 0xbf]), b"😿")
+assert.fails(lambda: bytes([300]),
+             "at index 0, 300 out of range .want value in unsigned 8-bit range")
+assert.fails(lambda: bytes([b"a"]),
+             "at index 0, got bytes, want int")
+assert.fails(lambda: bytes(1), "want string, bytes, or iterable of ints")
```
brandjon, line 20: You could imagine `bytes(ord=n)` to mean `bytes([n])` if you really, really want to avoid the list allocation. But I think that might be premature. (And how much worse is a list allocation than the function call's args arrays, anyway?)

starlark/testdata/bytes.star
```diff
@@ -0,0 +1,131 @@
+# Tests of 'bytes' (immutable byte strings).
+
+load("assert.star", "assert")
+
+# bytes(string) -- UTF-k to UTF-8 transcoding with U+FFFD replacement
```
brandjon, line 5: Add a case with an unpaired surrogate. And a case with a proper surrogate pair. If `bytes(string)` doesn't support unpaired surrogates, then construct the bytes object with a literal and inspect that it has the expected contents with indexing.

starlark/testdata/bytes.star
```diff
@@ -0,0 +1,131 @@
...
+assert.eq(str(empty), "")
+assert.eq(str(nonprinting), "\t\n\u200d")
+assert.eq(str(b"\udc00"), "�") # unpaired surrogate => U+FFFD
+
+# reprq
+assert.eq(repr(hello), r'b"hello, 世界"')
+assert.eq(repr(hello[:-1]), r'b"hello, 世\xe7\x95"')  # (incomplete UTF-8 encoding )
+assert.eq(repr(goodbye), 'b"goodbye"')
+assert.eq(repr(empty), 'b""')
+assert.eq(repr(nonprinting), 'b"\\t\\n\\u200d"')
```
brandjon, line 57: What about `repr(b"\udc00")`?

starlark/testdata/bytes.star
```diff
@@ -0,0 +1,131 @@
...
+
+# str(bytes) does UTF-8 to UTF-k transcoding.
+assert.eq(str(hello), "hello, 世界")
+assert.eq(str(hello[:-1]), "hello, 世��")  # incomplete UTF-8 encoding => U+FFFD
+assert.eq(str(goodbye), "goodbye")
+assert.eq(str(empty), "")
+assert.eq(str(nonprinting), "\t\n\u200d")
+assert.eq(str(b"\udc00"), "�") # unpaired surrogate => U+FFFD
+
+# reprq
```
brandjon, line 52: Nit: typo 'q'

starlark/testdata/bytes.star
```diff
@@ -0,0 +1,131 @@
...
+assert.fails(lambda: ord(b"ab"), "ord: bytes has length 2, want 1")
+assert.fails(lambda: ord(b""), "ord: bytes has length 0, want 1")
+
+# repeat (bytes * int)
+assert.eq(goodbye * 3, bytes("goodbyegoodbyegoodbye"))
+assert.eq(3 * goodbye, bytes("goodbyegoodbyegoodbye"))
+
+# elems() returns an iterable value over 1-byte substrings.
+assert.eq(type(hello.elems()), "bytes.elems")
+assert.eq(str(hello.elems()), "b\"hello, 世界\".elems()")
```
brandjon, line 98: The iterable elems view has a `str()` that identifies the original string and the method call? That seems surprisingly detailed. I'd have settled for "<elems object>" or similar.

starlark/testdata/bytes.star
```diff
@@ -0,0 +1,131 @@
...
+assert.eq(list(hello.elems()), [104, 101, 108, 108, 111, 44, 32, 228, 184, 150, 231, 149, 140])
+assert.eq(bytes([104, 101, 108, 108, 111, 44, 32, 228, 184, 150, 231, 149, 140]), hello)
+assert.eq(list(goodbye.elems()), [103, 111, 111, 100, 98, 121, 101])
+assert.eq(list(empty.elems()), [])
+
```
brandjon, line 103: Does the `bytes` constructor take an arbitrary iterable of octets? If so, test feeding it an elems object (bypassing the list intermediary).

starlark/testdata/bytes.star
brandjon: See my comment below. For implementations with k=8, I think you need \xXX literals up to 0xFF. For k=16, I think it suffices to have \uXXXX and allow unpaired surrogates.

starlark/testdata/bytes.star
```diff
@@ -0,0 +1,131 @@
...
+# TODO(adonovan);
+# - more methods: find, index, split, etc.
+# - hash(bytes)?
+#
+# Summary of string operations (put this in spec).
+#
+# string to number:
+# - bytes[i]  returns numeric value of ith byte.
+# - ord(string)  returns numeric value of sole code point in string.
+# - ord(string[i])  is not a useful operation: fails on non-ASCII; see below.
```
brandjon, line 113: This seems unfortunate. I feel like I'm getting mixed signals about whether strings are unicode data or sequences of k-bit elements. If it's the latter, you'd think we'd be freer to manipulate and inspect their contents, including getting the integer value of an element. It's enough to make me think `ord()` should have a companion `codepoint_ord()` toplevel function.

brandjon: See my [comment](https://github.com/bazelbuild/starlark/issues/112#issuecomment-750357023) on the design thread for more on this issue and the one I mentioned below.

starlark/testdata/bytes.star
```diff
@@ -0,0 +1,131 @@
+# Tests of 'bytes' (immutable byte strings).
```
brandjon, line 1: Do we have a test somewhere that you can't assign to `b[i]`? Same for strings.

alandonovan:
Hi Jon, Jay,

I have updated this change to the Go impl, and the spec change in https://github.com/bazelbuild/starlark/pull/161 to clean up the following primary parts of the problem: 
- bytes values
- bytes literal syntax
- \x \X \u \U escapes
- portability of strings

The remaining parts of the problem are:
- conversions (string, bytes, numbers), ord, chr, and strictness of conversion failures
- additional methods, including iterators
- hash(bytes)

To make this process more manageable and reduce the number of rounds of review on an increasingly large pair of PRs, I propose that we aim to commit the spec change and the Go implementation once we are happy with the "phase 1" features, and then make follow-up changes to address the "phase 2" items. Given that no-one is yet using bytes, I don't think this will be disruptive.

cheers
alan

jayconrod (commented):
Quick first look: this phased approach sounds good to me. I still need to look at the parser change and tests in more detail, but looks good so far.

internal/compile/compile.go
```diff
@@ -309,12 +310,15 @@ func (op Opcode) String() string {
 type Program struct {
 	Loads     []Binding     // name (really, string) and position of each load stmt
 	Names     []string      // names of attributes and predeclared variables
-	Constants []interface{} // = string | int64 | float64 | *big.Int
+	Constants []interface{} // = string | int64 | float64 | *big.Int | Bytes
```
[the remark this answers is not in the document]
alandonovan: Yes, well spotted.

brandjon (approved):

alandonovan:
PTAL; thanks.

starlark/testdata/bytes.star
alandonovan: Done.

starlark/testdata/bytes.star
alandonovan: That's a good idea, though I agree let's not add it just yet. Avoiding the list does save two extra allocations, and this is the kind of thing that might well appear in a loop over bytes.

starlark/testdata/bytes.star
alandonovan: \u disallows surrogates; see scan_test.

starlark/testdata/bytes.star
alandonovan: Fixed.

starlark/testdata/bytes.star
alandonovan: I don't expect these will be stringified except when debugging, and it's very easy to specify and implement this behavior, so why not?

starlark/testdata/bytes.star
alandonovan: Good idea. Done.

starlark/testdata/bytes.star
alandonovan: Added.

starlark/testdata/string.star
```diff
@@ -6,6 +6,14 @@ load("assert.star", "assert")
 # raw string literals:
 assert.eq(r"a\bc", "a\\bc")
 
+# Hex and octal escapes may encode any byte value
+# even in a "text" (not 'bytes') string.
+# This is not required by the spec, but is necessary
+# in the Go implementation so that repr(x) works for
+# invalid Unicode, such as half an emoji.
```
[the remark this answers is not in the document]
alandonovan: \X escapes have been removed from the spec change, since there is no possible portable interpretation of them.

starlark/testdata/bytes.star
alandonovan: \x escapes don't allow non-ASCII, and \u escapes don't allow unpaired surrogates, because that would be non-portable, as would the now-abandoned \X escape notation. Now, there is simply no literal syntax to denote a string with an invalid (non-text) encoding.  (In the Rust implementation, such values cannot even be represented.)

starlark/testdata/bytes.star
alandonovan: > The only missing case now is how to create a 1-element string, or a string from a list of int values, analogous to bytes([int]). Any ideas?

I suppose bytes([int]) is good enough for now, and we can optimize it later.

jayconrod (approved):
All looks good. Sorry for the delay; busy week.

Issue #332, opened 2021-01-02
Immutability of Int is broken by BigInt() method


gfcapella:
Hi,
First of all, great work! It is a pleasure working the library.
I realized that Int's interface is immutable, except for the following method (big.Int is mutable):

> // BigInt returns the value as a big.Int.
> // The returned variable must not be modified by the client.
> func (i Int) BigInt() *big.Int {
> ...
> }

Is that for a reason? I see the method is used internally in several places. I would imagine it is better to split it in 2:
- BigInt() -> public method, returns a copy of the BigInt to maintain the immutability properties
- bigInt() -> internal, returning directly the reference to the big.Int. Not sure if the naming I propose would be best, other ideas could be ptrInt(), internalBigInt(), bint(), ...

If this is something that would be approved, I can work on the change and send a pull request.

Best,
Germán

alandonovan:
Fair enough, let's make BigInt return a copy, and bigInt avoid the copy if possible. Thanks.

PR #333, opened 2021-01-04
Fix: Immutability of Int is broken by BigInt method (#332)

gfcapella:
The interface of Int assumes the value is immutable but a reference to
its content is leaked through the BigInt() method, which is mutable.
This fix resolves the leak by:
- making a copy of the *big.Int in MakeBigInt()
- returning a copy of the *big.Int content in BigInt()
- creating an internal method bigInt() for internal use that still
returns reference to the internal big.Int

Fixes: #332

### DIFF
--- a/starlark/int.go
+++ b/starlark/int.go
@@ -44,7 +44,7 @@ func MakeUint64(x uint64) Int {
 }
 
 // MakeBigInt returns a Starlark int for the specified big.Int.
-// The new Int value will contain a copy of x, the caller is safe to modify x.
+// The new Int value will contain a copy of x. The caller is safe to modify x.
 func MakeBigInt(x *big.Int) Int {
 	if n := x.BitLen(); n < 32 || n == 32 && x.Int64() == math.MinInt32 {
 		return makeSmallInt(x.Int64())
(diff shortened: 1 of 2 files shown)


starlark/int.go
```diff
@@ -44,12 +44,13 @@ func MakeUint64(x uint64) Int {
 }
 
 // MakeBigInt returns a Starlark int for the specified big.Int.
-// The caller must not subsequently modify x.
+// The new Int value will contain a copy of x, the caller is safe to modify x.
```
alandonovan, line 47: Replace comma by semicolon or period.

starlark/int.go
```diff
@@ -86,9 +87,20 @@ func (i Int) Int64() (_ int64, ok bool) {
 	return iSmall, true
 }
 
-// BigInt returns the value as a big.Int.
-// The returned variable must not be modified by the client.
+// BigInt returns a copy of the value as a big.Int.
```
alandonovan, line 90: // BigInt returns a new big.Int with the same value as the Int.

(and strike the second line

starlark/int.go
```diff
@@ -86,9 +87,20 @@ func (i Int) Int64() (_ int64, ok bool) {
...
+	iSmall, iBig := i.get()
+	if iBig != nil {
+		return new(big.Int).Set(iBig)
+	}
+	return big.NewInt(iSmall)
+}
+
+// bigInt returns the value as a big.Int.
+// It differs from BigInt in that this method returns the actual
+// reference and any modification will change the state of i
```
alandonovan, line 102: Missing period.

alandonovan (changes requested):
Sorry for the delay. Logic looks good. Only very minor doc comment tweaks. Thanks for the PR!

gfcapella:
No problem! comments have now the fixes requested

alandonovan (approved):

PR #338, opened 2021-01-18
Add functionality for parsing floats

jammiess:
The UnpackArgs command does not support parsing floats out of arguments. This pull request adds that functionality to the unpackOneArg function which then adds that functionality to the UnpackArgs command.

### DIFF
--- a/starlark/unpack.go
+++ b/starlark/unpack.go
@@ -202,6 +202,12 @@ func unpackOneArg(v Value, ptr interface{}) error {
 	case *int, *int8, *int16, *int32, *int64,
 		*uint, *uint8, *uint16, *uint32, *uint64, *uintptr:
 		return AsInt(v, ptr)
+    case *float64:
+        f, ok := v.(Float)
+        if !ok {
+            return fmt.Errorf("got %s, want float", v.Type())
+        }
+        *ptr = float64(f)
 	case **List:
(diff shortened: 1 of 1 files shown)


jammiess:
@googlebot I signed it!

johnyburd (approved):

starlark/unpack.go
```diff
@@ -202,6 +202,12 @@ func unpackOneArg(v Value, ptr interface{}) error {
 	case *int, *int8, *int16, *int32, *int64,
 		*uint, *uint8, *uint16, *uint32, *uint64, *uintptr:
 		return AsInt(v, ptr)
+    case *float64:
+        f, ok := v.(Float)
```
alandonovan, line 206: The indentation is wrong here. Please run gofmt on your code before sharing it with others: this is an inviolate expectation of essentially every project that uses Go. (All editors can be configured to run gofmt before saving a buffer containing Go code. I highly recommend it; the effect is liberating.)

alandonovan (commented):
Code change looks good. Thanks for adding this.

alandonovan (approved):
Thanks!

starlark/unpack.go
jammiess: This is my first time really getting into go. Thanks for the tip!

PR #345, opened 2021-01-26
Remove an unused Thread in the sequential load example

Wyverald:
A Thread is created before the outermost `load` and is then never used. (In `load`, a thread is created for each module being loaded.)

### DIFF
--- a/starlark/example_test.go
+++ b/starlark/example_test.go
@@ -116,8 +116,7 @@ func ExampleThread_Load_sequential() {
 		return e.globals, e.err
 	}
 
-	thread := &starlark.Thread{Name: "exec c.star", Load: load}
-	globals, err := load(thread, "c.star")
+	globals, err := load(nil, "c.star")
 	if err != nil {
 		log.Fatal(err)
 	}


starlark/example_test.go
```diff
@@ -116,8 +116,7 @@ func ExampleThread_Load_sequential() {
 		return e.globals, e.err
 	}
 
-	thread := &starlark.Thread{Name: "exec c.star", Load: load}
-	globals, err := load(thread, "c.star")
+	globals, err := load(nil, "c.star")
```
alandonovan, line 119: Your change looks sound; the thread is indeed unnecessary, though if this implementation of load were in future to require information from the thread (say, the mapping from module names to openable file locations---as for Bazel BUILD files, for example), then then thread parameter would no longer be unnecessary. Another feature of Bazel BUILD files is they are not loadable modules; only .bzl files can be explicitly loaded. So, if you need to access the thread during load, or if you need the 'main' file not to be directly loadable, then the appropriate structure is to replace the load(nil) call with this:

```
// Execute main, allowing it to load other .star files.
thread := &starlark.Thread{Name: "exec main", Load: load}
globals, err := starlark.ExecFile(thread, "main.star", "load('c.star', _c='c'); c = _c", nil)
```

alandonovan (approved):

Issue #346, opened 2021-01-31
 Specify starting line and column for expression parser ParseExpr


MonicaRJones:
One of the file formats in the application that I am developing uses Starlark for writing expressions.  The Starlark expressions are surrounded by source code written in another language.

I want to specify a starting line number and column number when parsing expressions so that errors point back to the original source code. 

Here's a suggested function for the syntax package.

    // ParseExpr parses a Starlark expression in src. A comma-separated
    // list of expressions is parsed as a tuple. Positions in the resulting 
    // expression are based on pos.
    func ParseExprPosition(pos Position, src []byte, mode Mode) (expr Expr, err error)

The proposed function does not have the usual filename, src pattern as in other functions in the syntax package because it's designed to work on a snippet from a larger file. 

My feature request is for a way to specify the starting line and column number for the scanner, not for the specific function shown above.

My fallback plan is to rewrite positions in syntax.Error and starlark.EvalError using a source map that I maintain on the side.  I can do that, but it seems that other users of Starlark expressions will have a similar requirement.

alandonovan:
It's not necessary to change the API; you can do this by decorating the expression text with whitespace and parens. If you want an expression to start at line L, column C, then feed the parser `(L-1)*"\n" + "("*C + expr + ")"*C`.

MonicaRJones:
That works. Thank you.

It seems that it will be more common to pass a file snippet to syntax.ParseExpr than an entire file. It would be nice if the package supported the common scenario more directly than the suggested workaround. The workaround for the package limitation was not obvious to me.  It may not be obvious to others.

alandonovan:
I don't want to expand the API, but I agree this trick is worth recording in a comment at ParseExpr.

MonicaRJones:
The workaround  does not work when there are mismatched parentheses in the input.

	_, err := syntax.Parse("x", "(", 0)
	fmt.Println(err)  // prints x:1:2: got end of file, want primary expression

        // Wrap "(" with ((( ))) for input starting at column 4.
	_, err = syntax.Parse("x", "(((()))", 0)
	fmt.Println(err)  // prints x:1:8: got end of file, want ')'

I want position x:1:4 for the second snippet, but got x:1:8.   Column 8 is outside the range of the original input,.

I am investigating two options:  
- rewrite positions in errors using a source map.
- copy the relevant parts of the syntax package to my own package so I can make the fix.

alandonovan:
[...]

PR #349, opened 2021-02-08
syntax: support setting initial line/col for scanner

alandonovan:
The new FilePortion type, which may be provided to the scanner,
parser, or ExecFile functions, combines a piece of text
with its start line/column numbers, for applications that
extract a Starlark expression from the middle of a larger file.

Fixes https://github.com/google/starlark-go/issues/346

### DIFF
--- a/syntax/parse.go
+++ b/syntax/parse.go
@@ -28,7 +28,7 @@ const (
 // If src != nil, ParseFile parses the source from src and the filename
 // is only used when recording position information.
 // The type of the argument for the src parameter must be string,
-// []byte, or io.Reader.
+// []byte, io.Reader, or FilePortion.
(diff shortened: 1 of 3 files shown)


jayconrod (approved):
Looks good.

Another approach would be to support `#line filename:line:column` comments in the scanner. That would be more general, but more complicated of course.

alandonovan:
> Another approach would be to support #line filename:line:column comments in the scanner. That would be more general, but more complicated of course.

That's a good point, though my experience of line directives in several languages is that they create a great deal of complexity: some clients want the raw line numbers, others want the cooked. They also break the invariant that all the statements within a given file have the same filename, but I suppose that's fixable by making them line-and-column directives instead of file-and-line directives.

Issue #347, opened 2021-02-04
UnpackArgs should accept None for optional args


nicks:
## Expected behavior:

Consider the following Go function implementation:

```
var optionalArg MyCustomValue
if err := UnpackArgs("my_func", args, kwargs, "optional_arg?", &optionalArg); err != nil {
  return nil, err
}
```

I would expect this starlark code to work:

```
my_func(optional_arg=None)
```

with the reasoning that "None" should be semantically equivalent to omitting the argument. 

## Actual behavior

You get an error:

```
for parameter "optional_arg": got NoneType, want MyCustomValueType
```

## Other notes

This is purely an argument about the ergonomics of UnpackArgs, not about the semantics of Starlark itself.

Also! I spotted at least one bug in starlark caused by this API gotcha. Note that in Python,

```
$ python3
Python 3.7.2 (default, Jul  6 2020, 12:18:55) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> sorted([], key=None)
[]
```

is fine, but in Starlark,

```
$ starlark
Welcome to Starlark (go.starlark.net)
>>> sorted([], key=None)
Traceback (most recent call last):
  <stdin>:1:7: in <expr>
Error in sorted: sorted: for parameter "key": got NoneType, want callable
```

alandonovan:
> with the reasoning that "None" should be semantically equivalent to omitting the argument.

I don't think we can assume this to be desirable across all Starlark (or Python) APIs. For example, `getattr(x, "f")` behaves differently from `getattr(x, "f", None)`: the former fails if the field x.f does not exist, whereas the latter returns None. For that reason I don't think we should hardwire this policy choice into the core APIs.

One can always defined an Unpacker that implements "None means absent" semantics. Perhaps we could provide a utility function to make that easier?

nicks:
re: `getattr` - ya, that's a good counter-example. I was trying to come up with a good rule thumb for when None is passed to optional arguments.

Maybe the rule should be: passing None to an optional argument should never be an error, and that UnpackArgs should leave the pointer untouched rather than returning an error. But it's OK if it's semantically meaningful.

A utility function for the "None means absent" semantics seems reasonable. I worry that it makes the API harder to use if you have to know about this function to handle an obscure edge case. Were you thinking something like:

```
var optionalArg MyCustomValue
[...]

PR #350, opened 2021-02-09
starlark: Add "??" syntax for none-coalescing parameters to UnpackArgs.

nicks:
Fixes https://github.com/google/starlark-go/issues/347

### DIFF
--- a/starlark/unpack.go
+++ b/starlark/unpack.go
@@ -24,8 +24,11 @@ type Unpacker interface {
 // Iterable, or user-defined implementation of Value,
 // UnpackArgs performs the appropriate type check.
 // Predeclared Go integer types uses the AsInt check.
-// If the parameter name ends with "?",
-// it and all following parameters are optional.
+//
+// If the parameter name ends with "?", it's optional.
+//
+// If the parameter name ends with "??", it's optional and treats the None value
+// as if the argument was absent.
 //
 // If the variable implements Unpacker, its Unpack argument
 // is called with the argument value, allowing an application
@@ -89,9 +92,11 @@ func UnpackArgs(fnname string, args Tuple, kwargs []Tuple, pairs ...interface{})
 
 	paramName := func(x interface{}) string { // (no free variables)
 		name := x.(string)
-		if name[len(name)-1] == '?' {
-			name = name[:len(name)-1]
-		}
+
+		// name may end in "?" or "??"
+		name = strings.TrimSuffix(name, "?")
+		name = strings.TrimSuffix(name, "?")
+
 		return name
 	}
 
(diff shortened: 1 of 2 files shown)


starlark/unpack.go
```diff
@@ -24,8 +24,11 @@ type Unpacker interface {
 // Iterable, or user-defined implementation of Value,
 // UnpackArgs performs the appropriate type check.
 // Predeclared Go integer types uses the AsInt check.
-// If the parameter name ends with "?",
-// it and all following parameters are optional.
+//
+// If the parameter name ends with "?", it's optional.
```
alandonovan, line 28: The previous text prescribes that "it and all following parameters" are optional. Is that no longer the case?

The prevailing style here avoids enclitics: "it is".

starlark/unpack.go
```diff
@@ -173,14 +188,19 @@ func UnpackPositionalArgs(fnname string, args Tuple, kwargs []Tuple, min int, va
...
-		if err := unpackOneArg(arg, vars[i]); err != nil {
+		if err := unpackOneArg(arg, vars[i], ""); err != nil {
 			return fmt.Errorf("%s: for parameter %d: %s", fnname, i+1, err)
 		}
 	}
 	return nil
 }
 
-func unpackOneArg(v Value, ptr interface{}) error {
+func unpackOneArg(v Value, ptr interface{}, paramSpec string) error {
```
alandonovan, line 198: Let's pass `skipNone bool` here and move the check to the caller. This function has no business with the name.

starlark/unpack.go
```diff
@@ -89,9 +92,11 @@ func UnpackArgs(fnname string, args Tuple, kwargs []Tuple, pairs ...interface{})
 
 	paramName := func(x interface{}) string { // (no free variables)
 		name := x.(string)
-		if name[len(name)-1] == '?' {
-			name = name[:len(name)-1]
-		}
+
+		// name may end in "?" or "??"
+		name = strings.TrimSuffix(name, "?")
+		name = strings.TrimSuffix(name, "?")
```
alandonovan, line 98: Change this function so that it returns (name string, skipNone bool), and avoid the redundant checks at each call.

starlark/unpack.go
```diff
@@ -133,7 +138,7 @@ kwloop:
 	// (We needn't check the first len(args).)
 	for i := len(args); i < nparams; i++ {
 		name := pairs[2*i].(string)
-		if strings.HasSuffix(name, "?") {
+		if isOptionalParamSpec(name) {
```
alandonovan, line 141: Inline this call.

starlark/unpack.go
nicks: done

starlark/unpack.go
nicks: done

starlark/unpack.go
nicks: done

nicks:
@alandonovan did you have any other questions or concerns about this PR?

starlark/unpack.go
```diff
@@ -113,12 +128,20 @@ kwloop:
...
-			if paramName(pairs[2*i]) == string(name) {
+			pName, skipNone := paramName(pairs[2*i])
+			if pName == string(name) {
 				// found it
 				if defined.set(i) {
 					return fmt.Errorf("%s: got multiple values for keyword argument %s",
 						fnname, name)
 				}
+
+				if skipNone {
```
alandonovan, line 139: For the record: setting `defined[i]` before the skipNone check appears to be consistent with Python3. In other words, `f(x=None, **dict(x=1))` gives a "duplicate named parameter" error. The x=None parameter is not completely ignored.

alandonovan (approved):
LGTM. Thanks!

PR #357, opened 2021-03-05
add math module

### LINKED ISSUE #78: a math library
Starlark seem mighty terrific. I just have a minor query about doing some math.

In starlark-go, how would one raise n to the m-th power?

In Go, I would call math.Pow(n, m). In Python, math.pow(n, m).

Another way of asking: is there a way to get at the math library from Go perhaps?

b5:
closes #78.

cc @dustmop, @adonovan, & @essobedo


lib/math/math.go
```diff
@@ -0,0 +1,100 @@
+package math
```
adonovan, line 1: Package needs doc comment. Follow the example of the json package for what form this should take.

lib/math/math.go
```diff
@@ -0,0 +1,100 @@
...
+
+const tau = math.Pi * 2
+const oneRad = tau / 360
+
+// Module math is a Starlark module of math-related functions.
+var Module = &starlarkstruct.Module{
+	Name: "math",
+	Members: starlark.StringDict{
+		"ceil":  starlark.NewBuiltin("ceil", ceil),
+		"fabs":  starlark.NewBuiltin("fabs", fabs),
```
adonovan, line 18: Let's call this abs, and overload it for int and float.

lib/math/math.go
```diff
@@ -0,0 +1,100 @@
...
+		"acos":  starlark.NewBuiltin("acos", acos),
+		"asin":  starlark.NewBuiltin("asin", asin),
+		"atan":  starlark.NewBuiltin("atan", atan),
+		"atan2": starlark.NewBuiltin("atan2", atan2),
+		"cos":   starlark.NewBuiltin("cos", cos),
+		"hypot": starlark.NewBuiltin("hypot", hypot),
+		"sin":   starlark.NewBuiltin("sin", sin),
+		"tan":   starlark.NewBuiltin("tan", tan),
+
+		"degrees": starlark.NewBuiltin("degrees", degrees),
```
adonovan, line 34: These don't seem necessary. It's easy to say `theta*60/(2*pi)` on the rare occasion that it is needed.

lib/math/math.go
```diff
@@ -0,0 +1,100 @@
...
+		"acosh": starlark.NewBuiltin("acosh", acosh),
+		"asinh": starlark.NewBuiltin("asinh", asinh),
+		"atanh": starlark.NewBuiltin("atanh", atanh),
+		"cosh":  starlark.NewBuiltin("cosh", cosh),
+		"sinh":  starlark.NewBuiltin("sinh", sinh),
+		"tanh":  starlark.NewBuiltin("tanh", tanh),
+
+		"e":   starlark.Float(math.E),
+		"pi":  starlark.Float(math.Pi),
+		"tau": starlark.Float(tau),
```
adonovan, line 46: Not necessary. `2*math.pi` is clearer to most readers.

lib/math/math.go
```diff
@@ -0,0 +1,100 @@
...
+		"phi": starlark.Float(math.Phi),
+		"inf": starlark.Float(math.Inf(1)),
```
adonovan, line 48: These aren't needed. Users can say float("nan"), float("inf").

lib/math/math.go
```diff
@@ -0,0 +1,100 @@
...
+		"nan": starlark.Float(math.NaN()),
+	},
+}
+
+// floatFunc unpacks a starlark function call, calls a passed in float64 function
+// and returns the result as a starlark value
+func floatFunc(name string, args starlark.Tuple, kwargs []starlark.Tuple, fn func(float64) float64) (starlark.Value, error) {
+	var x starlark.Float
```
adonovan, line 56: var x float64

(and avoid a conversion below)

lib/math/math.go
```diff
@@ -0,0 +1,100 @@
...
+	if err := starlark.UnpackPositionalArgs(name, args, kwargs, 1, &x); err != nil {
+		return nil, err
+	}
+	return starlark.Float(fn(float64(x))), nil
+}
+
+// floatFunc2 is a 2-argument float func
+func floatFunc2(name string, args starlark.Tuple, kwargs []starlark.Tuple, fn func(float64, float64) float64) (starlark.Value, error) {
+	var x, y starlark.Float
```
adonovan, line 65: ditto (float64)

lib/math/math.go
```diff
@@ -0,0 +1,100 @@
...
+	if err := starlark.UnpackPositionalArgs(name, args, kwargs, 2, &x, &y); err != nil {
+		return nil, err
+	}
+	return starlark.Float(fn(float64(x), float64(y))), nil
+}
+
+// Return the floor of x, the largest integer less than or equal to x.
```
adonovan, line 72: // floor returns...

lib/math/trig.go
```diff
@@ -0,0 +1,92 @@
+package math
```
adonovan, line 1: No need to split this stuff across two files.

starlark/testdata/math.star
```diff
@@ -0,0 +1,43 @@
+# Tests of math module.
+
+load('math.star', 'math')
+load('assert.star', 'assert')
+
+def assert_near(actual, desired, threshold):
```
adonovan, line 6: This function doesn't actually assert, and can be simplified to:
```
def near(got, want, threshold):
  return math.abs(got-want) < threshold  
```

That said, we'd get more informative error messages if they were of the form "want x±k, got y".

starlark/testdata/math.star
```diff
@@ -0,0 +1,43 @@
...
+  val = desired - actual
+  if val < 0:
+    val = val * -1
+  return val < threshold
+
+assert.eq(math.ceil(0.5), 1.0)
```
adonovan, line 12: The tests should demonstrate the behavior of each function more thoroughly: does it accept int as well as float? What if the argument is 0, 0.4, 0.5, 1, 10, inf, nan, or the negation of each? What is the round mode?

starlark/testdata/math.star
```diff
@@ -0,0 +1,43 @@
...
+assert.eq(math.fabs(-2.0), 2.0)
+assert.eq(math.floor(0.5), 0)
+
+assert.eq(math.exp(2.0), 7.38905609893065)
+assert.eq(math.sqrt(4.0), 2.0)
+
+assert.eq(math.acos(1.0), 0)
+assert.eq(math.asin(1.0), 1.5707963267948966)
```
adonovan, line 20: Ideally the test cases would be expressed to provide some mathematical insight,
e.g. `sin(math.pi / 6) = 0.5`

srebhan (commented):
Hooray another one! :-) Thanks @b5 for pushing this!
Looks almost perfect. Maybe just rename `floatFunc` to `floatFunc1` to match the other name and add the `in radians` comment to the documentation of all trigonometric functions. It's missing for some functions at the end of the file. But it's only my two cent...

lib/math/math.go
srebhan: Please keep them as code using it is much easier to understand.

essobedo:
@b5 I take it over

essobedo:
@googlebot I consent.

essobedo:
@adonovan I believe that I have made all the requested changes except [`Let's call this abs, and overload it for int and float.`](https://github.com/google/starlark-go/pull/357#discussion_r588602487). I tried something regarding this remark, but I don't like it, could you please tell me what you had in mind exactly?

lib/math/math.go
```diff
@@ -0,0 +1,198 @@
...
+var (
+	toDeg = func(x float64) float64 { return x / oneRad }
+	toRad = func(x float64) float64 { return x * oneRad }
+)
+
+// Module math is a Starlark module of math-related functions and constants.
+var Module = &starlarkstruct.Module{
+	Name: "math",
+	Members: starlark.StringDict{
+		"abs":   starlark.NewBuiltin("abs", abs),
```
adonovan, line 29: We can simplify this file substantially by defining a helper function that combines the call to NewBuiltin and oneArgFunc, like so:

```
Members: starlark.StringDict{
		"abs":   newUnaryBuiltin("abs", math.Abs),
                 ...
}

// newUnaryBuiltin wraps a unary floating-point Go function
// as a Starlark built-in that accepts int or float arguments.
func newUnaryBuiltin(name string, fn func(float64) float64) *starlark.Builtin {
    return starlark.NewBuiltin(name, func(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
        ...body of oneArgFunc...
    })
}
```

(Ditty for newBinaryFunction.)

This makes the table do more more, and avoids the need for lots of different wrappers, and it also allows us to consolidate our documentation at Module, instead of smearing it across all the unexported wrappers.

essobedo:
@adonovan Fixed, please check again

lib/math/math.go
```diff
@@ -0,0 +1,142 @@
...
+//     atanh(x) - Returns the inverse hyperbolic tangent of x.
+//     cosh(x) - Returns the hyperbolic cosine of x.
+//     sinh(x) - Returns the hyperbolic sine of x.
+//     tanh(x) - Returns the hyperbolic tangent of x.
+//
+// All function accept both int and float values as arguments.
+//
+// It defines the following constants:
+//
+//     e - The number e, also known as Euler's number, is a mathematical constant approximately equal to 2.71828.
```
adonovan, line 55: `e - The base of natural logarithms, approximately 2.71828.`

lib/math/math.go
```diff
@@ -0,0 +1,142 @@
...
+//     phi - The number phi, also known as the Golden Ratio, is a mathematical constant approximately equal to 1.61803.
```
adonovan, line 56: `phi - The Golden Ratio, (1+ sqrt(5))/2, or approximately 1.61803.`

lib/math/math.go
```diff
@@ -0,0 +1,142 @@
...
+//     pi - The number pi, also known as Archimedes' constant, is a mathematical constant approximately equal to 3.14159.
```
adonovan, line 57: pi - The ratio of a circle's circumference to its diameter, approximately 3.14159.

lib/math/math.go
```diff
@@ -0,0 +1,142 @@
...
+//     acosh(x) - Returns the inverse hyperbolic cosine of x.
+//     asinh(x) - Returns the inverse hyperbolic sine of x.
+//     atanh(x) - Returns the inverse hyperbolic tangent of x.
+//     cosh(x) - Returns the hyperbolic cosine of x.
+//     sinh(x) - Returns the hyperbolic sine of x.
+//     tanh(x) - Returns the hyperbolic tangent of x.
+//
+// All function accept both int and float values as arguments.
+//
+// It defines the following constants:
```
adonovan, line 53: `The module also defines approximations of the following constants:`

lib/math/math.go
adonovan: Speaking of which, were are the log10, log2, and ln functions?

lib/math/math.go
```diff
@@ -0,0 +1,142 @@
...
+		"tanh":  newUnaryBuiltin("tanh", math.Tanh),
+
+		"e":   starlark.Float(math.E),
+		"phi": starlark.Float(math.Phi),
+		"pi":  starlark.Float(math.Pi),
+	},
+}
+
+// The funcParam type of all the math functions.
+type funcParam float64
```
adonovan, line 96: // floatOrInt is an Unpacker that converts a Starlark int or float to Go's float64.
type floatOrInt

lib/math/math.go
```diff
@@ -0,0 +1,142 @@
...
+
+// Unpack is a custom argument unpacker
+func (p *funcParam) Unpack(v starlark.Value) error {
+	switch x := v.(type) {
```
adonovan, line 100: Use `switch v := v` unless there's a reason not to. This reduces the number of names and avoids accidental reference to v within the switch.

lib/math/math.go
```diff
@@ -0,0 +1,142 @@
...
+	case starlark.Int:
+		*p = funcParam(x.Float())
+		return nil
+	case starlark.Float:
+		*p = funcParam(x)
+		return nil
+	}
+
+	return fmt.Errorf("got %s, want a float or an int", v.Type())
```
adonovan, line 109: `want float or int` is how we usually write this.

lib/math/math.go
```diff
@@ -0,0 +1,142 @@
...
+	return starlark.NewBuiltin(name, func(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+		var x, y funcParam
+		if err := starlark.UnpackPositionalArgs(name, args, kwargs, 2, &x, &y); err != nil {
+			return nil, err
+		}
+		return starlark.Float(fn(float64(x), float64(y))), nil
+	})
+}
+
+func toDeg(x float64) float64 {
```
adonovan, line 136:
```
func degrees(radians float64)
func radians(degrees float64)
```

starlark/testdata/math.star
```diff
@@ -0,0 +1,294 @@
+# Tests of math module.
+
```
adonovan, line 2:
```
inf. nan = float("inf"), float("nan")
```
and factor.

essobedo:
@adonovan all fixed, please check again

essobedo:
It sounds that there is an issue with the build, the status is never reported

lib/math/math.go
```diff
@@ -0,0 +1,150 @@
...
+		"e":   starlark.Float(math.E),
+		"phi": starlark.Float(math.Phi),
+		"pi":  starlark.Float(math.Pi),
+	},
+}
+
+// floatOrInt is an Unpacker that converts a Starlark int or float to Go's float64.
+type floatOrInt float64
+
+// Unpack is a custom argument unpacker
```
adonovan, line 106: delete comment

lib/math/math.go
```diff
@@ -0,0 +1,150 @@
...
+		"radians": newUnaryBuiltin("radians", radians),
+
+		"acosh": newUnaryBuiltin("acosh", math.Acosh),
+		"asinh": newUnaryBuiltin("asinh", math.Asinh),
+		"atanh": newUnaryBuiltin("atanh", math.Atanh),
+		"cosh":  newUnaryBuiltin("cosh", math.Cosh),
+		"sinh":  newUnaryBuiltin("sinh", math.Sinh),
+		"tanh":  newUnaryBuiltin("tanh", math.Tanh),
+
+		"log":   newUnaryBuiltin("log", math.Log),
```
adonovan, line 93: Python's math.log function accepts an optional base argument (default=e), and so should Starlark's. The body of the function is `math.Log(x) / math.Log(base)`, but you'll need to specialize newBinaryFunction because of the optional argument.

lib/math/math.go
```diff
@@ -0,0 +1,150 @@
...
+//     log(x) - Returns the natural logarithm of x.
+//     log2(x) - Returns the decimal logarithm of x. The special cases are the same as for Log.
+//     log10(x) - Returns the natural logarithm of 1 plus its argument x. It is more accurate than Log(1 + x) when x is near.
+//
+// All function accept both int and float values as arguments.
+//
+// The module also defines approximations of the following constants:
+//
+//     e - The base of natural logarithms, approximately 2.71828.
+//     phi - The golden ratio, (1 + sqrt(5)) / 2, approximately 1.61803.
```
adonovan, line 60: Let's delete phi. It's neither well known nor widely needed, it's not part of Python's math library, and it's trivial for users to define themselves.

essobedo:
@adonovan done, please check it again

lib/math/math.go
```diff
@@ -0,0 +1,176 @@
...
+		"floor": newUnaryBuiltin("floor", math.Floor),
+		"round": newUnaryBuiltin("round", math.Round),
+
+		"exp":  newUnaryBuiltin("exp", math.Exp),
+		"sqrt": newUnaryBuiltin("sqrt", math.Sqrt),
+
+		"acos":  newUnaryBuiltin("acos", math.Acos),
+		"asin":  newUnaryBuiltin("asin", math.Asin),
+		"atan":  newUnaryBuiltin("atan", math.Atan),
+		"atan2": newBinaryFunction("atan2", math.Atan2, unpackRegularArgs),
```
adonovan, line 75: Please keep these cases and newBinaryFunction as they were. Log is the odd one out, so expand it out, instead of complicating the simple functions.

lib/math/math.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+func radians(x float64) float64 {
+	return 2 * math.Pi * x / 360
+}
+
+func log(x float64, base float64) float64 {
+	num := math.Log(x)
+	if base == 1 {
+		if num < 0 {
```
adonovan, line 170: Python reports an error (division by zero) if base == 1, and so too should Starlark.

essobedo:
@adonovan Done, I also added some new functions

lib/math/math.go
adonovan: Please expand out the implementation of log in full, and don't complicate all the other functions.

lib/math/math.go
adonovan: Still not done.

essobedo:
@adonovan done, please check again

lib/math/math.go
```diff
@@ -0,0 +1,181 @@
...
+		"radians": newUnaryBuiltin("radians", radians),
+
+		"acosh": newUnaryBuiltin("acosh", math.Acosh),
+		"asinh": newUnaryBuiltin("asinh", math.Asinh),
+		"atanh": newUnaryBuiltin("atanh", math.Atanh),
+		"cosh":  newUnaryBuiltin("cosh", math.Cosh),
+		"sinh":  newUnaryBuiltin("sinh", math.Sinh),
+		"tanh":  newUnaryBuiltin("tanh", math.Tanh),
+
+		"log": newLogFunction(),
```
adonovan, line 101: Inline this called-once function, so that the table entry is:
`"log": starlark.NewBuiltin("log", log),`

lib/math/math.go
```diff
@@ -0,0 +1,181 @@
...
+			return nil, err
+		}
+		return starlark.Float(fn(float64(x), float64(y))), nil
+	})
+}
+
+//  newLogFunction wraps the Log function
+// as a Starlark built-in that accepts int or float arguments.
+func newLogFunction() *starlark.Builtin {
+	return starlark.NewBuiltin("log", func(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
```
adonovan, line 152:
```
func log(Thread, Builtin, Tuple, Tuple) (Value, error) {
  ...
}
```

lib/math/math.go
```diff
@@ -0,0 +1,181 @@
...
+		var (
+			x    floatOrInt
+			base = floatOrInt(math.E)
```
adonovan, line 155: base floatOrInt = math.E

lib/math/math.go
```diff
@@ -0,0 +1,181 @@
...
+		)
+		if err := starlark.UnpackArgs("log", args, kwargs, "x", &x, "base?", &base); err != nil {
```
adonovan, line 157: UnpackPositionalArgs

(so that the names x and base don't leak into the contract)

lib/math/math.go
```diff
@@ -0,0 +1,181 @@
...
+			return nil, err
+		}
+		result, err := log(float64(x), float64(base))
```
adonovan, line 160: Inline this called-once function.

adonovan (commented):
Nearly there!

essobedo:
@adonovan done

adonovan (approved):
Looks good. Many thanks.

adonovan:
Jon, can you merge this PR? It LGTM.

b5:
Thanks again @adonovan & @essobedo! Excited to keep these moving

brandjon:
The status check for CI never returned. Can you try force-pushing a new commit (e.g. with `git commit --amend --no-edit`), as suggested [here](https://stackoverflow.com/questions/52200096/github-pull-request-waiting-for-status-to-be-reported)?

brandjon:
Actually looks like our travis CI integration method is [deprecated](https://travis-ci.community/t/known-issue-travis-ci-reports-expected-waiting-for-status-to-be-reported-on-the-github-status-api-but-the-status-never-arrives/1154). Will reconfigure on my side...

brandjon:
Actually it looks like we've received status updates recently from this hook and the deprecation that article's talking about happened a while ago, so let's just force push and see if that fixes it.

b5:
ok, force pushed, doesn't look like we're having any luck ☹️

fenollp:
Hi! Sorry to barge in a little late.
I see the `type floatOrInt float64` was introduced to `Unpack` arguments of all builtins.
Are we not loosing precision here? We seem to be discarding anything that doesn't fit in `float64` (so big.Ints).
But I don't see this mentioned. Did I miss it?

Am I correct in assuming that `Unpack`ing a litteral bigint into `float64` returns an error?
```python
abs(9007199254740992+1) # = abs(1+2^53) does this fail?
```

adonovan:
> I see the `type floatOrInt float64` was introduced to `Unpack` arguments of all builtins.
> Are we not loosing precision here? We seem to be discarding anything that doesn't fit in `float64` (so big.Ints).

An int of any magnitude is converted to a float. This will fail if the magnitude is bigger than about 1e308. However, the result is always a floating-point number, even if the argument was an integer.

This is consistent with Python3, but it does argue in favor of renaming math.abs to math.fabs, both for consistency with Python, and to indicate that the result is a float. I'm not aware of how to compute abs of a big int in Python.

tetromino:
I manually kicked off the travis build, and it finally worked.

But now it seems we are missing the google cla check from one of the authors.

@essobedo and @b5 - please see [googlebot's comment above](https://github.com/google/starlark-go/pull/357#issuecomment)-791891376 - you would need to make a comment with the literal words `@googlebot I consent.` to agree to the cla.

tetromino:
Never mind, googlebot is now happy :)

brandjon:
Thanks for taking care of the merge tetromino@.

PR #361, opened 2021-03-11
Port pytruth as lib/truth

fenollp:
Hi there,
I've just finished porting https://github.com/google/pytruth to Starlark for use in https://github.com/FuzzyMonkeyCo/monkey which requires expressive assertions.

I understand these are the points that should require discussion (some/all are mentioned in the package docs):
* starlark strings aren't iterable but are sometimes turned into rune tuples in this lib for convenience on the user of the lib
* duplicate counter compares string serialization of values instead of hash. (pytruth works around this)

Please share your thoughts and take some time to do a bit of reviewing :)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,8 @@ require (
 	github.com/chzyer/readline v0.0.0-20180603132655-2972be24d48e
 	github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1 // indirect
 	github.com/google/go-cmp v0.5.1 // indirect
+	github.com/pmezard/go-difflib v1.0.0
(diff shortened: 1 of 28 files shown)


tetromino:
There are 3 different Starlark implementations: this one in Go, the Java one built into Bazel, and the [Rust one](https://github.com/facebookexperimental/starlark-rust). And my immediate reaction to adding a major feature (a new module) to just the Go implementation would be negative.

Is it possible to write something good enough for your use case in pure Starlark and distribute it as a separate package that any Starlark interpreter can use?

If it's not possible, what is a minimal set of new interpreter features that you would need in order to be able to write it in pure Starlark?

Issue #241, opened 2019-09-09
Any plan on adding some support for regular expressions to the language?


fenollp:
This has many parts:
* a builtin type
* the usual methods
* special syntax (sigils, ...)
* what set of regexps to support
* what amount of backtracking should be allowed if at all

One can definitely implement the first 2 with minimum friction, but as I find Starlark's strength to be in its Turing-incompleteness guarantees I am wondering if this was / is being discussed as an addition to the language.

alandonovan:
After `time` and `json`,

time https://github.com/google/starlark-go/tree/wip-skylark-time
json https://github.com/google/starlark-go/tree/json

 `regexp` is the most requested package for the Starlark "standard library". No-one has sketched a prototype yet, but it shouldn't be hard to come up with a sound Starlark API for regular expressions that can be backed by the Go, Java, and C++ implementations of RE2, which is linear time (non-backtracking; see github.com/google/re2), though we will need to disallow `\C`.

I imagine the basic API would look something like the Go regexl API:
```
   load("regexp")
   rx = regexp.compile("b(an)*a")
   rx.matches(str src) # returns bool
   rx.find(str src) # returns text of match
   rx.find_all(str src, int max) # returns list of matches
   rx.replace_all(str src , str or function) # returns replaced text
   rx.find_submatches(str) # returns list of complete match and each subpattern match
```
We might also want variants that return the indices (not text) of each match.

KOGI:
Any update on this, yet?

alandonovan:
Sorry, no.

tv42:
For what it's worth, I stumbled on https://godoc.org/github.com/qri-io/starlib/re

essobedo:
@b5 What about this one? Ready for it? :-)

b5:
I think this would be great to get to attack next given the roll we're on with the `time` and `math` packages, but the design & implementation of any regex package or language level support requires more of upfront design discussion:

1. design an `re` package? write a proposal for inclusion in the language? both?
2. If we're designing a package, does it take inspiration from go's `regexp` or python's `re`?
3. What's providing the underlying regex parser in go?

My answers to these questions:
1. It should be a package. We should avoid introducing overhead into the language, even if it adds parity to python. I'm thinking specifically of the cost of implementing a starlark interpreter in another language  that doesn't have easy support for RE2-style regex syntax. Having it as a package makes it easier to opt out. 
[...]

PR #365, opened 2021-03-20
Add support of timestamps in nanoseconds

### LINKED ISSUE #364: Add support of timestamps in nanoseconds
The new built in function `from_timestamp` of the `time` module doesn't support nanoseconds which is inconsistent with other functions of the module and too limited. The goal of this ticket is to add the ability to convert timestamps in nanoseconds into date time.

essobedo:
fixes #364 

## Motivation

For many real life use cases, it is important to be a able to support timestamps expressed in nanoseconds.

## Modifications:

* Overloads the function `from_timestamp` to be able to set nanoseconds.
* Fixes the documentation of the built in function `from_timestamp` 
* Improves the documentation (not related to this issue)
* Adds a test for the function `now` since it was not covered by any tests (not related to this issue)

### DIFF
--- a/lib/time/time.go
+++ b/lib/time/time.go
@@ -20,6 +20,7 @@ var Module = &starlarkstruct.Module{
 		"time":              starlark.NewBuiltin("time", newTime),
 		"parse_time":        starlark.NewBuiltin("parse_time", parseTime),
 		"from_timestamp":    starlark.NewBuiltin("from_timestamp", fromTimestamp),
+		"from_timestamp_ns": starlark.NewBuiltin("from_timestamp_ns", fromTimestampNanos),
 
 		"nanosecond":  Duration(time.Nanosecond),
 		"microsecond": Duration(time.Microsecond),
@@ -89,14 +90,24 @@ func parseTime(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple
 	return Time(t), nil
 }
 
-// Parses the given timestamp string corresponding to
-// the total amount of seconds since January 1, 1970 UTC.
+// Converts the given timestamp corresponding to the total amount of seconds
+// since January 1, 1970 UTC into an object of type Time.
 func fromTimestamp(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
 	var x int64
 	if err := starlark.UnpackPositionalArgs("from_timestamp", args, kwargs, 1, &x); err != nil {
 		return nil, err
 	}
 	return Time(time.Unix(x, 0)), nil
+}
+
+// Converts the given timestamp corresponding to the total amount of nanoseconds
+// since January 1, 1970 UTC into an object of type Time.
+func fromTimestampNanos(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var x int64
+	if err := starlark.UnpackPositionalArgs("from_timestamp_ns", args, kwargs, 1, &x); err != nil {
+		return nil, err
+	}
+	return Time(time.Unix(x/1e9, x%1e9)), nil
 }
 
 // Generates the current time current time.
--- a/lib/time/time.go
+++ b/lib/time/time.go
@@ -1,4 +1,9 @@
-package time
+// Copyright 2021 The Bazel Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package time provides time-related constants and functions.
+package time // import "go.starlark.net/lib/time"
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


essobedo:
Hi @adonovan and @brandjon WDYT of this one?

essobedo:
Please note that I intentionally avoided to overload the function `from_timestamp` for clarity and simplicity

lib/time/time.go
```diff
@@ -99,6 +100,16 @@ func fromTimestamp(thread *starlark.Thread, _ *starlark.Builtin, args starlark.T
 	return Time(time.Unix(x, 0)), nil
 }
 
+// Converts the given timestamp corresponding to the total amount of nanoseconds
```
adonovan, line 103: A timestamp (aka UNIX time_t) is defined as the number of seconds since Jan 1, 1970, which may be an integer, or a fixed-point number (pair of int64) if sub-second precision is needed. It would be confusing for this function to introduce the new concept of a "timestamp" that is a quantity of nanoseconds.

Instead, change the existing from_timestamp function so that it accepts either one int (a number of seconds) or two ints (sec and nsec). It could additionally accept one float, a number of seconds. There is no need for a new function.

Also, the functions provided by this module should be documented at `Module`. I realize this is a preexisting problem, but would you mind addressing it while you're working here? (Look at json and math for examples.)

adonovan (changes requested):

lib/time/time.go
essobedo: @adonovan I understand your point of view even if it is not really helpful in my use case since I have timestamps expressed in nanoseconds so I will need to explicitly split seconds from nanoseconds which is not really convenient. Anyway I addressed your remarks.

lib/time/time.go
```diff
@@ -10,16 +15,54 @@ import (
 	"go.starlark.net/syntax"
 )
 
-// Module time is a Starlark module of time-related functions.
+// Module time is a Starlark module of time-related functions and constants.
+// The module defines the following functions:
+//
+//     from_timestamp(sec, nsec) - Converts the given Unix time corresponding to the total amount of seconds
+//                                 and nonoseconds since January 1, 1970 UTC into an object of type Time.
```
adonovan, line 22: "nano" (throughout)

lib/time/time.go
```diff
@@ -10,16 +15,54 @@ import (
 	"go.starlark.net/syntax"
 )
 
-// Module time is a Starlark module of time-related functions.
+// Module time is a Starlark module of time-related functions and constants.
+// The module defines the following functions:
+//
+//     from_timestamp(sec, nsec) - Converts the given Unix time corresponding to the total amount of seconds
```
adonovan, line 21: s/amount/number/

Use "number" for countable nouns, and "amount" for mass nouns (e.g. amount of time).

lib/time/time.go
```diff
@@ -10,16 +15,54 @@ import (
...
+//                                 and nonoseconds since January 1, 1970 UTC into an object of type Time.
+//                                 The total amount of nonoseconds is optional.
```
adonovan, line 23: This line can be deleted if you make the previous line read

"number of seconds and (optionally) nanoseconds"

lib/time/time.go
```diff
@@ -10,16 +15,54 @@ import (
...
+//                                 It is valid to pass nonoseconds outside the range [0, 999999999].
+//                                 Not all sec values have a corresponding time value. One such
```
adonovan, line 25: The function should return an error, not an invalid value, if the argument is out of range.

lib/time/time.go
```diff
@@ -10,16 +15,54 @@ import (
...
+//                                 value is 1<<63-1 (the largest int64 value).
+//
+//     is_valid_timezone(loc) - Indicates whether the given name of location is valid or not.
```
adonovan, line 28: "is_valid_timezone(loc) - Reports whether loc is a valid time zone name."

(And delete the second sentence.)

lib/time/time.go
```diff
@@ -10,16 +15,54 @@ import (
...
+//                              Returns true if it is valid, false otherwise.
+//
+//     now() - Returns the current local time.
```
adonovan, line 31: Add: Applications may replace this function by a deterministic one.

lib/time/time.go
```diff
@@ -10,16 +15,54 @@ import (
...
+//
+//     parse_duration(d) - Parses the given duration string.
+//                         A duration string is a possibly signed sequence of decimal numbers,
```
adonovan, line 34: The adjective "signed" applies to numbers, not character strings. Remove "possibly signed" and instead add:

A "-" prefix indicates a negative duration.

lib/time/time.go
```diff
@@ -10,16 +15,54 @@ import (
...
+//                         each with optional fraction and a unit suffix, such as "300ms", "-1.5h" or "2h45m".
+//                         Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
+//
+//     parseTime(x, format, location) - Parses the given time string using a specific time format and location.
+//                                      The expected arguments are a time string (mandatory), a time format
+//                                      (optional, set to RFC3339 by default) and a name of location
```
adonovan, line 40: Give an example of RFC 3339 format, e.g. "2021-03-22T23:20:50.52Z"

lib/time/time.go
```diff
@@ -10,16 +15,54 @@ import (
...
+//                                      (optional set to UTC by default).
```
adonovan, line 41: "optional, "

lib/time/time.go
```diff
@@ -10,16 +15,54 @@ import (
...
+//
+//     time(year, month, day, hour, minute, second, nanosecond, location) - Returns the Time corresponding to
+//	                                                                        yyyy-mm-dd hh:mm:ss + nsec nanoseconds
+//                                                                          in the appropriate zone for that time
+//                                                                          in the given location. All the parameters
+//                                                                          are optional.
+// The module also defines the following constants:
+//
+// nanosecond - A duration representing one nanosecond.
```
adonovan, line 50: Indent each line by at least three spaces so that it appears in godoc as a block quote.

lib/time/time.go
```diff
@@ -10,16 +15,54 @@ import (
...
+//                              Returns true if it is valid, false otherwise.
+//
+//     now() - Returns the current local time.
+//
+//     parse_duration(d) - Parses the given duration string.
+//                         A duration string is a possibly signed sequence of decimal numbers,
+//                         each with optional fraction and a unit suffix, such as "300ms", "-1.5h" or "2h45m".
+//                         Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
+//
+//     parseTime(x, format, location) - Parses the given time string using a specific time format and location.
```
adonovan, line 38: parse_time

The design of this function is quite Go-specific, and I wonder whether it was a good idea to expose it directly. In any case, it needs more documentation about how the format string is interpreted. For now, let's redirect readers to Go's time.Parse function, since that is the actual behavior, though this is not ideal for a Starlark spec.

adonovan (commented):
Thanks!

lib/time/time.go
essobedo: Oops, good catch

adonovan (approved):
@brandjon LGTM, ready for you to merge.

ssoroka (approved):

essobedo:
@brandjon Is-this PR good enough to be merged?

essobedo:
@brandjon Could you please at least tell me if this project won't be maintained anymore? Because I wanted to work on https://github.com/google/starlark-go/issues/241 during my spare time, once this PR will be merged, if this project is no more supported/maintained I won't do it. Thank you in advance for your answer 🙏

brandjon:
Apologies for letting this slip. Alan has write access to this repo and I've confirmed with him that he can merge.

PR #369, opened 2021-04-23
feat: introduce regexp package

### LINKED ISSUE #241: Any plan on adding some support for regular expressions to the language?

b5:
DRAFT. NOT READY FOR REVIEW.

@essobedo is going to be heading up turning this code into a package that closes #241. We're using the `re` package from [starlib](https:/github.com/qri-io/starlib) as a starting point, but it needs a _lot_ of work before it's ready for initial review.

@essobedo I think at a minimum we should:

* [ ] switch the defined here to match the API defined in #241
* [ ] remove `regexp/doc.go` (I've kept it only so you have context)
* [ ] document `regexp/regexp.go`
* [ ] completely re-write the test suite to test the new API
* [ ] add a test that confirms failure when attempting to compile a regular expression that that uses byte-ordered features:  `\C`


essobedo:
@b5 perfect, thank you

essobedo:
@googlebot I consent

essobedo:
Sorry for the delay, I was very busy these days 

@b5 I believe that it is ready for a first review since I addressed everything that needed to be done so, could you please "un-draft it" 🙏 

@adonovan Could you please review it? Thx in advance 🙏

essobedo:
@adonovan 

> This isn't correct: if the backslash is preceded by a backslash, then it should be legal. You will need to use a stripped down version of the scanning algorithm from the regexp package.

I double checked and unless I really missed your point, if I don't precede `\C` with a backslash, I end up with the error `invalid escape sequence \C`. I have just added two examples one with the patterns `\d`, `\w` and `\s` all preceded by a backslash to show that it works as expected so the code seems to be actually correct. 

Regarding the code, I only changed the code to use a raw string literal instead of a string literal to avoid having to escape the backslash but the logic is the same.

essobedo:
@brandjon could you please do the review 🙏 ? @adonovan seems to be busy and/or not available

adonovan:
Sorry for the delay, will take a look presently.

lib/regexp/regexp.go
```diff
@@ -0,0 +1,258 @@
+// Copyright 2021 The Bazel Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package regexp provides regular expression related functions.
```
adonovan, line 5: missing hyphen

Better, put the emphasis at the end: "functions related to regular expressions".

We should state that it uses RE2 specifically, as that implies certain bounds on running time and certain restrictions on features.

lib/regexp/regexp.go
```diff
@@ -0,0 +1,258 @@
...
+	"log"
+	"regexp"
+	"sort"
+	"strings"
+
+	"go.starlark.net/starlark"
+	"go.starlark.net/starlarkstruct"
+)
+
+// Module regexp is a Starlark module of regular expression related functions.
```
adonovan, line 19: ditto

lib/regexp/regexp.go
```diff
@@ -0,0 +1,258 @@
...
+// The module defines the following functions:
+//
+//     compile(pattern) - Compile a regular expression pattern into a regular expression object, which
```
adonovan, line 22: "Compiles" (present not imperative)

"compiles a pattern in RE2 syntax (https://github.com/google/re2/wiki/Syntax) to a value of type 'regexp'"
(avoids 'object', and duplication of "regular expression", and states the type of the value)

Should we describe the equivalence relation as:
"Each call to compile returns a distinct regexp value." ?
(The other choices are: specify equivalence based on pattern strings, or leave it unspecified. My preference is the wording above.)

"matches, find, find_all, and other methods"
(parens f() indicate a call to a function named f)

lib/regexp/regexp.go
```diff
@@ -0,0 +1,258 @@
...
+//                        can be used for matching using its matches(), find(), find_all and other methods.
+//
+//     re.find(src) - Returns a string holding the text of the leftmost match in the given string of the regular expression re.
+//                    If there is no match, the return value is an empty string, but it will also be empty if the
```
adonovan, line 26: The result is "" if there is no match or if the pattern successfully matches an empty string.

lib/regexp/regexp.go
```diff
@@ -0,0 +1,258 @@
...
+//                    regular expression successfully matches an empty string.
+//
+//     re.find_all(src, max) - Returns a tuple of all successive matches of the regular expression re. An empty tuple indicates
```
adonovan, line 29: Tuples are typically used like anonymous structs, where there are a fixed number of components identified by number. If the length can be arbitrary (including 0 or, confusingly, 1) a list is more appropriate.


"Returns a new, mutable list..."

lib/regexp/regexp.go
```diff
@@ -0,0 +1,258 @@
...
+//                             no match. If max > 0, at most max strings will be returned. If max == 0,
```
adonovan, line 30: "are returned" (use present tense, throughout)

lib/regexp/regexp.go
```diff
@@ -0,0 +1,258 @@
...
+//                             an empty tuple will be returned. If max < 0, all strings will be returned.
+//                             The parameter max is optional, by default no limit will be applied.
```
adonovan, line 32: s/,/:/

lib/regexp/regexp.go
```diff
@@ -0,0 +1,258 @@
...
+//
+//     re.find_submatches(src) - Returns a tuple of strings holding the text of the leftmost match of the regular expression re
+//                               in the given string and the matches, if any, of its subexpressions. An empty tuple indicates
+//                               no match.
+//
+//     re.matches(src) - Indicates whether the given string contains any match of the regular expression re.
```
adonovan, line 38: s/Indicates/Reports/

lib/regexp/regexp.go
```diff
@@ -0,0 +1,258 @@
...
+	"go.starlark.net/starlarkstruct"
+)
+
+// Module regexp is a Starlark module of regular expression related functions.
+// The module defines the following functions:
+//
+//     compile(pattern) - Compile a regular expression pattern into a regular expression object, which
+//                        can be used for matching using its matches(), find(), find_all and other methods.
+//
+//     re.find(src) - Returns a string holding the text of the leftmost match in the given string of the regular expression re.
```
adonovan, line 25: s/re/regexp/g   (since that's the name of the datatype)

lib/regexp/regexp.go
```diff
@@ -0,0 +1,258 @@
...
+//                             The parameter max is optional, by default no limit will be applied.
+//
+//     re.find_submatches(src) - Returns a tuple of strings holding the text of the leftmost match of the regular expression re
+//                               in the given string and the matches, if any, of its subexpressions. An empty tuple indicates
+//                               no match.
+//
+//     re.matches(src) - Indicates whether the given string contains any match of the regular expression re.
+//
+//     re.replace_all(src , repl) - Returns a copy of the given string, replacing matches of the regular expression re
+//                                  with the replacement string repl. Inside repl, $ signs are interpreted as in Expand,
```
adonovan, line 41: What is Expand?

What's with the $ treatment? This is not a feature of RE2 (and other RE packages use \1 notation for them).

lib/regexp/regexp.go
```diff
@@ -0,0 +1,258 @@
...
+func compile(_ *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var (
+		pattern string
+	)
+
+	if err := starlark.UnpackPositionalArgs(b.Name(), args, kwargs, 1, &pattern); err != nil {
+		return starlark.None, err
+	}
+
+	if strings.Contains(pattern, `\C`) {
```
adonovan, line 71: This is not the correct check, as it will incorrectly reject `\\C` (two backslashes).
It is necessary to build the rudiments of a scanner for RE2 strings to correctly detect this case.

lib/regexp/regexp.go
```diff
@@ -0,0 +1,258 @@
...
+
+	re, err := regexp.Compile(pattern)
+	if err != nil {
+		return nil, err
+	}
+	return &Regex{re: re}, nil
+}
+
+func toTuple(strs []string) starlark.Tuple {
+	var vals starlark.Tuple
```
adonovan, line 83: Preallocate the correct space.

lib/regexp/regexp.go
```diff
@@ -0,0 +1,258 @@
...
+	for _, s := range strs {
+		vals = append(vals, starlark.String(s))
+	}
+	return vals
+}
+
+// hashString computes the FNV hash of s.
+func hashString(s string) uint32 {
```
adonovan, line 91: Call starlark.String.Hash?

lib/regexp/regexp.go
```diff
@@ -0,0 +1,258 @@
...
+//                             an empty tuple will be returned. If max < 0, all strings will be returned.
+//                             The parameter max is optional, by default no limit will be applied.
+//
+//     re.find_submatches(src) - Returns a tuple of strings holding the text of the leftmost match of the regular expression re
+//                               in the given string and the matches, if any, of its subexpressions. An empty tuple indicates
+//                               no match.
+//
+//     re.matches(src) - Indicates whether the given string contains any match of the regular expression re.
+//
+//     re.replace_all(src , repl) - Returns a copy of the given string, replacing matches of the regular expression re
```
adonovan, line 40: no space before comma

lib/regexp/regexp.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+
+	re, err := regexp.Compile(pattern)
+	if err != nil {
+		return nil, err
+	}
+	return &Regex{re: re}, nil
+}
+
+func toTuple(strs []string) starlark.Tuple {
```
adonovan, line 82: toList (see comments above)

lib/regexp/regexp.go
```diff
@@ -0,0 +1,258 @@
...
+func hashString(s string) uint32 {
+	var h uint32
+	for i := 0; i < len(s); i++ {
+		h ^= uint32(s[i])
+		h *= 16777619
+	}
+	return h
+}
+
+// Regex is a starlark representation of a compiled regular expression
```
adonovan, line 100: "Regexp" (to match the data type name "regexp" and the Go package name).

All sentences should end with a period.

Starlark has a capital S.

Better: "A Regexp represents a compiled RE2 regular expression."
The data type should 
Starlark

lib/regexp/regexp.go
```diff
@@ -0,0 +1,258 @@
...
+// The interface regex presents to the starlark runtime renders it immutable,
+// making this a no-op
+func (r *Regex) Freeze() {}
+
+// Hash returns a function of x such that Equals(x, y) => Hash(x) == Hash(y)
+// required by starlark.Value interface
+func (r *Regex) Hash() (uint32, error) { return hashString(r.re.String()), nil }
+
+// Truth returns the truth value of an object required by starlark.Value
+// interface. Any non-empty regexp is considered truthy
```
adonovan, line 121: "Non-empty" is ambiguous: do you mean the regular language is empty, or that the pattern string is empty?

I can't see how this feature would be useful. Let's say regexp values are always True (and document this at the top).

lib/regexp/regexp.go
```diff
@@ -0,0 +1,258 @@
...
+	"replace_all":     starlark.NewBuiltin("replace_all", replaceAll),
+	"split":           starlark.NewBuiltin("split", split),
+}
+
+// AttrNames lists available dot expression strings for time. required by
+// starlark.HasAttrs interface
+func (r *Regex) AttrNames() []string { return builtinAttrNames(regexMethods) }
+
+func matches(_ *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var (
```
adonovan, line 144: var src string

and delete blank

lib/regexp/regexp.go
```diff
@@ -0,0 +1,258 @@
...
+		src string
+	)
+
+	if err := starlark.UnpackPositionalArgs(b.Name(), args, kwargs, 1, &src); err != nil {
+		return starlark.None, err
```
adonovan, line 149: return nil, err

If err is non-nil, the first component is ignored.

lib/regexp/regexp.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+
+	re := b.Receiver().(*Regex).re
+	switch x := repl.(type) {
+	case starlark.Callable:
+		fn := func(matched string) string {
+			res, err := starlark.Call(thread, repl, starlark.Tuple{starlark.String(matched)}, nil)
+			if err != nil {
+				log.Printf("An error occured while executing the function: %s", err.Error())
```
adonovan, line 212: In this case, it's important we return the original err to the caller, so that the stack is preserved and reported to the user. No need for log.Print. Please add a comment and a test for this.

lib/regexp/regexp.go
```diff
@@ -0,0 +1,258 @@
...
+				return matched
+			}
+			resp, ok := res.(starlark.String)
+			if !ok {
+				log.Printf("A string is expected as return type of the function but was %s", res.Type())
```
adonovan, line 217: Return an error, don't print a message. The error message should include the name of the Callable.

(As with the previous suggestion, you will need to save the error away until the ReplaceAllString call returns.)

lib/regexp/regexp.go
```diff
@@ -0,0 +1,258 @@
...
+				return matched
+			}
+			return string(resp)
+		}
+		return starlark.String(re.ReplaceAllStringFunc(src, fn)), nil
+	case starlark.String:
+		return starlark.String(re.ReplaceAllString(src, string(x))), nil
+	}
+	return nil, fmt.Errorf("got %s, want a function or string", repl.Type())
```
adonovan, line 226: s/function or string/string or callable/

lib/regexp/regexp.go
```diff
@@ -0,0 +1,258 @@
...
+
+	if err := starlark.UnpackPositionalArgs(b.Name(), args, kwargs, 1, &src, &max); err != nil {
+		return starlark.None, err
+	}
+
+	re := b.Receiver().(*Regex).re
+	return toTuple(re.Split(src, max)), nil
+}
+
+func builtinAttr(recv starlark.Value, name string, methods map[string]*starlark.Builtin) (starlark.Value, error) {
```
adonovan, line 243: Move these two functions up close to where they are referenced.

I wonder whether these functions (or something similar) should be part of the starlark package API.

adonovan (commented):
Thanks Nicolas (and Brendan), and sorry for the tardy review.

For me, the main question is: should we make our own API around RE2, as you have done here, or try to follow (a subset of) Python3's 're' package? I don't know that package well enough to know the pros and cons, nor what impedance mismatches we might encounter, but @brandjon is a Python expert, so perhaps he can opine.

essobedo:
@adonovan remarks addressed please check again

lib/regexp/regexp.go
```diff
@@ -0,0 +1,290 @@
...
+func compile(_ *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var (
+		pattern string
+	)
+
+	if err := starlark.UnpackPositionalArgs(b.Name(), args, kwargs, 1, &pattern); err != nil {
+		return nil, err
+	}
+
+	if forbiddenPatternRe.MatchString(pattern) {
```
adonovan, line 76: Now that I think clearly, the Go implementation of RE2 already rejects \C for us (because it too is a rune-oriented not byte-oriented machine). So we can delete this check. Sorry for wasting your time. (The regexp was neat though!)

lib/regexp/regexp.go
```diff
@@ -0,0 +1,290 @@
...
+	re *regexp.Regexp
+}
+
+// String implements the Stringer interface.
+func (r *Regexp) String() string { return r.re.String() }
+
+// Type returns a short string describing the value's type.
+func (r *Regexp) Type() string { return "regexp" }
+
+// Freeze renders time immutable. required by starlark.Value interface.
```
adonovan, line 106: "renders r"

"Required".

lib/regexp/regexp.go
```diff
@@ -0,0 +1,290 @@
...
+
+	re := b.Receiver().(*Regexp).re
+	switch x := repl.(type) {
+	case starlark.Callable:
+		var fnErr error
+		fn := func(matched string) string {
+			res, err := starlark.Call(thread, repl, starlark.Tuple{starlark.String(matched)}, nil)
+			if err != nil {
+				// Save the error to be able to return it to the caller
+				fnErr = fmt.Errorf("%s: An error occured while executing the callable: %s", x.Name(), err.Error())
```
adonovan, line 219: fnErr = err // propagate error unmodified, with its call stack

Let the call stack do the talking.

lib/regexp/regexp.go
```diff
@@ -0,0 +1,290 @@
...
+				return ""
+			}
+			resp, ok := res.(starlark.String)
+			if !ok {
+				// Save the error to be able to return it to the caller
+				fnErr = fmt.Errorf("%s: A string is expected as return type of the callable but was %s", x.Name(), res.Type())
```
adonovan, line 225: "%s returned %s, want string"

lib/regexp/regexp.go
```diff
@@ -0,0 +1,290 @@
...
+	repl = strings.ReplaceAll(repl, "$", "$$")
+	startIdx := 0
+	var sb strings.Builder
+	// Replaces the backreferences represented by backslash-escaped digits (\1 to \9) with their
+	// counter parts (${1} to ${9}).
+	for _, match := range backreferenceRe.FindAllStringSubmatchIndex(repl, -1) {
+		startIdxMatch := match[0]
+		endIdxMatch := match[1]
+		if startIdxMatch > 0 && repl[startIdxMatch-1] == '\\' {
+			// The preceding character is a slash so we don't have a real match let's skip it
```
adonovan, line 267: This too is not sound: the digit may be preceded by two backslashes. You can't do these kinds of lexical analysis correctly unless you make a complete pass from the beginning.

lib/regexp/regexp.go
```diff
@@ -0,0 +1,290 @@
...
+	}
+
+	re := b.Receiver().(*Regexp).re
+	return toList(re.Split(src, max)), nil
+}
+
+// Converts the replacement pattern to make it compatible with what the regexp package expects.
+func convertReplacementPattern(repl string) string {
+	// Escapes the specifal dollar characters if any
+	repl = strings.ReplaceAll(repl, "$", "$$")
```
adonovan, line 258: This will spurious replace an escaped dollar symbol.

lib/regexp/regexp.go
```diff
@@ -0,0 +1,290 @@
...
+//                                 The parameter max is optional: by default no limit is applied.
+//
+//     regexp.find_submatches(src) - Returns a new, mutable list of strings holding the text of the leftmost match of the regular expression regexp
+//                                   in the given string and the matches, if any, of its subexpressions. An empty list indicates
+//                                   no match.
+//
+//     regexp.matches(src) - Reports whether the given string contains any match of the regular expression regexp.
+//
+//     regexp.replace_all(src, repl) - Returns a copy of the given string, replacing matches of the regular expression regexp
+//                                     with the replacement string repl. Inside repl, backslash-escaped digits (\1 to \9) can be
```
adonovan, line 40: Which regexp package's design is this feature following? Python?  I can't tell whether convertReplacementFunction is correct without a more detailed description of its intention.

I wonder whether the feature is needed.

lib/regexp/regexp.go
```diff
@@ -0,0 +1,290 @@
...
+//                                     used to insert text matching corresponding parenthesized group from the pattern.
+//                                     \0 in repl refers to the entire matching text.
+//
+//     regexp.replace_all(src, replFunc) - Returns a copy of the given string in which all matches of the regular expression regexp
+//                                         have been replaced by the return value of the replacement function applied to the
+//                                         matched substring.
+//
+//     regexp.split(src, max) - Returns a new, mutable list of strings between all the matches of the regular expression regexp.
```
adonovan, line 48: What do we get if the regexp matches the empty string? Does it explode the string into single-codepoint substrings?

adonovan (commented):
Thanks, looking much better. 

Still waiting for @brandjon's thoughts on the API.

lib/regexp/regexp.go
essobedo: @adonovan I knew that it was not supported but as you wanted a clear error message, I added it but ok let's remove it

lib/regexp/regexp.go
essobedo: This comes from the C++ implementation of re2 https://github.com/google/re2/blob/master/re2/re2.h#L444-L459

essobedo: What `convertReplacementPattern` does is basically converting a replacement pattern composed of backreferences of type `\i` (where i is the index of the group) into a replacement pattern composed of backreferences of type `${i}`.

Backreferences of type backslash-escaped digits are what is we have in python as you can see in [this example ](https://regex101.com/r/jprisL/1)

lib/regexp/regexp.go
essobedo: This is needed to escape existing dollar characters since the dollar character is used to compose a backreference

lib/regexp/regexp.go
essobedo: What you describe: digit preceded by an even amount of backslashes is typically the reason why I added this test. If you check well, your usecase is already covered in the test cases https://github.com/google/starlark-go/pull/369/files#diff-bebff646c9e370f4768b2380210912756b9fb8ff362d777fbfcab94694ada0ceR75-R77

lib/regexp/regexp.go
essobedo: It will split the string character by character like in the test case that I've just added

essobedo:
@adonovan remarks addressed and questions answered, please check again

lib/regexp/regexp.go
essobedo: You have the same logic in the python lib https://docs.python.org/3/library/re.html#re.sub

essobedo:
@adonovan May I have a feedback please? 🙏

adonovan:
Let's get @brandjon to opine on the API before we go into details of the implementation.

essobedo:
Hi @brandjon, any feedbacks to give please 🙏 ?

lib/regexp/regexp.go
```diff
@@ -0,0 +1,283 @@
...
+//                                 no match. If max > 0, at most max strings are returned. If max == 0,
+//                                 an empty list is returned. If max < 0, all strings are returned.
+//                                 The parameter max is optional: by default no limit is applied.
+//
+//     regexp.find_submatches(src) - Returns a new, mutable list of strings holding the text of the leftmost match of the regular expression regexp
+//                                   in the given string and the matches, if any, of its subexpressions. An empty list indicates
+//                                   no match.
+//
+//     regexp.matches(src) - Reports whether the given string contains any match of the regular expression regexp.
+//
```
brandjon, line 38: Have you given any thought to including variants of the find* functions that return indices instead of strings? I guess that kind of optimized API might not be worth its weight in the context of Starlark?

lib/regexp/regexp.go
```diff
@@ -0,0 +1,283 @@
...
+//                        A regexp value can be used for matching using its matches, find, find_all and other methods.
+//
+//     regexp.find(src) - Returns a string holding the text of the leftmost match in the given string of the regular expression regexp.
+//                         The result is "" if there is no match or if the pattern successfully matches an empty string.
+//
+//     regexp.find_all(src, max) - Returns a new, mutable list of all successive matches of the regular expression regexp. An empty list indicates
+//                                 no match. If max > 0, at most max strings are returned. If max == 0,
+//                                 an empty list is returned. If max < 0, all strings are returned.
+//                                 The parameter max is optional: by default no limit is applied.
+//
```
brandjon, line 32: "find_submatches" -- have you considered naming it "find_submatch" to more exactly match Go's API? That said, I think the plural form is slightly more readable.

lib/regexp/regexp.go
brandjon: What are the merits of using `\i` syntax instead of Go's `$i` or `${i}`? I understand staying closer to what a Python programmer may be familiar with, but then we're already using Go as the inspiration for the method names and signatures.

Can `\i` notation support more than 9 subexpressions?

Make sure to clarify how a literal `\` can be represented. Consider including `replace_all_literal()`, though that can already be achieved by using `replace_all(src, lambda x: "literal_result")`.

FWIW, `re.sub` has a (to me) ridiculous behavior of expanding substring values representing backslash escapes into corresponding values. I.e., `re.sub` turns `"\\n"` in the replacement text into `"\n"` when substituted. So I'm not in favor of mirroring Python's re API just for mere parity.

lib/regexp/regexp.go
```diff
@@ -0,0 +1,283 @@
...
+//     regexp.find_submatches(src) - Returns a new, mutable list of strings holding the text of the leftmost match of the regular expression regexp
+//                                   in the given string and the matches, if any, of its subexpressions. An empty list indicates
+//                                   no match.
+//
+//     regexp.matches(src) - Reports whether the given string contains any match of the regular expression regexp.
+//
+//     regexp.replace_all(src, repl) - Returns a copy of the given string, replacing matches of the regular expression regexp
```
brandjon, line 39: Go's docs don't clarify this, but how are overlapping matches handled? Is it all leftmost non-overlapping matches? And if so, leftmost must be defined (just as with all the other functions in this library) so it's clear whether we're talking longest or shortest match.

lib/regexp/regexp.go
```diff
@@ -0,0 +1,283 @@
...
+//                                     with the replacement string repl. Inside repl, backslash-escaped digits (\1 to \9) can be
+//                                     used to insert text matching corresponding parenthesized group from the pattern.
+//                                     \0 in repl refers to the entire matching text.
+//
```
brandjon, line 43: Seems a little strange to me that there's no function to replace only the first occurrence or at most k occurrences, but then Go doesn't have that either.

lib/regexp/regexp.go
```diff
@@ -0,0 +1,283 @@
...
+//     regexp.replace_all(src, replFunc) - Returns a copy of the given string in which all matches of the regular expression regexp
```
brandjon, line 44: You need to also specify the order in which replFunc is called on matches, since it may have observable side effects.

lib/regexp/regexp.go
```diff
@@ -0,0 +1,283 @@
...
+//                                         have been replaced by the return value of the replacement function applied to the
+//                                         matched substring.
+//
+//     regexp.split(src, max) - Returns a new, mutable list of strings between all the matches of the regular expression regexp.
+//                              If max > 0, at most max strings are returned knowing that the last string is
+//                              the unsplit remainder. If max == 0, an empty list is returned. If max < 0,
```
brandjon, line 50: Go's `split()` behavior is different for the `max=0` case: It returns nil. (In Python the corresponding 0 behavior is to apply no limit.)

I don't see a benefit to Go's behavior of returning a sentinel rather than an empty sequence. It breaks the guarantee that for all non-negative k, `max=k` ensures the result is a list of at most size `k`.

lib/regexp/regexp.go
brandjon: This differs from both Go and Python, which will consider `foo*` to match the beginning and end of the string and thus include leading and trailing empty string elements.

brandjon (commented):
Sorry for the delay. My comments are limited to this PR, I haven't formed any detailed reaction to the linked issue.

I've never used/seen the Go regex API before, but I like it, and it seems fine to try to mimic a subset of it here.

lib/regexp/regexp.go
```diff
@@ -0,0 +1,283 @@
...
+// The module defines the following functions:
+//
+//     compile(pattern) - Compiles a pattern in RE2 syntax (https://github.com/google/re2/wiki/Syntax) to a value of type 'regexp'.
+//                        Each call to compile returns a distinct regexp value.
+//                        A regexp value can be used for matching using its matches, find, find_all and other methods.
+//
+//     regexp.find(src) - Returns a string holding the text of the leftmost match in the given string of the regular expression regexp.
+//                         The result is "" if there is no match or if the pattern successfully matches an empty string.
+//
+//     regexp.find_all(src, max) - Returns a new, mutable list of all successive matches of the regular expression regexp. An empty list indicates
```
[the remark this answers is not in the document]
essobedo: > What is the order of the list? Is it unspecified? Is it guaranteed at least to be deterministic?
It is ordered by position in the string.

For example if the regular expression is `a[a|b]*c` and the input string is `xabbabbcaaabbcx`, the result is `["abbabbc", "aaabbc"]`

> What is a "match"? Is it a string?

Yes it is a part of the input string that matches with the regular expression

lib/regexp/regexp.go
```diff
@@ -0,0 +1,283 @@
...
+//                                 no match. If max > 0, at most max strings are returned. If max == 0,
```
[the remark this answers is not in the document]
essobedo: It is something that we have in the Go lib (as a guard?), but let's remove it

lib/regexp/regexp.go
```diff
@@ -0,0 +1,283 @@
...
+//                                 an empty list is returned. If max < 0, all strings are returned.
+//                                 The parameter max is optional: by default no limit is applied.
```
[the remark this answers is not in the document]
essobedo: Yes it is possible, I will do it for the split method since there is the same logic (setting the max split) in python

essobedo:
@brandjon Could you please check again? 🙏

essobedo:
@adonovan @brandjon any remarks about this PR?

johntdyer:
Any update here folks?

luispadron:
Checking back a year later 😅 any update here? Is this PR blocked by something?

essobedo:
@luispadron yes this PR is blocked because unfortunately there is no more reviewers in this project 😞

essobedo:
@b5 I guess you can close it

b5:
Dang. Frustrating. Ok closing now

jawilking:
> @luispadron I guess we need to consider that this project is in maintenance mode and no more features will be accepted

So is this project dead now?

adonovan:
No, Starlark-go is very much alive, but time is precious and a regexp package is nowhere near our highest priorities. I'd still be willing to approve a PR to add the right API, but I don't know what that looks like at the moment and it would be a shame to add something that is too closely coupled to the underlying Go implementation to be implemented in Java or -Rust (as we may have done with the time package).

Until such time as a standard regexp package is added, you can of course fork the code in this PR in your own project.

snehal520:
Is there any plan of adding regexp to starlark anytime soon ?

Issue #370, opened 2021-04-26
Some confusion surrounding mixture of float and int operands


Algebra8:
The [specs regarding floating-point numbers](https://github.com/google/starlark-go/blob/master/doc/spec.md#floating-point-numbers) mentions the following:

> An arithmetic operation applied to a mixture of `float` and `int `operands works as if the `int` operand is first converted to a `float`... 

And gives the example of `x // y`, by stating:

> `x // y` yields `floor(x / y)`, that is, the largest integer value not greater than `x / y`. Although the resulting number is integral, it is represented as a `float` if either operand is a `float`.

However, the examples at the bottom of the section include the following:

```python
3.0 // 2.0                                      # 1
```

Where `1 != 1.0`.

 I am assuming the example is incorrect because tests of [`math.ceil` in `testdata/math.star`](https://github.com/google/starlark-go/blob/master/starlark/testdata/math.star#L17) seem to corroborate what the spec states. E.g.:

```python
assert.eq(math.ceil(0), 0.0)
assert.eq(math.ceil(1), 1.0)
assert.eq(math.ceil(10), 10.0)
```

Besides the possible (minor) issue with the example described above, there is something else I am curious/confused about. As far as I can tell, there does not exist any functionality to allow arithmetic operations between `starlark.Int` and `starlark.Float`.

Is the reason for this the portion of the specs that I have pointed out above? Or can there be some situations that call for built-in `float`, `int` operations.

To give you an example, [here](https://github.com/tdakkota/sun/pull/12) I am working on implementing Python's `itertools.count()`  method for starlark go. One issue is the following snippet:

```bash
>>> itertools.count(0.5, 5)
count(0.5, 5)
```

Please note that in Python's implementation, `5` — which is integral, is represented as an `int` and not a `float`.

However, using the [`floatOrInt`](https://github.com/google/starlark-go/blob/master/lib/math/math.go#L111) methodology that `lib/math.go` uses (which can be seen in [`math.ceil()`](https://github.com/google/starlark-go/blob/master/lib/math/math.go#L129)), will yield the following:

```bash
>>> itertools.count(0.5, 5)
count(0.5, 5.0)
```
where `5` has become `5.0`.

To get around the issue I described above, I used the following workaround:

```go
// float or int type to allow mixed inputs.
type floatOrInt struct {
	value starlark.Value
}

// Unpacker for floatOrInt.
func (p *floatOrInt) Unpack(v starlark.Value) error {
	switch v := v.(type) {
[...]

PR #373, opened 2021-05-02
Fix bug in math.ceil by returning integral instead of float (fixes #372)

### LINKED ISSUE #372: Bug with math.ceil returning float
The current implementation of Starlark's `math.ceil` will always return a `starlark.Float` as a result of being wrapped by [`newUnaryBuiltin`](https://github.com/google/starlark-go/blob/master/lib/math/math.go#L70).

Python's `math.ceil`, however, will always return an integral, as can be seen in the [docs](https://docs.python.org/3/library/math.html#math.ceil).

Also, Python's `math.ceil` will return an error when called with non-finite numeric values, i.e. `nan`, `+inf`, and `-inf`:

```bash
>>> math.ceil(float("nan"))
  File "<stdin>", line 1, in <module>
ValueError: cannot convert float NaN to integer
>>> math.ceil(float("inf"))
  File "<stdin>", line 1, in <module>
OverflowError: cannot convert float infinity to integer
>>> math.ceil(float("-inf"))
  File "<stdin>", line 1, in <module>
OverflowError: cannot convert float infinity to integer
```

But Starlark's implementation allows for them, as can be seen in the [tests](https://github.com/google/starlark-go/blob/master/starlark/testdata/math.star#L20):

```python
...
assert.eq(math.ceil(inf), inf)
assert.eq(math.ceil(nan), nan)
...
assert.eq(math.ceil(-inf), -inf)
...
```

This bug was first mentioned in #370, as can be seen in [this thread](https://github.com/google/starlark-go/issues/370#issuecomment-828793228).

Possible solution:

Extract `math.ceil` from `newUnaryBuiltin` into its own function, `ceil`, that returns the result as an integral and returns an error when said result is not finite.

Algebra8:
Python's math.ceil function always returns a value's ceiling as an
integral, as can be seen in https://docs.python.org/3/library/math.html#math.ceil.

The previous iteration of Starlark's math.ceil, however, would always return a
float, as a result of being wrapped in newUnaryBuiltin. Since ceil is different
from the other wrapped methods, as described above, separate it in its own function.

It is worth noting that Python's math.ceil will also not accept non-finite values,
such as nan, +inf, -inf and will return custom error messages for nan and infinities.
These are also handled in the new ceil function.

Finally, tests are incorporated in testdata/math.star to reflect the changes
described above.

fixes #372


Algebra8:
@adonovan 

Hello!

Here is a PR to handle the bug discussed in #370.

I look forward to any feedback.

Thanks!

Algebra8:
It is also worth noting that there seems to be some issue with comparing types, as can be seen in the tests [here](https://github.com/google/starlark-go/pull/373/files#diff-517c4372e0c729c7af57c3b172c205e5651b4b064fb0604a734dfd595f11a8daR27).

Essentially, in Python:

```bash
>>> type(3) == "int"
False
>>> type(3) == int
True
```

But in Starlark, the opposite seems to be true. That is:

```python
assert.eq(type(3), "int")  # ok
assert.eq(type(3), int)    # not ok
```

This probably has something to do with [`type_` returning a `String`](https://github.com/google/starlark-go/blob/master/starlark/library.go#L1132), but I haven't looked to deep into it. If it is a bug I would be happy to have a go at fixing it.

lib/math/math.go
```diff
@@ -162,6 +162,37 @@ func log(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, kwar
...
 
+// ceil wraps the Ceil function as a Starlark built-in that
+// returns the ceiling as an Integral.
+func ceil(
+	thread *starlark.Thread,
+	_ *starlark.Builtin,
+	args starlark.Tuple,
+	kwargs []starlark.Tuple,
+) (starlark.Value, error) {
+	var x floatOrInt
```
adonovan, line 173: It is not correct to use floatOrInt here. If x is an int, we should return it unchanged, not convert it to a float, which may lose information.

lib/math/math.go
```diff
@@ -162,6 +162,37 @@ func log(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, kwar
 	return starlark.Float(math.Log(float64(x)) / math.Log(float64(base))), nil
 }
 
+// ceil wraps the Ceil function as a Starlark built-in that
+// returns the ceiling as an Integral.
+func ceil(
+	thread *starlark.Thread,
```
adonovan, line 168: Put all this on one line: func ceil(...) ... {

lib/math/math.go
```diff
@@ -162,6 +162,37 @@ func log(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, kwar
...
+
+	if err := starlark.UnpackPositionalArgs(
+		"ceil", args, kwargs, 1, &x,
+	); err != nil {
+		return nil, err
+	}
+
+	result := math.Ceil(float64(x))
+
+	// Python's math.ceil raises an error for inf and nan.
```
adonovan, line 183: You don't need this comment: it's obvious from the need to return an int that these inputs are invalid.

Also, combine the checks and errors, and report the supplied value, not the intermediate `result`:
```
if !isInf... || isNan... {
   return nil, fmt.Errorf("ceil: invalid argument %v", x)
}
```
But if you use NumberToInt, you don't need this check.

lib/math/math.go
```diff
@@ -162,6 +162,37 @@ func log(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, kwar
...
+	// This is because it always returns an integral and cannot convert
+	// any of these to an integral.
+	if math.IsInf(result, 0) {
+		return nil, fmt.Errorf("cannot convert float infinity to integer")
+	}
+	if math.IsNaN(result) {
+		return nil, fmt.Errorf("cannot convert float NaN to integer")
+	}
+
+	return starlark.MakeInt64(int64(result)), nil
```
adonovan, line 193: This conversion is lossy for large values. Use starlark.NumberToInt.

starlark/testdata/math.star
```diff
@@ -17,16 +17,24 @@ assert.eq(math.ceil(10.0), 10.0)
...
+assert.eq(type(math.ceil(0.4)), "int")
+assert.eq(type(math.ceil(10)), "int")
+assert.eq(type(math.ceil(-10.0)), "int")
+assert.eq(type(math.ceil(-0.5)), "int")
+assert.fails(
+  lambda: math.ceil(inf), "cannot convert float infinity to integer")
+assert.fails(
+  lambda: math.ceil(-inf), "cannot convert float infinity to integer")
+assert.fails(
+  lambda: math.ceil(nan), "cannot convert float NaN to integer")
```
adonovan, line 37: Let's add a test for a float value larger than int64.

lib/math/math.go
```diff
@@ -67,7 +67,7 @@ import (
 var Module = &starlarkstruct.Module{
 	Name: "math",
 	Members: starlark.StringDict{
-		"ceil":      newUnaryBuiltin("ceil", math.Ceil),
+		"ceil":      starlark.NewBuiltin("ceil", ceil),
 		"copysign":  newBinaryBuiltin("copysign", math.Copysign),
 		"fabs":      newUnaryBuiltin("fabs", math.Abs),
 		"floor":     newUnaryBuiltin("floor", math.Floor),
```
adonovan, line 73: The floor function needs the same treatment.

adonovan:
> It is also worth noting that there seems to be some issue with comparing types

This is required by the spec, it's not a bug. I wouldn't have designed it this way, but it is what it is.

lib/math/math.go
Algebra8: Thanks for pointing that out. `NumberToInt` is a great abstraction and carries the same error messages.

lib/math/math.go
Algebra8: Great point, thanks for pointing that out.

starlark/testdata/math.star
Algebra8: So, if the float value is significantly larger than `math.MaxInt64`, then the function works as expected, e.g.

```python
f = 1.8446744073709552e+19
assert.eq(math.ceil(f), f)  # ok
```

However, if it is slightly larger, then we'll get an overflow:

```python
q = 9223372036854775807.5
print(math.ceil(q))  # -9223372036854775808
```

This seems to be because `q` is close enough to `math.MaxInt64` that the [check in `finiteFloatToInt` will register it as an int](https://github.com/google/starlark-go/blob/master/starlark/int.go#L427).

Is this ok? Or should the `ceil` function take this into consideration?

adonovan:
(I really don't understand the theory of GitHub's PR comments UI. It won't let me respond to your comment directly.)

> This seems to be because q is close enough to math.MaxInt64 that the check in finiteFloatToInt will register it as an int.

You've discovered a beautifully subtle bug which I have filed as https://github.com/google/starlark-go/issues/375 since it affects the interpreter core. I'll send a fix for that presently.

starlark/testdata/math.star
adonovan: See https://github.com/google/starlark-go/pull/376/files

starlark/testdata/math.star
```diff
@@ -17,16 +17,25 @@ assert.eq(math.ceil(10.0), 10.0)
...
 assert.eq(math.ceil(-10.0), -10.0)
 assert.eq(math.ceil(-1), -1.0)
 assert.eq(math.ceil(-10), -10.0)
-assert.eq(math.ceil(-inf), -inf)
+assert.eq(type(math.ceil(0)), "int")
+assert.eq(type(math.ceil(0.4)), "int")
+assert.eq(type(math.ceil(10)), "int")
+assert.eq(type(math.ceil(-10.0)), "int")
+assert.eq(type(math.ceil(-0.5)), "int")
+assert.eq(math.ceil(1.8446744073709552e+19), 1.8446744073709552e+19)
```
adonovan, line 32: It's unclear what the first constant signifies. `float(1<<64)`? The same thing plus or minus one? Derive the argument to make it clear. 

Is the second argument intentionally a float? It may compare equal to the int returned by ceil, but it would be better to use an expression that is actually an int.

adonovan (commented):
Production code looks good. One last question about the tests! Thanks for your patience.

Algebra8:
@adonovan 

> Production code looks good. One last question about the tests! Thanks for your patience.

No, thank you for your patience and feedback! I really appreciate the time you've taken for it 🙏 

> It's unclear what the first constant signifies. float(1<<64)? The same thing plus or minus one? Derive the argument to make it clear.

Honestly, I think I just multiplied `math.MaxInt` by `2` and added `0.5`, so thank you for bringing that up, it should be derived.

> Is the second argument intentionally a float? It may compare equal to the int returned by ceil, but it would be better to use an expression that is actually an int.

Great point, and will do :)

adonovan (approved):
Looks good! Thanks again.

PR #379, opened 2021-07-01
Support maps in lib/proto

vasyharan:
Adds support for maps in lib/proto.


vasyharan:
@googlebot I fixed it.

PR #383, opened 2021-08-30
starlark: disable int32 optimization on openbsd

### LINKED ISSUE #382: Requests a 'large" amount of virtual memory addresses, which may fail on some operating systems.
I am trying to use a piece of software that relies on  starlark-go on a system that rigourously enforces ulimits (in particular on virtual memory). Which means that I have to patch out the the line 62 in int_posix64.go, which tries to allocate 4Gb of memory, which the machine does not possess.

```
var smallints = reserveAddresses(1 << 32)

func reserveAddresses(len int) uintptr {
	b, err := unix.Mmap(-1, 0, len, unix.PROT_READ, unix.MAP_PRIVATE|unix.MAP_ANON)
	if err != nil {
		log.Fatalf("mmap: %v", err)
	}
	return uintptr(unsafe.Pointer(&b[0]))
}
```

Please, consider implementing proper memory management, instead of "just allocating 4Gb of RAM".

adonovan:
This PR disables the int32 optimization on openbsd because its default ulimit is only about 1GB.

Fixes #382

[Note: comments by reviewers refer to PR's description of original approach, which was abandoned, as explained in the code comment.]

### DIFF
--- a/starlark/int_generic.go
+++ b/starlark/int_generic.go
@@ -1,4 +1,4 @@
-//+build !linux,!darwin,!dragonfly,!freebsd,!netbsd,!openbsd,!solaris darwin,arm64 !amd64,!arm64,!mips64x,!ppc64x
+//+build !linux,!darwin,!dragonfly,!freebsd,!netbsd,!solaris darwin,arm64 !amd64,!arm64,!mips64x,!ppc64x
 
 package starlark
 
--- a/starlark/int_posix64.go
+++ b/starlark/int_posix64.go
@@ -1,4 +1,4 @@
-//+build linux darwin dragonfly freebsd netbsd openbsd solaris
+//+build linux darwin dragonfly freebsd netbsd solaris
 //+build amd64 arm64,!darwin mips64x ppc64x
 
 package starlark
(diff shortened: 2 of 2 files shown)


jayconrod:
I like this idea; V8 did something very similar. The test panic looks hard to overcome though.

(Incidentally, I've been told we'll need to switch away from Travis CI soon. I'll start looking at that soon, but let me know if you have opinions on how that should work).

falsifian:
Would making `intImpl` a `uintptr` instead of an `unsafe.Pointer` make the test happy?

adonovan:
> The test panic looks hard to overcome though.

Yeah; I wondered whether the runtime would care about or get confused by manufactured pointers. (Historically its conservative GC allowed such things, but it has been getting stricter every year.) This crash clearly answers the question in the negative. This particular error means that a pointer contained a value that points into the zero page of memory, which is reserved as a nomansland: https://github.com/golang/go/blob/master/src/runtime/stack.go#L614

But more generally, it appears that user programs are not supposed to put bad pointers in the heap. The comments here make clear that pointers into mmap'd regions are ok, but pointers into non-stack spans are forbidden:
https://github.com/golang/go/blob/master/src/runtime/mbitmap.go#L387-L411


> Would making intImpl a uintptr instead of an unsafe.Pointer make the test happy?

No. The union must appear to the GC to be a pointer so that it keeps the big.Int variables live. In the old scheme, all its values were valid pointers, and this change broke that.

Reverting to the conservative strategy of disabling the optimization on OpenBSD.

adonovan:
I updated the PR description to reflect the new implementation, which of course renders all the subsequent comments meaningless. I won't be doing that again.

jayconrod (approved):

adonovan:
> (Incidentally, I've been told we'll need to switch away from Travis CI soon. I'll start looking at that soon, but let me know if you have opinions on how that should work).

Sorry, I haven't even thought about it, though I have been vaguely aware of it looming. (Presumably it is this: https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing.) Is Google paying the TravisCI bill one of the options? :)

jayconrod:
Paying the bill would probably be fine, but there's more I can't say publicly.

PR #384, opened 2021-08-31
Add GitHub action for testing

jayconrod:
It runs all tests and confirms go.mod and go.sum are tidy, just like
the Travis CI test.

The action runs with Go 1.16 and 1.17 on darwin, linux, and windows.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,19 @@
+name: Starlark Go Tests
+on: [push, pull_request]
+jobs:
+  build:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        go-version: [1.16.x, 1.17.x]
+    runs-on: ${{ matrix.os }}
+    steps:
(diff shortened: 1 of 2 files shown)


jayconrod:
I'm not sure if there's a way to verify this works before merging it. I'm cargo-culting from https://github.com/google/wire, which did this migration yesterday.

adonovan:
> I'm not sure if there's a way to verify this works before merging it. I'm cargo-culting from https://github.com/google/wire, which did this migration yesterday.

I think the CI system runs the new configuration in PRs that edit the configuration. This is good for developers wanting to iterate, but I do wonder what are the security implications of allowing anyone who can send a PR to run arbitrary code. Actually, I know the answer: bitcoin mining, and the end of free Travis service.

Seems like it passed---though it took me a moment to realize that all the "The command foo exited with 0" were telling me "nothing to see here".

Thanks.

adonovan (approved):

adonovan:
Hmm, I thought the UI was telling me all was ok when I committed it.

PR #385, opened 2021-08-31
avoid cross-package references in tests

adonovan:
The previous implementation of DataFile (which abstracts different build systems' ways of locating test data files) constructed file names of data files using $GOPATH; also, it used a cross-package reference to locate the assert.star file. This doesn't work under Go modules. Now, DataFile looks only in the current directory, and the assert.star file is moved into a constant section of the executable.

Also, fix tests historically broken on Mac/Windows due to diff and CR LF differences.

### DIFF
--- a/internal/chunkedfile/chunkedfile.go
+++ b/internal/chunkedfile/chunkedfile.go
@@ -28,6 +28,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"regexp"
+	"runtime"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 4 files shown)


jayconrod (approved):
+2, looks like the tests are happy. And thanks for fixing this. Sorry for the unexpected work.

I always forget that Git rewrites line endings on checkout in Windows. Making the test code tolerate that seems like the best way to fix it.

With modules, cross-package references are okay, as long as they're in the same module. Since each test runs in its package directory, relative paths within the module should work. I should have caught this with local testing, but I'm still in the habit of checking things out where they would have been in GOPATH, so it worked on my machine.

adonovan:
> +2, looks like the tests are happy. And thanks for fixing this. Sorry for the unexpected work.

No, thank you for making our automated tests run again, and on Darwin + Windows for (apparently) the first time.

> I always forget that Git rewrites line endings on checkout in Windows. Making the test code tolerate that seems like the best way to fix it.

Yeah, I forgot that too. Presumably at this point in history every program in Darwin and Windows can handle UNIX line endings, which does make me wonder why programs continue to emit them.

> With modules, cross-package references are okay, as long as they're in the same module.

Ah, good point, that would have worked. Still, I'm happy with this approach.

PR #386, opened 2021-08-31
Remove Travis CI configuration and replace README badge

jayconrod:
Also update links from godoc.org to pkg.go.dev since we're here.


jayconrod:
I think the travis-ci presubmit hook needs to be removed before I can submit this, but I don't have the level of access needed to do that. I expect the org admins will do that within a day or so, but if not, I'll follow up on this side.

adonovan (approved):
Nice!

PR #387, opened 2021-10-10
feat: update deps

ecasilla:


go.mod
```diff
@@ -1,13 +1,13 @@
 module go.starlark.net
 
-go 1.13
+go 1.17
```
adonovan, line 3: This diff doesn't seem necessary: this module doesn't require a recent version of Go.

PR #390, opened 2021-12-03
Fix the link to the Java implementation

tetsuok:
This fixes the link to the Java implementation of Starlark in spec.md.


adonovan (approved):
Thanks again!

PR #393, opened 2022-02-11
Suggest keyword argument based on spelling in UnpackArgs

fenollp:
Users of tools such as Bazel and ours arguably edit mostly
the kwargs of functions the tool provides. Typos often slip
through and spelling suggestion can save some headache.

This change may affect the message string of some errors.

Signed-off-by: Pierre Fenoll <pierrefenoll@gmail.com>

### DIFF
--- a/starlark/testdata/function.star
+++ b/starlark/testdata/function.star
@@ -321,3 +321,9 @@ b(1, y=2, )
 #c(1, *[], )
 #d(1, *[], z=None, )
 #e(1, *[], z=None, *{}, )
+
+---
+# Unpack provides spell check for argument names.
+load("assert.star", "assert")
+
+assert.fails(lambda: min([], keg=1), ".+did you mean key\\?")
--- a/starlark/unpack.go
+++ b/starlark/unpack.go
@@ -8,6 +8,8 @@ import (
 	"log"
 	"reflect"
 	"strings"
+
+	"go.starlark.net/internal/spell"
 )
 
 // An Unpacker defines custom argument unpacking behavior.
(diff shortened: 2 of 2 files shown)


starlark/eval_test.go
```diff
@@ -764,6 +764,16 @@ func TestUnpackRequiredAfterOptional(t *testing.T) {
 	}
 }
 
+func TestUnpackSuggestsKwargNameBasedOnSpelling(t *testing.T) {
```
adonovan, line 767: This test could be expressed in two lines in testdata/function.star:

```
def misspelled(response): None
misspelled(reponse=1) ### ...did you mean...
```

adonovan (commented):
Very nice!

starlark/eval_test.go
fenollp: I am missing something here. How can this test be able to suggest the correct spelling?

My attempt fails:
```python
---
# Unpacking kwargs should suggest keyword name based on spelling
load("assert.star", "assert")

def f():
    def misspelled(response): None
    misspelled(reponse=1)

assert.fails(f, ".+did you mean response?.+")

# --- FAIL: TestExecFile (0.12s)
#     starlarktest.go:117: Traceback (most recent call last):
#           testdata/function.star:333:13: in <toplevel>
#           builtins/assert.star:39:14: in _fails
#         Error: regular expression (.+did you mean response?.+) did not match error (function misspelled got an unexpected keyword argument "reponse")
```

adonovan: Ah, of course: it has to be a a built-in function. Try:

```
---
# Unpack provides spell check for argument names.
min([], keg=1) ### `did you mean key\?`
```

fenollp:
Thanks for the review and I'm glad you like this!

starlark/testdata/function.star
```diff
@@ -321,3 +321,9 @@ b(1, y=2, )
 #c(1, *[], )
 #d(1, *[], z=None, )
 #e(1, *[], z=None, *{}, )
+
+---
+# Unpack provides spell check for argument names.
+load("assert.star", "assert")
+
+assert.fails(lambda: min([], keg=1), ".+did you mean key\\?")
```
adonovan, line 329: FYI, the triple-# syntax below means "implicitly assert that execution fails with an error that matches the regexp", so you don't need the assert package. But what you have here is fine too.

```
---
# Unpack provides spell check for argument names.
min([], keg=1) ### `did you mean key\?`
```

PR #395, opened 2022-02-15
Add strings.remove{pre,suf}fix methods

adonovan:
See https://github.com/bazelbuild/starlark/pull/212 for spec change

See https://github.com/bazelbuild/bazel/pull/14824 for Java PR
See https://github.com/facebookexperimental/starlark-rust/commit/7a7ef4b61afd375e832ee0f5cac1ddd38cc349d5 for Rust commit

Updates https://github.com/bazelbuild/starlark/issues/185

### DIFF
--- a/starlark/library.go
+++ b/starlark/library.go
@@ -123,6 +123,8 @@ var (
 		"lower":          NewBuiltin("lower", string_lower),
 		"lstrip":         NewBuiltin("lstrip", string_strip), // sic
 		"partition":      NewBuiltin("partition", string_partition),
+		"removeprefix":   NewBuiltin("removeprefix", string_removefix),
+		"removesuffix":   NewBuiltin("removesuffix", string_removefix),
 		"replace":        NewBuiltin("replace", string_replace),
 		"rfind":          NewBuiltin("rfind", string_rfind),
 		"rindex":         NewBuiltin("rindex", string_rindex),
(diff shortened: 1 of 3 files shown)


starlark/library.go
```diff
@@ -1889,6 +1891,22 @@ func string_partition(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value,
...
 
+// https://github.com/google/starlark-go/blob/master/doc/spec.md#string·removeprefix
+// https://github.com/google/starlark-go/blob/master/doc/spec.md#string·removesuffix
+func string_removefix(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error) {
+	recv := string(b.Receiver().(String))
+	var fix string
+	if err := UnpackPositionalArgs(b.Name(), args, kwargs, 1, &fix); err != nil {
+		return nil, err
+	}
+	if b.name[6] == 'p' {
```
laurentlb, line 1902: As the function is short, it might be worth duplicating it instead of using `b.name` magic.

If you want to keep it like this, add a comment on the line

laurentlb (approved):
lgtm

starlark/library.go
adonovan: Since it's easy to avoid the duplication of source and object code (240 insns -> 160), and the cost of the extra load and two cjmps is trivial overall, I'd prefer to keep it this way.

But you're right it could be clearer. Commented in the form of self-documenting code:
```
	if b.name[len("remove")] == 'p' {
```

laurentlb:
I don't think I've ever had write access in this repo.

tetromino (approved):

PR #396, opened 2022-02-16
Add build tag to disable small int optimization on 64bit posix systems

### LINKED ISSUE #394: The int optimization poses problems for users running with ulimit
Reference: https://github.com/go-delve/delve/issues/2907

The code in int_posix64.go assumes that the program has the ability to make a 4G allocation, this is not necessarily the case if the system is using ulimit to limit the amount of memory that processes can allocate. Given that this is an optimization the interpreter should be able to continue functioning even if the allocation fails (possibly attempting a smaller allocation first?)

aarzilli:
Adds a new build tag, starlark_int_generic, to disable the small
integer optimization on posix systems running on 64bit architecture.

This allows starlark to be used on systems with ulimits set.

Also updates the mmap error to be more helpful.

Fixes #394


adonovan:
Thanks @aarzilli. I tried out the alternative approach in https://github.com/google/starlark-go/pull/398 and it doesn't seem to cost much. I think I prefer it because it encapsulates all the complexity in the interpreter.

PR #398, opened 2022-02-16
starlark: fall back to slow Ints if mmap fails

### LINKED ISSUE #394: The int optimization poses problems for users running with ulimit

adonovan:
The Int optimization uses a 4GB virtual address allocation to represent
32-bit values as addresses to avoid allocation. However, some environments
have limited address space.

This change permits the mmap to fail, and in that case it prints a warning
and falls back to always allocating a big.Int, even for small numbers.
Each access to an Int must check the smallints global to see whether the
optimization is active.  The extra dynamic check doesn't cost much:

```
$ go test -run=nope -bench=BenchmarkStarlark -count=3 ./starlark    # > before, after
$ go install golang.org/x/perf/cmd/benchstat@latest
$ ~/go/bin/benchstat  before after
name                                  old time/op  new time/op   delta
Starlark/bench_bigint-16               206µs ± 0%    212µs ± 0%   ~     (p=0.100 n=3+3)
Starlark/bench_builtin_method-16       283µs ± 1%    294µs ± 9%   ~     (p=1.000 n=3+3)
Starlark/bench_calling-16              280µs ± 1%    290µs ± 1%   ~     (p=0.100 n=3+3)
Starlark/bench_gauss-16               9.30ms ± 5%   9.55ms ± 2%   ~     (p=0.700 n=3+3)
Starlark/bench_int-16                 53.7µs ± 1%   59.6µs ± 1%   ~     (p=0.100 n=3+3)
Starlark/bench_mix-16                 99.3µs ± 1%  106.6µs ±10%   ~     (p=0.100 n=3+3)
Starlark/bench_range_construction-16   238ns ± 1%    244ns ± 2%   ~     (p=0.100 n=3+3)
Starlark/bench_range_iteration-16     5.44µs ± 2%   5.67µs ± 1%   ~     (p=0.100 n=3+3)
```

Also, a linux-only test.

Also, simplify the build tags now that the 64-bit POSIX exceptions (iOS, openbsd)
are handled dynamically. As a side effect, M1-based Macs should get the optimization
for the first time. (Requires updating sys module.)

Fixes #394

### DIFF
--- a/starlark/int_test.go
+++ b/starlark/int_test.go
@@ -5,9 +5,15 @@
 package starlark
 
 import (
+	"flag"
 	"fmt"
+	"log"
 	"math"
 	"math/big"
+	"os"
+	"os/exec"
+	"runtime"
+	"strings"
 	"testing"
 )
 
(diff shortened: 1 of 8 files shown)


adonovan:
Ping @brandjon...

aarzilli:
Ping?

tetromino (approved):
Seems reasonable. Would be good to have a test, though (e.g. set RLIMIT_AS, see if int tests still pass).

adonovan:
> Seems reasonable. Would be good to have a test, though (e.g. set RLIMIT_AS, see if int tests still pass).

Done, though it's a pain because it requires fork+execing a child process, and it requires ulimit -v which is not available on MacOS or Windows.

adonovan:
@tetromino, PTAL at the changes to the build tags, go.mod/sum files, Actions workflow config, and of course the test.

starlark/int_test.go
```diff
@@ -86,17 +92,49 @@ func TestImmutabilityMakeBigInt(t *testing.T) {
 	}
 }
 
-func TestImmutabilityBigInt(t *testing.T) {
```
tetromino, line 89: Was `TestImmutabilityBigInt` deleted by mistake?

tetromino (changes requested):

starlark/int_test.go
adonovan: Yes! Good catch. Restored now.

tetromino (approved):
Looks good!

PR #400, opened 2022-02-28
struct: Add SetField method for mutable fields

emcfarlane:
Modifies the struct definition to implement `starlark.SetField`. It changes the storage of fields from an `[]entries` to a `starlark.StringDict`, I've added a benchmark on `Attr`. On setting non-existing fields the new method errors with NoSuchAttrError.

This makes the struct mutable, fields can be edited after creation but no new fields can be added. A frozen struct behaves similar to before. Couldn't see if this violates the starlark spec? If so please feel free to close.

### DIFF
--- a/starlarkstruct/struct.go
+++ b/starlarkstruct/struct.go
@@ -17,14 +17,9 @@ package starlarkstruct // import "go.starlark.net/starlarkstruct"
 //    marginal: finding the index of a field by binary searching on the
 //    sorted list of field names is quite fast compared to the other
 //    overheads.
-// 3) the gains in compactness and spatial locality are also rather
-//    marginal: the array behind the []entry slice is (due to field name
-//    strings) only a factor of 2 larger than the corresponding Go struct
-//    would be, and, like the Go struct, requires only a single allocation.
 
 import (
 	"fmt"
-	"sort"
 	"strings"
 
 	"go.starlark.net/starlark"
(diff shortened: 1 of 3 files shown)


starlarkstruct/struct.go
```diff
@@ -17,22 +17,17 @@ package starlarkstruct // import "go.starlark.net/starlarkstruct"
 //    marginal: finding the index of a field by binary searching on the
 //    sorted list of field names is quite fast compared to the other
 //    overheads.
-// 3) the gains in compactness and spatial locality are also rather
```
emcfarlane, line 20: I'm not sure what impact this has.

adonovan:
A mutable struct is a reasonable idea, and perhaps there should be one in the Starlark library---though there are a surprising number of design questions to consider: flat or hash-based representation? Restrictions on allowed field names? Restrictions on allowed values or types of fields? And so on.

However,  this struct has been defined as an immutable data type, and we can't change that now.

emcfarlane:
I separated out some of the changes for field storage here: https://github.com/google/starlark-go/pull/401

It does feel like struct should have reassignable fields on creation and only when frozen become immutable. Understand this is a behavior change though, any way to support both?

adonovan:
> I separated out some of the changes for field storage here: #401

Thanks; that change looks basically sound.

> It does feel like struct should have reassignable fields on creation and only when frozen become immutable. Understand this is a behavior change though, any way to support both?

You can always add a second data type to your application that supports both. Bazel relies on the existing struct contract and doesn't have a mutable struct (or namedtuple, as Python calls it), but there's no reason you can't create one.

PR #401, opened 2022-03-01
struct: store members as starlark.StringDict for faster access

emcfarlane:
Wraps starlark.StringDict for storing fields.
Improves the performance on the `Attr` method for accessing fields.

I added a benchmark to check the happy path for access to a fields value.

OLD:
BenchmarkAttr_4-4       44931932                24.03 ns/op
BenchmarkAttr_8-4       39342510                28.64 ns/op
BenchmarkAttr_16-4      33596271                32.18 ns/op
BenchmarkAttr_32-4      30505212                37.62 ns/op
BenchmarkAttr_64-4      24424558                43.33 ns/op
BenchmarkAttr_128-4     17947136                64.29 ns/op

NEW:
BenchmarkAttr_4-4       48772784                21.45 ns/op
BenchmarkAttr_8-4       34554286                30.95 ns/op
BenchmarkAttr_16-4      56185027                19.59 ns/op
BenchmarkAttr_32-4      58317289                19.60 ns/op
BenchmarkAttr_64-4      58073444                20.30 ns/op
BenchmarkAttr_128-4     52069291                21.30 ns/op

### DIFF
--- a/starlarkstruct/struct.go
+++ b/starlarkstruct/struct.go
@@ -14,17 +14,15 @@ package starlarkstruct // import "go.starlark.net/starlarkstruct"
 //    them using more specific types such as String, Int, *Depset, and
 //    *File, as such types give no way to represent missing fields.
 // 2) the efficiency gain of direct struct field access is rather
-//    marginal: finding the index of a field by binary searching on the
-//    sorted list of field names is quite fast compared to the other
-//    overheads.
+//    marginal: finding the index of a field by map access is O(1)
+//    and is quite fast compared to the other overheads.
 // 3) the gains in compactness and spatial locality are also rather
-//    marginal: the array behind the []entry slice is (due to field name
+//    marginal: the map behind the field members is (due to field name
 //    strings) only a factor of 2 larger than the corresponding Go struct
 //    would be, and, like the Go struct, requires only a single allocation.
 
 import (
 	"fmt"
-	"sort"
 	"strings"
 
 	"go.starlark.net/starlark"
@@ -53,17 +51,16 @@ func FromKeywords(constructor starlark.Value, kwargs []starlark.Tuple) *Struct {
 	if constructor == nil {
 		panic("nil constructor")
 	}
-	s := &Struct{
-		constructor: constructor,
-		entries:     make(entries, 0, len(kwargs)),
-	}
+	members := make(starlark.StringDict, len(kwargs))
 	for _, kwarg := range kwargs {
 		k := string(kwarg[0].(starlark.String))
 		v := kwarg[1]
-		s.entries = append(s.entries, entry{k, v})
+		members[k] = v
 	}
-	sort.Sort(s.entries)
-	return s
+	return &Struct{
+		constructor: constructor,
+		members:     members,
+	}
 }
 
 // FromStringDict returns a new struct instance whose elements are those of d.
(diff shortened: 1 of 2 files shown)


starlarkstruct/struct.go
```diff
@@ -268,10 +237,13 @@ func structsEqual(x, y *Struct, depth int) (bool, error) {
 		return false, nil
 	}
 
-	for i, n := 0, x.len(); i < n; i++ {
-		if x.entries[i].name != y.entries[i].name {
+	for k, xv := range x.members {
```
adonovan, line 240: This loop is now nondeterministic because EqualDepth may fail before or after an unequal field comparison, depending on iteration order. I don't see any way to make it deterministic other than by iterating over Keys(), which could increase the cost of comparison quite a lot, since it allocates.

Perhaps you could add benchmarks of these operations to testdata/benchmark.star?

starlarkstruct/struct.go
```diff
@@ -14,17 +14,15 @@ package starlarkstruct // import "go.starlark.net/starlarkstruct"
...
 //    *File, as such types give no way to represent missing fields.
 // 2) the efficiency gain of direct struct field access is rather
-//    marginal: finding the index of a field by binary searching on the
-//    sorted list of field names is quite fast compared to the other
-//    overheads.
+//    marginal: finding the index of a field by map access is O(1)
+//    and is quite fast compared to the other overheads.
 // 3) the gains in compactness and spatial locality are also rather
-//    marginal: the array behind the []entry slice is (due to field name
+//    marginal: the map behind the field members is (due to field name
```
adonovan, line 20: Let's replace this paragraph with:

// Such an implementation is likely to be more compact than
// the current map-based representation, though.

starlarkstruct/struct.go
```diff
@@ -268,14 +230,31 @@ func structsEqual(x, y *Struct, depth int) (bool, error) {
 		return false, nil
 	}
 
-	for i, n := 0, x.len(); i < n; i++ {
-		if x.entries[i].name != y.entries[i].name {
-			return false, nil
-		} else if eq, err := starlark.EqualDepth(x.entries[i].value, y.entries[i].value, depth-1); err != nil {
-			return false, err
+	// Loop over every key returning the lowest error by key, if any.
```
adonovan, line 233: This loop must now visit every pair of elements even when the first pair differs. That's a regression. I'm not sure StringDict really makes sense as a representation for struct. The requirements are that:

1. struct is compact.
2. struct.f is fast.
3. str(struct) uses field-name order.
4. hash(struct) uses field-name order (or is order-independent).
5. struct==struct compares fields in name order.

The previous representation using a name-ordered k/v array was better on all requirements except 2, and even then the advantage is only material when N=16, which is an unusually large struct in a typical Starlark program.

I suspect the access patterns of Starlark fields in Attr are highly temporally localized: the most likely field name to be accessed is the previous one. If that is true (and it would be worth measuring), it might be possible to outperform both the existing implementation and the hash table by caching the resulting index of the binary search in a field in the struct (without a mutex!) and using it as a hint next time.

```
func FromStringDict()  {
   ...
   s := &Struct{
	constructor: constructor,
	hint: maxuint32,
   }
}

func (s *Struct) Attr(name string) (starlark.Value, error) {
        if hint := s.hint < n && s.entries[hint].name == name {
              return ... // yay the hint worked
        }
	// Binary search the entries.
	for i < j { ...binary search...}
	if i < n && s.entries[i].name == name {
                s.hint = i // save the hint
		return s.entries[i].value, nil
	}
        ...
}
```

What do you think?

emcfarlane: I agree. I think the original binary search implementation is better. Hint idea sounds good, I'll have a look at how to validate it. Please let me know if you'd like to keep the benchmarking code and I could modify this PR to just include that? Otherwise please close! Thanks for taking the time to review.

I did try a solution of based on the `starlark.hashtable` here: https://github.com/emcfarlane/starlark-go/pull/1 . It does look like it could beat both implementations but requires modify the original hashtable, and cloning it out of the package. Just an experiment.

emcfarlane:
Closing for https://github.com/google/starlark-go/pull/403

PR #402, opened 2022-03-02
starlark: remove allocs on Dict ranges for dictsEqual and write

emcfarlane:
Drops Items() range in favour of looking through the hashtable list
directly. This avoids an allocation, added a benchmark to verify.

Old:
Benchmark/bench_dict_equal-4               18782             61652 ns/op           57344 B/op          2 allocs/op

New:
Benchmark/bench_dict_equal-4               29802             39585 ns/op               0 B/op          0 allocs/op

### DIFF
--- a/starlark/testdata/benchmark.star
+++ b/starlark/testdata/benchmark.star
@@ -60,3 +60,11 @@ def bench_mix(b):
             if i:
                 x += 1
             a = [x for x in range(i)]
+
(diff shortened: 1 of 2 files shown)


adonovan:
Thanks. I think it's probably safe to assume that no element's Equal method has a side effect of modifying the Dict during the iteration.

adonovan (approved):

PR #403, opened 2022-03-04
starlark: add OrderedStringDict for fast iteratable string hash maps

emcfarlane:
Adds a new type OrderedStringDict like StringDict but based on the internal hashtable
implementation with restrictions optimsed for fast iterations. Goals are:

1. compact.
2. fast access by string lookup.
3. ordered fast access by iteration.
4. (opt): fast access by index.

Compared to StringDict, lookups by key are slower but with much faster iteration.
Benchmark results (map is of type StringDict, order is the new OrderedStringDict type):

```
BenchmarkOrderedStringDict_4/map-4      75890469                15.38 ns/op
BenchmarkOrderedStringDict_4/order-4            66676100                17.85 ns/op
BenchmarkOrderedStringDict_4/mapRange-4         18475881                64.27 ns/op
BenchmarkOrderedStringDict_4/orderIter-4        292341111                4.107 ns/op
BenchmarkOrderedStringDict_8/map-4              61201066                19.49 ns/op
BenchmarkOrderedStringDict_8/order-4            63436520                18.40 ns/op
BenchmarkOrderedStringDict_8/mapRange-4         16054929                73.87 ns/op
BenchmarkOrderedStringDict_8/orderIter-4        149468599                7.597 ns/op
BenchmarkOrderedStringDict_16/map-4             57940926                19.34 ns/op
BenchmarkOrderedStringDict_16/order-4           62314574                18.68 ns/op
BenchmarkOrderedStringDict_16/mapRange-4         6022573               195.2 ns/op
BenchmarkOrderedStringDict_16/orderIter-4       74512936                15.62 ns/op
BenchmarkOrderedStringDict_32/map-4             58462815                19.18 ns/op
BenchmarkOrderedStringDict_32/order-4           60321585                18.68 ns/op
BenchmarkOrderedStringDict_32/mapRange-4         3379855               351.2 ns/op
BenchmarkOrderedStringDict_32/orderIter-4       36388614                28.92 ns/op
BenchmarkOrderedStringDict_64/map-4             59516572                19.35 ns/op
BenchmarkOrderedStringDict_64/order-4           59605867                19.14 ns/op
BenchmarkOrderedStringDict_64/mapRange-4         1636936               731.6 ns/op
BenchmarkOrderedStringDict_64/orderIter-4       18821319                60.84 ns/op
BenchmarkOrderedStringDict_128/map-4            60399526                19.63 ns/op
BenchmarkOrderedStringDict_128/order-4          56299444                21.19 ns/op
BenchmarkOrderedStringDict_128/mapRange-4         743340              1507 ns/op
BenchmarkOrderedStringDict_128/orderIter-4       9566968               121.3 ns/op
```

### DIFF
--- a/starlark/hashtable.go
+++ b/starlark/hashtable.go
@@ -55,16 +55,9 @@ func (ht *hashtable) init(size int) {
 func (ht *hashtable) freeze() {
 	if !ht.frozen {
 		ht.frozen = true
-		for i := range ht.table {
-			for p := &ht.table[i]; p != nil; p = p.next {
-				for i := range p.entries {
-					e := &p.entries[i]
-					if e.hash != 0 {
-						e.key.Freeze()
-						e.value.Freeze()
-					}
-				}
-			}
+		for e := ht.head; e != nil; e = e.next {
+			e.key.Freeze()
+			e.value.Freeze()
 		}
 	}
 }
@@ -126,7 +119,7 @@ retry:
 	}
 
 	if insert == nil {
-		// No space in existing buckets.  Add a new one to the bucket list.
+		// No space in existing buckets. Add a new one to the bucket list.
 		b := new(bucket)
 		p.next = b
 		insert = &b.entries[0]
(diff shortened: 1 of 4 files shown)


starlark/hashtable.go
```diff
@@ -55,16 +55,9 @@ func (ht *hashtable) init(size int) {
...
-			for p := &ht.table[i]; p != nil; p = p.next {
-				for i := range p.entries {
-					e := &p.entries[i]
-					if e.hash != 0 {
-						e.key.Freeze()
-						e.value.Freeze()
-					}
-				}
-			}
+		for e := ht.head; e != nil; e = e.next {
```
adonovan, line 58: This is a nice simplification.

starlark/hashtable.go
```diff
@@ -371,3 +364,186 @@ func softHashString(s string) uint32 {
 	}
 	return h
 }
+
+// An OrderedStringDict is a mapping from names to values with support
```
adonovan, line 368: This code is all very well, but I wonder why it needs to be in this package, rather than in your application? It doesn't use any internals of the starlark package (other than `overloaded`, which is tiny, and hashString, which is exported as String.Hash).

If this exists only for the implementation of Struct, then let's move it into that package, but mutability and insertion ordering don't seem relevant to struct.

If a general need for a mutable insertion-ordered struct-like type arises across several projects, you could contribute your implementation to the lib/ tree, but it seems premature to put it here to start with.

starlark/hashtable.go
```diff
@@ -371,3 +364,186 @@ func softHashString(s string) uint32 {
...
+	// Append value to entries, linking the bucket to the entries list.
+	d.entries = append(d.entries, osdEntry{
+		hash:  h,
+		key:   k,
+		value: v,
+	})
+	d.table[n] = append(p, len(d.entries)-1)
+}
+
+func OrderStringDict(d StringDict) OrderedStringDict {
```
adonovan, line 472: This function should return a pointer. (In general, if type T has *T methods, it should never be copied by value.)

starlark/hashtable.go
```diff
@@ -371,3 +364,186 @@ func softHashString(s string) uint32 {
...
+// Keys are fast to add and access but slow to delete.
+// It is not a true starlark.Value.
+type OrderedStringDict struct {
+	// Implementation based on the above hashtable.
+	table   []osdBucket     // len is zero or a power of two
+	bucket0 [bucketSize]int //*osdEntry // inline allocation for small maps.
+	entries []osdEntry      // sorted list of entries
+}
+
+type osdBucket []int //*osdEntry
```
adonovan, line 379: // each element is the index into entries


Use uint32 to save space (and panic if len >= 4G).

starlark/hashtable.go
```diff
@@ -371,3 +364,186 @@ func softHashString(s string) uint32 {
...
 	return h
 }
+
+// An OrderedStringDict is a mapping from names to values with support
+// for fast indexing. Keys are stored in order of unique insertion.
+// Keys are fast to add and access but slow to delete.
+// It is not a true starlark.Value.
+type OrderedStringDict struct {
+	// Implementation based on the above hashtable.
+	table   []osdBucket     // len is zero or a power of two
```
adonovan, line 374: // Hash table that maps names to indices within entries.

starlark/hashtable.go
adonovan, line 368: "mutable mapping"

starlark/hashtable.go
```diff
@@ -371,3 +364,186 @@ func softHashString(s string) uint32 {
...
+	bucket0 [bucketSize]int //*osdEntry // inline allocation for small maps.
+	entries []osdEntry      // sorted list of entries
```
adonovan, line 376: // entries in order of insertion

starlark/hashtable.go
adonovan, line 472: The doc comment should state that the resulting value's fields are ordered by name.

starlark/hashtable.go
emcfarlane: This optimisation is added here: https://github.com/google/starlark-go/pull/404

starlark/hashtable.go
emcfarlane: My thinking was that most applications using starlark will implement custom objects they need to pass in and out of starlark scripts. My goto has been `starlark.StringDict` but then it does make it a little difficult to implement comparisons other methods. Seems like the missing type is hashmap with iteration support.

I modified starlarkstruct to showcase how it would bind to the new type.

I agree this is probably premature to add now, will use in my application to see more use cases, happy to close!

starlark/hashtable.go
emcfarlane: Removed in favour of a more generic `NewOrderedStringDict(size int)` that matches `NewSet` and `NewDict`.

adonovan:
> I agree this is probably premature to add now, will use in my application to see more use cases, happy to close!

Closing for now.

PR #404, opened 2022-03-04
starlark: set internal iteration use hashtable linked lists

emcfarlane:
Removes allocations on Set operations. Added a single benchmark testing
 equality, should improve other set operations too.

Old:
Benchmark/bench_set_equal-8                35050             34043 ns/op           16384 B/op          1 allocs/op

New:
Benchmark/bench_set_equal-8                41312             28884 ns/op               0 B/op          0 allocs/op

### DIFF
--- a/starlark/value.go
+++ b/starlark/value.go
@@ -7,35 +7,35 @@
 // Starlark values are represented by the Value interface.
 // The following built-in Value types are known to the evaluator:
 //
-//      NoneType        -- NoneType
-//      Bool            -- bool
-//      Bytes           -- bytes
-//      Int             -- int
-//      Float           -- float
-//      String          -- string
-//      *List           -- list
-//      Tuple           -- tuple
-//      *Dict           -- dict
-//      *Set            -- set
-//      *Function       -- function (implemented in Starlark)
-//      *Builtin        -- builtin_function_or_method (function or method implemented in Go)
+//	NoneType        -- NoneType
+//	Bool            -- bool
+//	Bytes           -- bytes
+//	Int             -- int
+//	Float           -- float
+//	String          -- string
+//	*List           -- list
+//	Tuple           -- tuple
+//	*Dict           -- dict
+//	*Set            -- set
+//	*Function       -- function (implemented in Starlark)
+//	*Builtin        -- builtin_function_or_method (function or method implemented in Go)
 //
 // Client applications may define new data types that satisfy at least
 // the Value interface.  Such types may provide additional operations by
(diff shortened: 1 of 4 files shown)


adonovan (commented):
(Sorry for the delay: I thought I'd sent my review comments months ago but they were still pending.)

starlark/value.go
```diff
@@ -276,7 +274,7 @@ type HasSetKey interface {
 var _ HasSetKey = (*Dict)(nil)
 
 // A HasBinary value may be used as either operand of these binary operators:
-//     +   -   *   /   //   %   in   not in   |   &   ^   <<   >>
+// +   -   *   /   //   %   in   not in   |   &   ^   <<   >>
```
emcfarlane, line 277: `//   + ...` will get formatted to `//  - ...` with the new go1.19 doc strings: https://tip.golang.org/doc/comment. To keep the plus I removed the padding from the start of the line. Is there a better way to solve this? Can remove doc string changes too.

starlark/value.go
```diff
@@ -296,7 +294,7 @@ const (
 )
 
 // A HasUnary value may be used as the operand of these unary operators:
-//     +   -   ~
+// +   -   ~
```
emcfarlane, line 297: Same here.

starlark/value.go
```diff
@@ -7,35 +7,35 @@
...
-//      Int             -- int
-//      Float           -- float
-//      String          -- string
-//      *List           -- list
-//      Tuple           -- tuple
-//      *Dict           -- dict
-//      *Set            -- set
-//      *Function       -- function (implemented in Starlark)
-//      *Builtin        -- builtin_function_or_method (function or method implemented in Go)
+//	NoneType        -- NoneType
```
emcfarlane, line 10: go1.19 formatting changes to tab.

emcfarlane:
Thanks for reviewing, sorry for the long delay in my changes to feedback too.  Have removed the allocation changes but introduced some doc changes for go1.19. Please let me know if this should be separated out.

starlark/value.go
adonovan: I'm inclined to leave it alone. The source is clear and even if the formatting loses one operator the meaning remains obvious.

adonovan: This looks fine. I don't see a perfect solution.

adonovan (approved):
In general one should be aware that iterating the linked list directly is not safe if the loop body might modify the hashtable, and in that case ht.elems() must be used, but all the changes here look good.

Thanks again.

PR #409, opened 2022-07-07
 Fix 'math' tests to tolerate float imprecision

bdd:
Came across this while fixing broken Starlark builds for Nixpkgs. It was already market broken for linux-aarch64 and failing for darwin-aarch64.

```
=== RUN   TestExecFile
    starlarktest.go:117: Traceback (most recent call last):
          testdata/math.star:258:10: in <toplevel>
          builtins/assert.star:15:14: in _eq
        Error: 0.9999999999999998 != 1
    starlarktest.go:117: Traceback (most recent call last):
          testdata/math.star:259:10: in <toplevel>
          builtins/assert.star:15:14: in _eq
        Error: -0.9999999999999998 != -1
```

I was reproducible with Go 1.16, 1.17, and 1.18 on both Linux and macOS with `math.Tan(math.Pi / 4)` as well. On amd64, it's consistently `1` but on `arm64` it's consistently `0.9999999999999998`.

`testdata/math.star` already provides a `near` helper for these floating point approximation issues.

This PR modifies two lines two fix the tests on arm64.


vcunat:
I'm no lawyer, but I'm convinced you don't really need CLA or other copyright stuff for such tiny contributions.

starlark/testdata/math.star
```diff
@@ -255,8 +255,8 @@ assert.fails(lambda: math.sin("0"), "got string, want float or int")
 # tan
 assert.eq(math.tan(0.0), 0)
 assert.eq(math.tan(0), 0)
-assert.eq(math.tan(math.pi / 4), 1)
-assert.eq(math.tan(-math.pi / 4), -1)
+assert.true(near(math.tan(math.pi / 4), 1, 0.00000000000001))
```
adonovan, line 258: Thanks for the fix. The code change looks good, but it's odd that there are 2 ulps of difference in the result
```
fmt.Printf("%[1]s: %[2]v %.16[2]x\n", runtime.GOARCH, math.Tan(math.Pi / 4))

arm64: 0.9999999999999998 0x1.ffffffffffffe000p-01
amd64: 1 0x1.0000000000000000p+00
```
even though both architectures use the [same software implementation](https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/math/tan.go;l=83;drc=8ac58de1857637f372a00ea16ab5497193b784a6). I wonder if this is due to greater width of Intel's FPU registers.

adonovan:
> I'm no lawyer, but I'm convinced you don't really need CLA or other copyright stuff for such tiny contributions.

Nor am I, but the CLA policy (written by Google's lawyers) seems to grant an exception only for changes to the copyright headers, but not for changes below some size threshold. Sorry.

adonovan (approved):

bdd:
Gentle ping.

adonovan:
Sorry for the delay---I thought I was waiting for the CLA check to pass, but it already had.

Thanks for the fix.

PR #414, opened 2022-08-03
CI: run benchmarks on go18 and go19

fenollp:
From [go19 release notes](https://go.dev/doc/go1.19#compiler):
> The compiler now uses a [jump table](https://en.wikipedia.org/wiki/Branch_table) to implement large integer and string switch statements. Performance improvements for the switch statement vary but can be on the order of 20% faster. (GOARCH=amd64 and GOARCH=arm64 only)


tetromino:
Fixed by Alan's cfacd890221418a2dc2c736f7b5e3476c38709b1

Apologies for the lack of review.

Issue #419, opened 2022-09-05
panic from built-in function leaves Thread in a bad state


vanishs:
This is the test code:
https://go.dev/play/p/SZFzWgchzeZ
```golang
package main

import (
	"fmt"
	"log"

	"go.starlark.net/starlark"
)

func check(err error) {
	if err != nil {
		log.Fatal(err)
	}
}

func main() {
	t := &starlark.Thread{}
	sd := starlark.StringDict{
		"test": starlark.NewBuiltin("test", func(*starlark.Thread, *starlark.Builtin, starlark.Tuple, []starlark.Tuple) (starlark.Value, error) {
			panic("panic")
			// println("123")
			return starlark.None, nil
		}),
	}
	g, err := starlark.ExecFile(t, "", `
def mytest():
	test()
`, sd)
	check(err)
	g.Freeze()
	v := g["mytest"]

	calltest := func() {
		defer func() {
			if e := recover(); e != nil {
				fmt.Println("get panic", e)
			}
		}()
		_, err = starlark.Call(t, v, make(starlark.Tuple, 0), nil)
		if err != nil {
			println("get err:", err.Error())
		}
	}

	calltest()
	calltest()

}

```

The output I am expecting is:
```
get panic panic
get panic panic
```

But the output I get is：
```
get panic panic
get err: function mytest called recursively
```

When starlark is aborted by Builtin panic, calling starlark.Call again will get an error. I would like a way to have starlark revert to a usable state without renew a Thread

adonovan:
You need to enable the [AllowRecursion flag](https://github.com/google/starlark-go/blob/b0039bd2cfe369fe8f2bdba0614bafd1f9402dbb/cmd/starlark/starlark.go#L41) to cause the interpreter not to fail when a recursive call is encountered. (This flag also enables the use of `while` loops.)

adonovan:
AllowRecursion is not the root problem, it's just one possible cause of panic. The fundamental issue here is that the interpreter is not designed to deal gracefully with panics in built-in functions. Perhaps it should? I've reopened the issue to track the decision.

vanishs:
Change from
`if !resolve.AllowRecursion {`
to
`if !resolve.AllowRecursion && thread.frameAt(0).locals != nil {`

what do you think?

vanishs:
change from
`if !resolve.AllowRecursion {`
to
`if !resolve.AllowRecursion && len(thread.stack) <= 1 {`

what do you think?

[...]

PR #421, opened 2022-09-06
Fix a thread bad status caused by panic

vanishs:
fix issue #419

### DIFF
--- a/starlark/interp.go
+++ b/starlark/interp.go
@@ -20,7 +20,7 @@ const vmdebug = false // TODO(adonovan): use a bitfield of specific kinds of err
 // - optimize position table.
 // - opt: record MaxIterStack during compilation and preallocate the stack.
 
-func (fn *Function) CallInternal(thread *Thread, args Tuple, kwargs []Tuple) (Value, error) {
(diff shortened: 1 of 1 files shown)


starlark/interp.go
```diff
@@ -83,8 +83,20 @@ func (fn *Function) CallInternal(thread *Thread, args Tuple, kwargs []Tuple) (Va
 
 	sp := 0
 	var pc uint32
-	var result Value
 	code := f.Code
+
+	defer func() {
+		if e := recover(); e != nil {
```
adonovan, line 89: Using recover is not part of the solution. If the interpreter encounters a panic, it's because the application code called by it has a bug, and the interpreter should not attempt to hide it.

There's still a question of whether the interpreter should use 'defer' in Call and CallInternal so that it pops the various stacks (Thread.stack, iterstack, profiler spans) to allow a panic to safely pass through the interpreter and be recovered in the application, without leaving the Thread in a bad state. It's an easy change, but whether we should make it depends on the cost it would add to returning from a Starlark function call in the normal case, for which we'd need a benchmark result.

adonovan: See #422

vanishs:
fix in #422

adonovan:
Fix still pending.

PR #422, opened 2022-09-06
starlark: use defer for cleanup so that panics pass through safely

### LINKED ISSUE #419: panic from built-in function leaves Thread in a bad state

adonovan:
This change uses defer for the cleanup logic in in Call and
Function.CallInternal so that panics originating in application
functions called by the interpreter can safely pass through
the interpreter without leaving the thread in a bad state
(e.g. frames still on the stack).

This seems to have an impact of ~0.7% on the 'calling' benchmark,
which represents the worst possible case.

Also, a test.

Fixes #419

### DIFF
--- a/starlark/eval.go
+++ b/starlark/eval.go
@@ -1225,8 +1225,22 @@ func Call(thread *Thread, fn Value, args Tuple, kwargs []Tuple) (Value, error) {
 	fr.callable = c
 
 	thread.beginProfSpan()
+
+	// Use defer to ensure that panics from built-ins
+	// pass through the interpreter without leaving
+	// it in a bad state.
+	defer func() {
+		thread.endProfSpan()
+
+		// clear out any references
+		// TODO(adonovan): opt: zero fr.Locals and
+		// reuse it if it is large enough.
+		*fr = frame{}
+
+		thread.stack = thread.stack[:len(thread.stack)-1] // pop
+	}()
+
 	result, err := c.CallInternal(thread, args, kwargs)
-	thread.endProfSpan()
 
 	// Sanity check: nil is not a valid Starlark value.
 	if result == nil && err == nil {
@@ -1239,9 +1253,6 @@ func Call(thread *Thread, fn Value, args Tuple, kwargs []Tuple) (Value, error) {
 			err = thread.evalError(err)
 		}
 	}
-
-	*fr = frame{}                                     // clear out any references
-	thread.stack = thread.stack[:len(thread.stack)-1] // pop
 
 	return result, err
 }
--- a/starlark/eval_test.go
+++ b/starlark/eval_test.go
@@ -995,3 +995,52 @@ func TestDeps(t *testing.T) {
 		}
 	}
 }
+
+// TestPanicSafety ensures that a panic from an application-defined
+// built-in may traverse the interpreter safely; see issue #411.
+func TestPanicSafety(t *testing.T) {
+	predeclared := starlark.StringDict{
+		"panic": starlark.NewBuiltin("panic", func(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+			panic(args[0])
+		}),
+		"list": starlark.NewList([]starlark.Value{starlark.MakeInt(0)}),
+	}
+
+	// This program is executed twice, using the same Thread,
+	// and panics both times, with values 1 and 2, while
+	// main is on the stack and a for-loop is active.
+	//
+	// It mutates list, a predeclared variable.
+	// This operation would fail if the previous
+	// for-loop failed to close its iterator during the panic.
+	//
+	// It also calls main a second time without recursion enabled.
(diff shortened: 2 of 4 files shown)


vanishs:
I've been using this branch for over a week and it works just fine.

In my program, the built-in functions I provided would use the panic trick to implement some functionality. but until this fix, the panic trick would leave the starlark stack in a wrong state.

I hope this branch will be merged soon.
thanks.

vanishs:
@brandjon hi. could you take a few time to do a review?

starlark/interp.go
```diff
@@ -81,6 +81,17 @@ func (fn *Function) CallInternal(thread *Thread, args Tuple, kwargs []Tuple) (Va
...
 
+	// Use defer so that application panics can pass through
+	// interpreter without leaving thread in a bad state.
+	defer func() {
+		// ITERPOP the rest of the iterator stack.
+		for _, iter := range iterstack {
+			iter.Done()
+		}
+
+		fr.locals = nil
```
brandjon, line 92: Nit: Do you think it's more readable to move this to a separate defer immediately after fr.locals is initialized?

starlark/eval.go
```diff
@@ -1225,8 +1225,22 @@ func Call(thread *Thread, fn Value, args Tuple, kwargs []Tuple) (Value, error) {
 	fr.callable = c
 
 	thread.beginProfSpan()
+
+	// Use defer to ensure that panics from built-ins
```
brandjon, line 1229: Should we document that the interpreter is still usable after an application-defined builtin panic()s? Or would this be conventionally assumed in Go?

starlark/eval.go
```diff
@@ -1225,8 +1225,22 @@ func Call(thread *Thread, fn Value, args Tuple, kwargs []Tuple) (Value, error) {
...
+	// pass through the interpreter without leaving
+	// it in a bad state.
+	defer func() {
+		thread.endProfSpan()
```
brandjon, line 1233: Nit: Readability risk: If someone adds more logic or cleanup following CallInternal, it'll be included in the profile span. We could 1) wrap the call to CallInternal in another function that just manages the profiling, or 2) add a comment below, or 3) ignore this if you think it's suitably obvious.

starlark/eval_test.go
```diff
@@ -995,3 +995,52 @@ func TestDeps(t *testing.T) {
...
+// built-in may traverse the interpreter safely; see issue #411.
+func TestPanicSafety(t *testing.T) {
+	predeclared := starlark.StringDict{
+		"panic": starlark.NewBuiltin("panic", func(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+			panic(args[0])
+		}),
+		"list": starlark.NewList([]starlark.Value{starlark.MakeInt(0)}),
+	}
+
+	// This program is executed twice, using the same Thread,
```
brandjon, line 1009: Nicely stated.

brandjon (approved):

starlark/interp.go
adonovan: Not really, and more importantly I don't want to add the cost of an additional deferred call to the loop.

starlark/eval.go
adonovan: I'd rather not make any promises about what the interpreter does when the client code is buggy. But it's easy enough for us to actually implement this property for now, and both the code here and the test in this CL record the intent for the maintainers.

starlark/eval.go
adonovan: That's true, but it's not worth complicating the control flow to avoid. And there's no law that says the prof span should include only the user code and not the other overheads that are inevitable when it is called.

PR #426, opened 2022-10-01
Bump go version to 1.19

ahrtr:
The pipeline is already using go1.18 and 1.19, so we'd better to bump go version in go.mod file as well. 

I intentionally created three different commits, see detailed info below.

Steps what I did:
1. Manually updated go version to 1.19 in the go.mod file;  See https://github.com/google/starlark-go/commit/a9d10e312664bb26fae99b0803e189555c981e7a
2. Format all go source file by just executing "`gofmt -w .`". See https://github.com/google/starlark-go/commit/591514d7356a5c3655b02e985b65dfdac248e4a1
3. Update all dependencies by just executing "`go mod tidy`". See https://github.com/google/starlark-go/commit/64f0299a4ab1f5edf81a940824d54ef85793fe16

### DIFF
--- a/docs/update.go
+++ b/docs/update.go
@@ -1,4 +1,5 @@
-//+build ignore
+//go:build ignore
+// +build ignore
 
 // The update command creates/updates the <html><head> elements of
 // each subpackage beneath docs so that "go get" requests redirect
@@ -6,9 +7,8 @@
 //
 // Usage:
 //
-//   $ cd $GOPATH/src/go.starlark.net
-//   $ go run docs/update.go
-//
+//	$ cd $GOPATH/src/go.starlark.net
+//	$ go run docs/update.go
 package main
 
 import (
--- a/go.mod
+++ b/go.mod
@@ -3,12 +3,15 @@ module go.starlark.net
 go 1.19
 
 require (
-	github.com/chzyer/logex v1.1.10 // indirect
 	github.com/chzyer/readline v0.0.0-20180603132655-2972be24d48e
-	github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1 // indirect
-	github.com/google/go-cmp v0.5.1 // indirect
 	golang.org/x/sys v0.0.0-20220715151400-c0bba94af5f8
 	golang.org/x/term v0.0.0-20220526004731-065cf7ba2467
-	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1 // indirect
 	google.golang.org/protobuf v1.25.0
+)
+
+require (
+	github.com/chzyer/logex v1.1.10 // indirect
+	github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1 // indirect
+	github.com/google/go-cmp v0.5.1 // indirect
+	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1 // indirect
 )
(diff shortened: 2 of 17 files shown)


adonovan:
I'm not sure what you mean by "the pipeline", but we shouldn't drop support for go1.18 (or even go1.17) yet. This will have to wait for another cycle. Sorry.

ahrtr:
> I'm not sure what you mean by "the pipeline", but we shouldn't drop support for go1.18 (or even go1.17) yet. This will have to wait for another cycle. Sorry.

Thanks for the response. Points from my side,
1. [pipeline](url) means github workflow, and the version is [1.18 and 1.19](https://github.com/google/starlark-go/blob/5fccb4daaf6d81d6de296eb308ddcd8a11d36b4b/.github/workflows/tests.yml#L8)
2. Bumping go version to 1.19 just means developers & contributors need to setup dev environment using go 1.19. It doesn't mean to drop support for any version. As long as the project continue to test against 1.18 and 1.19, then it means that it supports 1.18 and 1.19 at least.

adonovan:
> Bumping go version to 1.19 just means developers & contributors need to setup dev environment using go 1.19. It doesn't mean to drop support for any version. As long as the project continue to test against 1.18 and 1.19, then it means that it supports 1.18 and 1.19 at least.

The `go` version in the go.mod file states the minimum version of Go required by the package and enables certain features of the go build system. In fact, the tools try to be flexible by proceeding to execute a build even when the tool version is lower than the go.mod version, deferring the version error until some compilation fails (for any reason). If the project doesn't use, say, generics, it would build fine within a go1.19 mod file and a go1.18 compiler, and our CI workflow's test of go1.18 effectively ensures that we don't introduce go1.19 features. So you are right that updating it wouldn't fundamentally affect our users.

But what then is the value of upgrading the go declaration to go1.19 if we don't actually require go1.19?

(I was wrong about go1.17: we dropped it as part of #398.)

ahrtr:
Thanks for the response. 



> But what then is the value of upgrading the go declaration to go1.19 if we don't actually require go1.19?

If you introduce more workflow test/check (e.g. gofmt), then you will find different versions have different `gofmt` results. The [golangci](https://github.com/golangci/golangci-lint) also reads the go version from go.mod by default. 

`go mod tidy` also has different results if you set different go version in go.mod. 

So it would be better to clearly set a go version in go.mod, and clearly requires all contributors setup the dev environment using the same go version.

adonovan:
> Thanks for the response.
> 
> > But what then is the value of upgrading the go declaration to go1.19 if we don't actually require go1.19?
> 
> If you introduce more workflow test/check (e.g. gofmt), then you will find different versions have different `gofmt` results. The [golangci](https://github.com/golangci/golangci-lint) also reads the go version from go.mod by default.

In effect this proposes to add an unnecessary minimum version constraint to all users of this package in order to slightly reduce the burden on the developers of the package, who might forget to test with the latest Go toolchain. I don't think that's the right trade-off, as there are many more of the former than the latter.

ahrtr:
I think we need to clearly differentiate two roles: `users of this package `and `contributors of this package`. Note that there isn't a minimum version constraint to all users, they can use whatever go version in their applications which depend on starlark-go. 

For contributors, I already explained the reasons in my previous comment. 

go 1.13 is out of support for a long time, it's not a good pattern to always pin a old & unsupported go version in go.mod.

ahrtr:
Closing this ticket as I do not have bandwidth to follow up it.

PR #427, opened 2022-10-05
fix a few function names on comments

cuishuang:


fenollp (approved):

adonovan (approved):
Thanks for the correction!

PR #428, opened 2022-10-05
Update spec.md to correct some mistakes

ahrtr:
Changed items:
1. The section `Module execution` has the same level as `Statements`, both has the prefix "##", but it's just a sub item under `Statements`. It seems not correct. So increased `Module execution` to +1 level.
2. `#name-resolution` is a wrong link, so replaced it with `#built-in-constants-and-functions`.
3. The lambda example has wrong output, corrected.
4. typo, `element` --> `tuple`

There might be another error in https://github.com/google/starlark-go/blob/master/doc/spec.md#lists , I am not quite sure about it and don't know how to update it. See below,
```
For most types, x += y is equivalent to x = x + y, except that it evaluates x only once, that is, it allocates a new list to hold the concatenation of x and y. However, if x refers to a list, the statement does not allocate a new list but instead mutates the original list in place, similar to x.extend(y).
```

My immediate feeling is that the `"it allocates a new list"` isn't correct. The following `"However, if x refers to a list"` indicates that previous sentence is talking about most other types instead of list.

### DIFF
--- a/doc/spec.md
+++ b/doc/spec.md
@@ -104,7 +104,7 @@ reproducibility is paramount, such as build tools.
     * [For loops](#for-loops)
     * [Break and Continue](#break-and-continue)
     * [Load statements](#load-statements)
-    * [Module execution](#module-execution)
+  * [Module execution](#module-execution)
   * [Built-in constants and functions](#built-in-constants-and-functions)
     * [None](#none)
     * [True and False](#true-and-false)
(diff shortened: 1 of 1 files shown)


doc/spec.md
```diff
@@ -1193,7 +1193,7 @@ The [type](#type) of a built-in function is `"builtin_function_or_method"`.
 A built-in function value used in a Boolean context is always considered true.
 
 Many built-in functions are predeclared in the environment
-(see [Name Resolution](#name-resolution)).
+(see [Built-in constants and functions](#built-in-constants-and-functions)).
```
adonovan, line 1196: I think the previous link was in support of the concept of "predeclared". How about:

```
Many [built-in functions](#built-in-constants-and-functions) are
[predeclared](#name-resolution) in the environment.
```
?

Or remove the first link and keep the "(see...)" remark as you have it.

adonovan (commented):
Well spotted! Thanks for the fixes.

doc/spec.md
ahrtr: Thanks for the comment. 

I think it would be better to add links as much as we can because it can improve the users' reading experience. 

Yous proposed change still has link `#name-resolution`, which doesn't exist at all. I guess you wanted to say `#name-binding-and-variables`?

Just updated. PTAL, thx

ahrtr:
> It's correct in that the current section is about lists, but I agree it could be clearer. How about:
> 
> ```
> For most types, `x += y` is equivalent to `x = x + y`, except that it
> evaluates `x` only once. However, if `x` refers to a list, the statement
> `x += y` does not allocate a new list as `x = x + y` would,
> but instead mutates the original list in place, similar to `x.extend(y)`.
> ```

Makes sense & updated & thx

adonovan (approved):
Thanks again!

PR #431, opened 2022-10-13
Detecting json loops

marco6:
Json encoding is not checking for loops in Starlark values, so code like this:

```python
a = {}
a["a"] = a
json.encode(a)
```

crash the application with a stack overflow like:
```plain
runtime: goroutine stack exceeds 1000000000-byte limit
runtime: sp=0xc020260370 stack=[0xc020260000, 0xc040260000]
fatal error: stack overflow
<...>
```
This PR should fix the behavior.

### DIFF
--- a/lib/json/json.go
+++ b/lib/json/json.go
@@ -106,9 +106,13 @@ func encode(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, k
 
 	var emit func(x starlark.Value) error
 	emit = func(x starlark.Value) error {
+
+		// It is only necessary to push/pop the item when it might contain
+		// itself (i.e. the last three switch cases), but omitting it in the other
+		// cases did not show significant improvement on the benchmarks.
 		if ptr := pointer(x); ptr != nil {
 			if pathContains(path, ptr) {
-				return fmt.Errorf("Detected cycle in json structure")
+				return fmt.Errorf("cycle in json structure")
 			}
 
 			path = append(path, ptr)


lib/json/json.go
```diff
@@ -100,8 +111,17 @@ func encode(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, k
...
-	var emit func(x starlark.Value) error
-	emit = func(x starlark.Value) error {
+	var emit func(x starlark.Value, pointers map[uintptr]struct{}) error
+	emit = func(x starlark.Value, pointers map[uintptr]struct{}) error {
+		if ptr, ok := getPointer(x); ok {
+			if _, ok = pointers[ptr]; ok {
+				return fmt.Errorf("Detected cycle in json structure")
+			}
+
+			pointers[ptr] = struct{}{}
```
adonovan, line 121: I suspect it would be more efficient in practice to represent the set of active pointers not as a map but as a slice, which we push and pop in stack-like fashion, as the stack rarely gets deep enough for the hash table to pay off.  Also, the push and pop operations should be done only in the last three cases of the switch, but not for each scalar value.

See the implementation of [writeValue](https://github.com/google/starlark-go/blob/f738f5508c12fe5a9fae44bbdf07a94ddcf5030e/starlark/value.go#L1155-L1162) for the pattern.

This function must deal with arbitrary Values that satisfy Iterable, Mapping, and so on, whereas writeValue only deals with known specific types, which makes its task easier. For example, it can safely omit the recursion check for Tuple because there's no way a Starlark program can create a recursive tuple.  I see that your code is careful not to compare Values for == equality, which is not safe for (e.g.) Tuples, which are represented as slices.

marco6: Hi @adonovan! Thank you for your review! 
I actually almost copy-pasted this solution from the [Go json marshaller](https://github.com/golang/go/blob/master/src/encoding/json/encode.go#L295) the only difference being that I don't wait for `startDetectingCyclesAfter = 1000` levels of recursion, but I check right away. The reason for this difference is the error message: Starlark returns a descriptive error which contains the json path to the error (the loop in this case). I didn't think that it would have been useful if it included 1000 times `in dict key "x":`.

However, I am very open to try to use a slice instead. If you think it's better, I will try to first write a benchmark for it.

marco6: I wrote 4 benchmarks to simple use-cases (and I rebased them on top so that it was easier to test the change commit-by-commit).

I updated the code using slices. Seems like you were right about slices. I put together some data to reason about, commit by commit (on my Intel Core i7-7567U) which I add here, split by test.

# bench_to_json_deep_list
image

# bench_to_json_deep
image

# bench_to_json_flat_big
image

# bench_to_json_flat_mixed
image

lib/json/json.go
```diff
@@ -23,6 +24,16 @@ import (
 	"go.starlark.net/starlarkstruct"
 )
 
+func getPointer(i interface{}) (uintptr, bool) {
+	v := reflect.ValueOf(i)
+	switch v.Kind() {
+	case reflect.Ptr, reflect.Chan, reflect.Map, reflect.UnsafePointer, reflect.Slice:
+		return v.Pointer(), true
```
adonovan, line 31: Let's use v.UnsafePointer() and type unsafe.Pointer instead, since we don't need the numeric value of the pointer and we're not trying to hide anything from the GC.

lib/json/json.go
adonovan: Thanks. Could you also do this part?

> Also, the push and pop operations should be done only in the last three cases of the switch, but not for each scalar value.

Rather than "push" and "defer pop" you can simply pass the result of append in the recursive call. For example:
```
	case starlark.IterableMapping:
		if ptr := pointer(x); pathContains(path, ptr) {
			...error...
		} else {
                          ...emit(subelement, append(path, ptr))...
		}
```

lib/json/json.go
```diff
@@ -23,6 +24,16 @@ import (
 	"go.starlark.net/starlarkstruct"
 )
 
+func getPointer(i interface{}) (uintptr, bool) {
```
adonovan, line 27: Let's call this `pointer`, and make it return nil instead of (0, false). (It seems highly unlikely that anyone would declare a starlark.Iterable value that contains itself and yet whose representation is a nil pointer, so let's not worry about the boolean.)

Also, let's move it after the main function that uses it.

lib/json/json.go
marco6: If i change "push" and "defer pop" as you suggested it will waste some allocations while using them will help reuse the memory (and offset the cost of allocation + copying of the elements).

I will try to explain me better. Let's imagine I used as a default `path := make([]uintptr, 2)` (just 2 instead of 8 as I did in the PR).

In the case of this object (operations annotated as comments):
```python
{# saving base object in position 0, no reallocation
  "a" : { # saving $.a in position 1
    "b": [ # doubling the size of the slice with a buffer from the heap and copying the first 2 elements. The result will be saved **on the stack** of this call. Saving $.a.b in position 2
      1, # not saving anything (not a ptr)
      2, # not saving anything (not a ptr)
      3 # not saving anything (not a ptr)
    ] # returning. This will delete the newly allocated buffer (since I didn't save the result of append)
  }, # returning. This will delete the path, but the buffer is still held by the initial call
  "d": { # doubling the size of the slice with a buffer from the heap and copying the first 2 elements. The result will be saved **on the stack** of this call. Saving $.a.d in position 2
    "e": 1 # not saving anything
  }
}
```

marco6: Also, the push and pop will appen just in case the `Value` is a pointer-like, otherwise the `getPointer` function will return `false`...

adonovan: I see what you are saying: that my approach would pay the cost of reallocation on every call (not just the first) that crosses the depth threshold. That's true. But a call to append is cheap if there is capacity in the array, which there will be if the initial capacity in make() is larger than the depth of the tree. Most JSON trees have a depth less than 10, so if you allocate, say, 32 elements, calls that cross the depth threshold will be vanishingly rare.

> Also, the push and pop will happen just in case the Value is a pointer-like, otherwise the getPointer function will return false...

If getPointer returns nil, we shouldn't check/push/pop the stack. If getPointer returns false, we shouldn't check/push/pop the stack. So why not merge those two cases?

marco6: I updated this PR to reflect your comments.

Regarding the push-pop, I tried some combination. Sorry for the long post, but github won't let me attach diffs.

## Baseline

On my CPU the current version gives:

```bash
> go test -run=$^ -bench=Starlark . -benchmem
...
BenchmarkStarlark/bench_to_json_deep-4                     98647             10966 ns/op            2848 B/op         76 allocs/op
BenchmarkStarlark/bench_to_json_deep_list-4               124208              9750 ns/op            2848 B/op         76 allocs/op
BenchmarkStarlark/bench_to_json_flat_big-4                  1731            697163 ns/op          153084 B/op       6017 allocs/op
BenchmarkStarlark/bench_to_json_flat_mixed-4              238779              4759 ns/op            1168 B/op         41 allocs/op
```

## Checking pointers only in the last 3 cases, leaving append/defer

The diff of the changes is:

```diff
diff --git a/lib/json/json.go b/lib/json/json.go
index a55cbde..f57584f 100644
--- a/lib/json/json.go
+++ b/lib/json/json.go
@@ -106,15 +106,6 @@ func encode(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, k
 
 	var emit func(x starlark.Value) error
 	emit = func(x starlark.Value) error {
-		if ptr := pointer(x); ptr != nil {
-			if pathContains(path, ptr) {
-				return fmt.Errorf("Detected cycle in json structure")
-			}
-
-			path = append(path, ptr)
-			defer func() { path = path[0 : len(path)-1] }()
-		}
-
 		switch x := x.(type) {
 		case json.Marshaler:
 			// Application-defined starlark.Value types
@@ -147,68 +138,80 @@ func encode(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, k
 		case starlark.String:
 			quote(string(x))
 
-		case starlark.IterableMapping:
-			// e.g. dict (must have string keys)
-			buf.WriteByte('{')
-			items := x.Items()
-			for _, item := range items {
-				if _, ok := item[0].(starlark.String); !ok {
-					return fmt.Errorf("%s has %s key, want string", x.Type(), item[0].Type())
-				}
-			}
-			sort.Slice(items, func(i, j int) bool {
-				return items[i][0].(starlark.String) < items[j][0].(starlark.String)
-			})
-			for i, item := range items {
-				if i > 0 {
-					buf.WriteByte(',')
-				}
-				k, _ := starlark.AsString(item[0])
-				quote(k)
-				buf.WriteByte(':')
-				if err := emit(item[1]); err != nil {
-					return fmt.Errorf("in %s key %s: %v", x.Type(), item[0], err)
+		case starlark.IterableMapping, starlark.Iterable, starlark.HasAttrs:
+			if ptr := pointer(x); ptr != nil {
+				if pathContains(path, ptr) {
+					return fmt.Errorf("Detected cycle in json structure")
 				}
+
+				path = append(path, ptr)
+				defer func() { path = path[0 : len(path)-1] }()
 			}
-			buf.WriteByte('}')
-
-		case starlark.Iterable:
-			// e.g. tuple, list
-			buf.WriteByte('[')
-			iter := x.Iterate()
-			defer iter.Done()
-			var elem starlark.Value
-			for i := 0; iter.Next(&elem); i++ {
-				if i > 0 {
-					buf.WriteByte(',')
-				}
-				if err := emit(elem); err != nil {
-					return fmt.Errorf("at %s index %d: %v", x.Type(), i, err)
+
+			switch x := x.(type) {
+			case starlark.IterableMapping:
+				// e.g. dict (must have string keys)
+				buf.WriteByte('{')
+				items := x.Items()
+				for _, item := range items {
+					if _, ok := item[0].(starlark.String); !ok {
+						return fmt.Errorf("%s has %s key, want string", x.Type(), item[0].Type())
+					}
 				}
-			}
-			buf.WriteByte(']')
-
-		case starlark.HasAttrs:
-			// e.g. struct
-			buf.WriteByte('{')
-			var names []string
-			names = append(names, x.AttrNames()...)
-			sort.Strings(names)
-			for i, name := range names {
-				v, err := x.Attr(name)
-				if err != nil || v == nil {
-					log.Fatalf("internal error: dir(%s) includes %q but value has no .%s field", x.Type(), name, name)
+				sort.Slice(items, func(i, j int) bool {
+					return items[i][0].(starlark.String) < items[j][0].(starlark.String)
+				})
+				for i, item := range items {
+					if i > 0 {
+						buf.WriteByte(',')
+					}
+					k, _ := starlark.AsString(item[0])
+					quote(k)
+					buf.WriteByte(':')
+					if err := emit(item[1]); err != nil {
+						return fmt.Errorf("in %s key %s: %v", x.Type(), item[0], err)
+					}
 				}
-				if i > 0 {
-					buf.WriteByte(',')
+				buf.WriteByte('}')
+
+			case starlark.Iterable:
+				// e.g. tuple, list
+				buf.WriteByte('[')
+				iter := x.Iterate()
+				defer iter.Done()
+				var elem starlark.Value
+				for i := 0; iter.Next(&elem); i++ {
+					if i > 0 {
+						buf.WriteByte(',')
+					}
+					if err := emit(elem); err != nil {
+						return fmt.Errorf("at %s index %d: %v", x.Type(), i, err)
+					}
 				}
-				quote(name)
-				buf.WriteByte(':')
-				if err := emit(v); err != nil {
-					return fmt.Errorf("in field .%s: %v", name, err)
+				buf.WriteByte(']')
+
+			case starlark.HasAttrs:
+				// e.g. struct
+				buf.WriteByte('{')
+				var names []string
+				names = append(names, x.AttrNames()...)
+				sort.Strings(names)
+				for i, name := range names {
+					v, err := x.Attr(name)
+					if err != nil || v == nil {
+						log.Fatalf("internal error: dir(%s) includes %q but value has no .%s field", x.Type(), name, name)
+					}
+					if i > 0 {
+						buf.WriteByte(',')
+					}
+					quote(name)
+					buf.WriteByte(':')
+					if err := emit(v); err != nil {
+						return fmt.Errorf("in field .%s: %v", name, err)
+					}
 				}
+				buf.WriteByte('}')
 			}
-			buf.WriteByte('}')
 
 		default:
 			return fmt.Errorf("cannot encode %s as JSON", x.Type())
```

The result is:

```bash
> go test -run=$^ -bench=Starlark . -benchmem
BenchmarkStarlark/bench_to_json_deep-4                    119862             10694 ns/op            2848 B/op         76 allocs/op
BenchmarkStarlark/bench_to_json_deep_list-4               128067              9612 ns/op            2848 B/op         76 allocs/op
BenchmarkStarlark/bench_to_json_flat_big-4                  1737            695309 ns/op          153084 B/op       6017 allocs/op
BenchmarkStarlark/bench_to_json_flat_mixed-4              272838              4392 ns/op            1168 B/op         41 allocs/op
```
## Checking pointers only in the last 3 cases, avoiding defer

The diff is:

```diff
diff --git a/lib/json/json.go b/lib/json/json.go
index a55cbde..84f3eb3 100644
--- a/lib/json/json.go
+++ b/lib/json/json.go
@@ -106,15 +106,6 @@ func encode(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, k
 
 	var emit func(x starlark.Value) error
 	emit = func(x starlark.Value) error {
-		if ptr := pointer(x); ptr != nil {
-			if pathContains(path, ptr) {
-				return fmt.Errorf("Detected cycle in json structure")
-			}
-
-			path = append(path, ptr)
-			defer func() { path = path[0 : len(path)-1] }()
-		}
-
 		switch x := x.(type) {
 		case json.Marshaler:
 			// Application-defined starlark.Value types
@@ -147,69 +138,82 @@ func encode(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, k
 		case starlark.String:
 			quote(string(x))
 
-		case starlark.IterableMapping:
-			// e.g. dict (must have string keys)
-			buf.WriteByte('{')
-			items := x.Items()
-			for _, item := range items {
-				if _, ok := item[0].(starlark.String); !ok {
-					return fmt.Errorf("%s has %s key, want string", x.Type(), item[0].Type())
-				}
-			}
-			sort.Slice(items, func(i, j int) bool {
-				return items[i][0].(starlark.String) < items[j][0].(starlark.String)
-			})
-			for i, item := range items {
-				if i > 0 {
-					buf.WriteByte(',')
-				}
-				k, _ := starlark.AsString(item[0])
-				quote(k)
-				buf.WriteByte(':')
-				if err := emit(item[1]); err != nil {
-					return fmt.Errorf("in %s key %s: %v", x.Type(), item[0], err)
+		case starlark.IterableMapping, starlark.Iterable, starlark.HasAttrs:
+			oldPathLen := len(path)
+			if ptr := pointer(x); ptr != nil {
+				if pathContains(path, ptr) {
+					return fmt.Errorf("Detected cycle in json structure")
 				}
+
+				path = append(path, ptr)
 			}
-			buf.WriteByte('}')
-
-		case starlark.Iterable:
-			// e.g. tuple, list
-			buf.WriteByte('[')
-			iter := x.Iterate()
-			defer iter.Done()
-			var elem starlark.Value
-			for i := 0; iter.Next(&elem); i++ {
-				if i > 0 {
-					buf.WriteByte(',')
-				}
-				if err := emit(elem); err != nil {
-					return fmt.Errorf("at %s index %d: %v", x.Type(), i, err)
+
+			switch x := x.(type) {
+			case starlark.IterableMapping:
+				// e.g. dict (must have string keys)
+				buf.WriteByte('{')
+				items := x.Items()
+				for _, item := range items {
+					if _, ok := item[0].(starlark.String); !ok {
+						return fmt.Errorf("%s has %s key, want string", x.Type(), item[0].Type())
+					}
 				}
-			}
-			buf.WriteByte(']')
-
-		case starlark.HasAttrs:
-			// e.g. struct
-			buf.WriteByte('{')
-			var names []string
-			names = append(names, x.AttrNames()...)
-			sort.Strings(names)
-			for i, name := range names {
-				v, err := x.Attr(name)
-				if err != nil || v == nil {
-					log.Fatalf("internal error: dir(%s) includes %q but value has no .%s field", x.Type(), name, name)
+				sort.Slice(items, func(i, j int) bool {
+					return items[i][0].(starlark.String) < items[j][0].(starlark.String)
+				})
+				for i, item := range items {
+					if i > 0 {
+						buf.WriteByte(',')
+					}
+					k, _ := starlark.AsString(item[0])
+					quote(k)
+					buf.WriteByte(':')
+					if err := emit(item[1]); err != nil {
+						return fmt.Errorf("in %s key %s: %v", x.Type(), item[0], err)
+					}
 				}
-				if i > 0 {
-					buf.WriteByte(',')
+				buf.WriteByte('}')
+
+			case starlark.Iterable:
+				// e.g. tuple, list
+				buf.WriteByte('[')
+				iter := x.Iterate()
+				defer iter.Done()
+				var elem starlark.Value
+				for i := 0; iter.Next(&elem); i++ {
+					if i > 0 {
+						buf.WriteByte(',')
+					}
+					if err := emit(elem); err != nil {
+						return fmt.Errorf("at %s index %d: %v", x.Type(), i, err)
+					}
 				}
-				quote(name)
-				buf.WriteByte(':')
-				if err := emit(v); err != nil {
-					return fmt.Errorf("in field .%s: %v", name, err)
+				buf.WriteByte(']')
+
+			case starlark.HasAttrs:
+				// e.g. struct
+				buf.WriteByte('{')
+				var names []string
+				names = append(names, x.AttrNames()...)
+				sort.Strings(names)
+				for i, name := range names {
+					v, err := x.Attr(name)
+					if err != nil || v == nil {
+						log.Fatalf("internal error: dir(%s) includes %q but value has no .%s field", x.Type(), name, name)
+					}
+					if i > 0 {
+						buf.WriteByte(',')
+					}
+					quote(name)
+					buf.WriteByte(':')
+					if err := emit(v); err != nil {
+						return fmt.Errorf("in field .%s: %v", name, err)
+					}
 				}
+				buf.WriteByte('}')
 			}
-			buf.WriteByte('}')
 
+			path = path[0:oldPathLen]
 		default:
 			return fmt.Errorf("cannot encode %s as JSON", x.Type())
 		}
```

The results are:
```bash
> go test -run=$^ -bench=Starlark . -benchmem
BenchmarkStarlark/bench_to_json_deep-4                    119103              9342 ns/op            2848 B/op         76 allocs/op
BenchmarkStarlark/bench_to_json_deep_list-4               111163              9241 ns/op            2848 B/op         76 allocs/op
BenchmarkStarlark/bench_to_json_flat_big-4                  1881            654173 ns/op          153084 B/op       6017 allocs/op
BenchmarkStarlark/bench_to_json_flat_mixed-4              246733              4582 ns/op            1168 B/op         41 allocs/op
```

## Appending on the stack

The diff:
```diff
diff --git a/lib/json/json.go b/lib/json/json.go
index a55cbde..3eaaac9 100644
--- a/lib/json/json.go
+++ b/lib/json/json.go
@@ -102,19 +102,8 @@ func encode(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, k
 		}
 	}
 
-	path := make([]unsafe.Pointer, 0, 8)
-
-	var emit func(x starlark.Value) error
-	emit = func(x starlark.Value) error {
-		if ptr := pointer(x); ptr != nil {
-			if pathContains(path, ptr) {
-				return fmt.Errorf("Detected cycle in json structure")
-			}
-
-			path = append(path, ptr)
-			defer func() { path = path[0 : len(path)-1] }()
-		}
-
+	var emit func(x starlark.Value, path []unsafe.Pointer) error
+	emit = func(x starlark.Value, path []unsafe.Pointer) error {
 		switch x := x.(type) {
 		case json.Marshaler:
 			// Application-defined starlark.Value types
@@ -147,68 +136,79 @@ func encode(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, k
 		case starlark.String:
 			quote(string(x))
 
-		case starlark.IterableMapping:
-			// e.g. dict (must have string keys)
-			buf.WriteByte('{')
-			items := x.Items()
-			for _, item := range items {
-				if _, ok := item[0].(starlark.String); !ok {
-					return fmt.Errorf("%s has %s key, want string", x.Type(), item[0].Type())
-				}
-			}
-			sort.Slice(items, func(i, j int) bool {
-				return items[i][0].(starlark.String) < items[j][0].(starlark.String)
-			})
-			for i, item := range items {
-				if i > 0 {
-					buf.WriteByte(',')
-				}
-				k, _ := starlark.AsString(item[0])
-				quote(k)
-				buf.WriteByte(':')
-				if err := emit(item[1]); err != nil {
-					return fmt.Errorf("in %s key %s: %v", x.Type(), item[0], err)
+		case starlark.IterableMapping, starlark.Iterable, starlark.HasAttrs:
+			if ptr := pointer(x); ptr != nil {
+				if pathContains(path, ptr) {
+					return fmt.Errorf("Detected cycle in json structure")
 				}
+
+				path = append(path, ptr)
 			}
-			buf.WriteByte('}')
-
-		case starlark.Iterable:
-			// e.g. tuple, list
-			buf.WriteByte('[')
-			iter := x.Iterate()
-			defer iter.Done()
-			var elem starlark.Value
-			for i := 0; iter.Next(&elem); i++ {
-				if i > 0 {
-					buf.WriteByte(',')
-				}
-				if err := emit(elem); err != nil {
-					return fmt.Errorf("at %s index %d: %v", x.Type(), i, err)
+
+			switch x := x.(type) {
+			case starlark.IterableMapping:
+				// e.g. dict (must have string keys)
+				buf.WriteByte('{')
+				items := x.Items()
+				for _, item := range items {
+					if _, ok := item[0].(starlark.String); !ok {
+						return fmt.Errorf("%s has %s key, want string", x.Type(), item[0].Type())
+					}
 				}
-			}
-			buf.WriteByte(']')
-
-		case starlark.HasAttrs:
-			// e.g. struct
-			buf.WriteByte('{')
-			var names []string
-			names = append(names, x.AttrNames()...)
-			sort.Strings(names)
-			for i, name := range names {
-				v, err := x.Attr(name)
-				if err != nil || v == nil {
-					log.Fatalf("internal error: dir(%s) includes %q but value has no .%s field", x.Type(), name, name)
+				sort.Slice(items, func(i, j int) bool {
+					return items[i][0].(starlark.String) < items[j][0].(starlark.String)
+				})
+				for i, item := range items {
+					if i > 0 {
+						buf.WriteByte(',')
+					}
+					k, _ := starlark.AsString(item[0])
+					quote(k)
+					buf.WriteByte(':')
+					if err := emit(item[1], path); err != nil {
+						return fmt.Errorf("in %s key %s: %v", x.Type(), item[0], err)
+					}
 				}
-				if i > 0 {
-					buf.WriteByte(',')
+				buf.WriteByte('}')
+
+			case starlark.Iterable:
+				// e.g. tuple, list
+				buf.WriteByte('[')
+				iter := x.Iterate()
+				defer iter.Done()
+				var elem starlark.Value
+				for i := 0; iter.Next(&elem); i++ {
+					if i > 0 {
+						buf.WriteByte(',')
+					}
+					if err := emit(elem, path); err != nil {
+						return fmt.Errorf("at %s index %d: %v", x.Type(), i, err)
+					}
 				}
-				quote(name)
-				buf.WriteByte(':')
-				if err := emit(v); err != nil {
-					return fmt.Errorf("in field .%s: %v", name, err)
+				buf.WriteByte(']')
+
+			case starlark.HasAttrs:
+				// e.g. struct
+				buf.WriteByte('{')
+				var names []string
+				names = append(names, x.AttrNames()...)
+				sort.Strings(names)
+				for i, name := range names {
+					v, err := x.Attr(name)
+					if err != nil || v == nil {
+						log.Fatalf("internal error: dir(%s) includes %q but value has no .%s field", x.Type(), name, name)
+					}
+					if i > 0 {
+						buf.WriteByte(',')
+					}
+					quote(name)
+					buf.WriteByte(':')
+					if err := emit(v, path); err != nil {
+						return fmt.Errorf("in field .%s: %v", name, err)
+					}
 				}
+				buf.WriteByte('}')
 			}
-			buf.WriteByte('}')
 
 		default:
 			return fmt.Errorf("cannot encode %s as JSON", x.Type())
@@ -216,7 +216,7 @@ func encode(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, k
 		return nil
 	}
 
-	if err := emit(x); err != nil {
+	if err := emit(x, make([]unsafe.Pointer, 0, 8)); err != nil {
 		return nil, fmt.Errorf("%s: %v", b.Name(), err)
 	}
 	return starlark.String(buf.String()), nil
```

The results:
```bash
> go test -run=$^ -bench=Starlark . -benchmem
BenchmarkStarlark/bench_to_json_deep-4                    117927              9276 ns/op            2912 B/op         77 allocs/op
BenchmarkStarlark/bench_to_json_deep_list-4               128575             10556 ns/op            2912 B/op         77 allocs/op
BenchmarkStarlark/bench_to_json_flat_big-4                  1616            655284 ns/op          153147 B/op       6018 allocs/op
BenchmarkStarlark/bench_to_json_flat_mixed-4              269059              4693 ns/op            1232 B/op         42 allocs/op
```

Changing the default to 32 items seems to get worse (all allocations increase and time/op increases as well).

Given these results, if you are happy with it, I will push the first variation since it's memory footprint seems the same of the second one, but it feels more Go-ish IMHO.

lib/json/json.go
marco6: Done

lib/json/json.go
marco6: Done. I also moved `pathContains` after the function to be in line with this guideline.

lib/json/json.go
```diff
@@ -100,8 +102,19 @@ func encode(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, k
 		}
 	}
 
+	path := make([]unsafe.Pointer, 0, 8)
+
 	var emit func(x starlark.Value) error
 	emit = func(x starlark.Value) error {
+		if ptr := pointer(x); ptr != nil {
```
adonovan, line 109: // It is only necessary to push/pop the item when it might contain
// itself (i.e. the last three switch cases), but omitting it in the other
// cases did not show significant improvement on the benchmarks.

lib/json/json.go
```diff
@@ -100,8 +102,19 @@ func encode(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, k
...
+			if pathContains(path, ptr) {
+				return fmt.Errorf("Detected cycle in json structure")
```
adonovan, line 111: "cycle in..."

(no need for "Detected"; also lower case)

adonovan (approved):
Thanks for being so thorough in your evaluation of the alternatives. The code looks good aside from two very minor tweaks.

lib/json/json.go
marco6: Done

PR #432, opened 2022-10-13
Fixing example in documentation

marco6:
`String` type is not directly indexable (at least not in go-starlark).


adonovan (approved):

adonovan:
Thanks @marco6!

PR #433, opened 2022-10-17
Imlemented ResetStepsAndCancelReason()

vanishs:
Imlemented ResetStepsAndCancelReason() and a test
This ResetStepsAndCancelReason() is very helpful for limiting cpu usage per Call or CallInternal

### DIFF
--- a/starlark/eval.go
+++ b/starlark/eval.go
@@ -84,6 +84,12 @@ func (thread *Thread) SetMaxExecutionSteps(max uint64) {
 	thread.maxSteps = max
 }
 
+// ResetStepsAndCancelReason reset steps counts and cancel reason
+func (thread *Thread) ResetStepsAndCancelReason() {
+	thread.steps = 0
+	thread.cancelReason = nil
+}
+
 // Cancel causes execution of Starlark code in the specified thread to
 // promptly fail with an EvalError that includes the specified reason.
 // There may be a delay before the interpreter observes the cancellation
--- a/starlark/eval.go
+++ b/starlark/eval.go
@@ -51,8 +51,13 @@ type Thread struct {
 	// The default behavior is to call thread.Cancel("too many steps").
 	OnMaxSteps func(thread *Thread)
 
-	// steps counts abstract computation steps executed by this thread.
-	steps, maxSteps uint64
+	// Steps a count of abstract computation steps executed
+	// by this thread. It is incremented by the interpreter. It may be used
+	// as a measure of the approximate cost of Starlark execution, by
+	// computing the difference in its value before and after a computation.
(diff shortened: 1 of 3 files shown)


starlark/eval.go
```diff
@@ -84,6 +84,12 @@ func (thread *Thread) SetMaxExecutionSteps(max uint64) {
 	thread.maxSteps = max
 }
 
+// ResetStepsAndCancelReason reset steps counts and cancel reason
+func (thread *Thread) ResetStepsAndCancelReason() {
+	thread.steps = 0
```
adonovan, line 89: If we're going to allow users to reset it, then I don't really see any reason not to make the steps variable public; let's do that, and move the comment from ExecutionSteps to it. ExecutionSteps (which we must retain for compatibility) can just say "returns the current value of Steps".

Then this function need only reset the cancellation state. Call it Uncancel, and update the comment "Cancellation cannot be undone" to read "Call [Uncancel] to reset the cancellation state".  For symmetry, this method should use atomic.SetPointer and copy the "Unlike most methods..." comment from Cancel.

(The fact that we use the "reason" *string as the representation is an implementation detail that doesn't belong in the method name.)

adonovan (commented):
Thanks for the PR. This seems reasonable, though I have some comments on the details.

starlark/eval.go
vanishs: Based on your suggestion, I submitted some changes

starlark/eval.go
```diff
@@ -65,14 +70,9 @@ type Thread struct {
 	proftime time.Duration
 }
 
-// ExecutionSteps returns a count of abstract computation steps executed
-// by this thread. It is incremented by the interpreter. It may be used
-// as a measure of the approximate cost of Starlark execution, by
-// computing the difference in its value before and after a computation.
-//
-// The precise meaning of "step" is not specified and may change.
+// returns the current value of Steps.
```
adonovan, line 73: "ExecutionSteps returns..."

starlark/eval.go
```diff
@@ -84,12 +84,20 @@ func (thread *Thread) SetMaxExecutionSteps(max uint64) {
 	thread.maxSteps = max
 }
 
+// Uncancel reset the cancellation state.
```
adonovan, line 87: "resets"

adonovan (approved):
Two minor comment tweaks, but otherwise looks good. Thanks again!

starlark/eval.go
vanishs: fixed.

adonovan (approved):
LGTM. Thanks.

PR #440, opened 2022-11-29
lib/proto: Allow replacing repeated fields, instead of appending.

WojciechP:
Consider the following Stralark line:

        msg.repeated = [1,2,3]

Without this commit, that line appends three elements to the previous existing repeated values. With this commit, that line replaces the previous repeated values with new three items.

I believe the old semantics is simply a bug, and that most users would expect an assignment to overwrite the previous value.

### DIFF
--- a/lib/proto/proto.go
+++ b/lib/proto/proto.go
@@ -396,6 +396,7 @@ func setField(msg protoreflect.Message, fdesc protoreflect.FieldDescriptor, valu
 		defer iter.Done()
 
 		// TODO(adonovan): handle maps
+		msg.Clear(fdesc)
 		list := msg.Mutable(fdesc).List()
 		var x starlark.Value
 		for i := 0; iter.Next(&x); i++ {
--- a/lib/proto/proto.go
+++ b/lib/proto/proto.go
@@ -396,8 +396,8 @@ func setField(msg protoreflect.Message, fdesc protoreflect.FieldDescriptor, valu
 		defer iter.Done()
 
 		// TODO(adonovan): handle maps
-		msg.Clear(fdesc)
 		list := msg.Mutable(fdesc).List()
+		list.Truncate(0)
 		var x starlark.Value
 		for i := 0; iter.Next(&x); i++ {
 			v, err := toProto(fdesc, x)
--- a/starlark/testdata/proto.star
+++ b/starlark/testdata/proto.star
@@ -1,0 +1,14 @@
+# Tests of the experimental 'lib/proto' module.
+
+load("assert.star", "assert")
+load("proto.star", "proto")
+
+schema = proto.file("google/protobuf/descriptor.proto")
+
+m = schema.FileDescriptorProto(name = "somename.proto", dependency = ["a", "b", "c"])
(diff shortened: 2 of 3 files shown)


lib/proto/proto.go
```diff
@@ -396,6 +396,7 @@ func setField(msg protoreflect.Message, fdesc protoreflect.FieldDescriptor, valu
 		defer iter.Done()
 
 		// TODO(adonovan): handle maps
+		msg.Clear(fdesc)
 		list := msg.Mutable(fdesc).List()
```
adonovan, line 400: list.Truncate(0) may avoid an allocation when the existing array is sufficient.

adonovan:
Thanks for pointing out this bug, and for the fix.

I realize it would be cheeky of me to suggest that a CL like this might benefit from a test, but if you plan on investing in this package it would be worth your time to build one sooner rather than later.

adonovan (approved):

lib/proto/proto.go
WojciechP: Thanks - switched to list.Truncate(0).

WojciechP:
> I realize it would be cheeky of me to suggest that a CL like this might benefit from a test, but if you plan on investing in this package it would be worth your time to build one sooner rather than later.

Asking for tests is never cheeky :)

I added a small test for this, although there is number of choices to make, and I tried to choose reasonably:

* Where should that test be? One could argue that within lib/proto, but the simplest way was to extend the existing tests by adding another .star file in starlark/testdata, so I did that.
* Where to get a proto descriptor for the test? Running `protoc` as part of the test seemed excessive, so I decided to use `protoregistry.GlobalFiles` and link against `FileDescriptorProto` itself.

I agree that if we are to make more use of the lib/proto module, it deserves a more comprehensive test suite.
For the time being I think this is a good start to accompany the one-line fix.

starlark/testdata/proto.star
```diff
@@ -0,0 +1,14 @@
+# Tests of the experimental 'lib/proto' module.
+
+load("assert.star", "assert")
+load("proto.star", "proto")
+
+schema = proto.file("google/protobuf/descriptor.proto")
```
adonovan, line 6: Thanks. I think this is a reasonable starting point. At some point we will likely want to define our own .proto descriptors to exercise a larger range of features (and to be a little less "meta"). In that case, add the .proto file to the testdata tree, and define a go-generate command that runs protoc to update the .pb.go file from it, and commit the .pb.go as a subpackage of the test. 

Make sure the //go:generate directive is accompanied by instructions for anyone that doesn't have protoc installed on their machine, something like this:

```
/*
One-time installation:
curl --location --silent -o protoc.zip https://github.com/protocolbuffers/protobuf/releases/download/v3.19.4/protoc-3.19.4-linux-x86_64.zip 
unzip protoc.zip
go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.28.1 
*/
//go:generate protoc [-flags] test.proto
```
The google3 version of this package already contains a reasonable test suite, so I suggest you export that.

adonovan (approved):

Issue #356, opened 2021-03-04
API: add simpler CompareSameType interface


adonovan:
The current CompareSameType interface, which answers queries like "is x < y?" or "is x >= y?", arose from my misunderstanding of the requirements of typical sort routines. Go's sort.Sort accepts a `less func(x, y) bool` parameter, which misled me into thinking that it could sort values that are not totally nor even weakly ordered, such as IEEE 754 float NaN values. This is not the case: nearly all efficient sort algorithms assume the relation is at least weakly ordered. (For this reason, you can't use sort.Sort on floats; the sort.SortFloats wrapper defines a weak order over floats by mapping NaNs to the least element.)

Once this was recognized, we changed the spec so that Starlark's sole sort function, `sorted`, would work correctly on a list of floats: the Starlark spec diverges from IEEE 754 and states that floats are totally ordered, with all NaN values considered equal, and larger than infinity.

Consequently, it would be sufficient for the CompareSameType interface not to accept an operator and return a boolean, but to accept two values and return a three-valued comparison (<0, 0, >0), and this would simplify all implementations, which must currently copy the `threeway` function. We can't break the API, but we could add a new Compare interface and have the interpreter test for it first. Let's do that.

SamWheating:
Closed by https://github.com/google/starlark-go/pull/457.

(my bad, I didn't correctly reference the issue in the PR so it didn't auto-close)

Issue #292, opened 2020-07-07
clarify named parameters of int, sorted


alandonovan:
The implementation accepts int(x="1", base=10), like Python2 but not Python3.
Similarly, it accepts sorted(iterable=[]), unlike both Pythons.
All other functions that accept named arguments are clearly documented.

Issue #271, opened 2020-04-30
evaluator incorrectly allows f(*args, k=v)


alandonovan:
The spec says arguments must conform to the order positionals < keywords < *args < **kwargs, but the implementation in fact implements the weaker order required by Python2: positionals < {keywords,*args} < **kwargs. For example:

```
$ starlark
Welcome to Starlark (go.starlark.net)
>>> def f(x): print(x); return x
>>> def g(*args, **kwargs): return (args, kwargs)
>>> g(f(1), k=f(2), *[f(3)], k2=f(4), **dict(k3=f(5)))
1
2
4
3
5
((1, 3), {"k": 2, "k2": 4, "k3": 5})
>>>  
```

The reason the spec requires the stricter order is that it is easy to compile while ensuring that side effects occur in left-to-right source order. The Go implementation should reject keyword arguments following a *args argument. The Java implementation already does so.

Also, the spec wording could be improved on argument passing and the rationale for the strict order.

See https://github.com/bazelbuild/starlark/issues/13 for original discussion leading to current spec, and https://github.com/google/starlark-go/blob/master/internal/compile/compile.go#L1677 for discussion of rejected alternatives.

SamWheating:
I believe this was fixed in https://github.com/google/starlark-go/pull/317

Issue #270, opened 2020-04-23
specify equivalence relation for functions


alandonovan:
The execution of each def statement creates a new function that is equal only to itself. We should state that.

PR #441, opened 2022-12-04
lib/time: consider any Time other than zero instant to be truthy

SamWheating:
I know that the truthiness of timestamps isn't well defined, but the current behaviour is the opposite of what I would expect:

```
Welcome to Starlark (go.starlark.net)
>>> bool(time.now())
False
>>> bool(time.parse_time("0001-01-01T00:00:00Z"))
True
```

This change inverts the existing logic so that any time other than `0001-01-01T00:00:00Z` will evaluate to True.

Alternatively, I think that considering all times to be `True` would also be acceptable. Thoughts?

### DIFF
--- a/lib/time/time.go
+++ b/lib/time/time.go
@@ -348,7 +348,7 @@ func (t Time) Hash() (uint32, error) {
 
 // Truth returns the truth value of an object required by starlark.Value
 // interface.
-func (t Time) Truth() starlark.Bool { return starlark.Bool(time.Time(t).IsZero()) }
+func (t Time) Truth() starlark.Bool { return !starlark.Bool(time.Time(t).IsZero()) }
 
 // Attr gets a value for a string attribute, implementing dot expression support
 // in starklark. required by starlark.HasAttrs interface.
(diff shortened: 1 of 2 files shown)


lib/time/time.go
```diff
@@ -348,7 +348,7 @@ func (t Time) Hash() (uint32, error) {
 
 // Truth returns the truth value of an object required by starlark.Value
 // interface.
-func (t Time) Truth() starlark.Bool { return starlark.Bool(time.Time(t).IsZero()) }
```
adonovan, line 351: This was clearly just a mistake that no-one noticed until now. Thanks for the fix!

adonovan (approved):

lib/time/time.go
SamWheating: No problem!

I'm pretty interested in this project - do you mind if I pick up a few of the TODOs? Got any new features or library modules you've been thinking about?

adonovan: > I'm pretty interested in this project - do you mind if I pick up a few of the TODOs? Got any new features or library modules you've been thinking about?

Please do! I don't have any specific new features in mind. Much of the effort of this project has been spent in trying to harmonize the two (now three) implementations around a common spec, and to be very judicious about new features. Feel free to have a crack at any TODO comment in the code, and don't hesitate to ask for clarification before spending significant effort. These issues too might be suitable tasks, depending on your tastes:

https://github.com/google/starlark-go/issues/356
https://github.com/google/starlark-go/issues/249
https://github.com/google/starlark-go/issues/270
https://github.com/google/starlark-go/issues/271
https://github.com/google/starlark-go/issues/292

Many thanks for the kind offer.

PR #442, opened 2022-12-05
int_test: Move `--entry` flag definition from TestMain to package init.

WojciechP:
This change matters only in google3/blaze.

Within google3/blaze, all Go tests have additional initialization logic added by the build system. That causes the flags to be parsed before TestMain runs, resulting in a "flag provided but not defined" error when the test spawns itself from a subshell. Moving the flag definition to package-level variable is enough to fix the initialization order.

### DIFF
--- a/starlark/int_test.go
+++ b/starlark/int_test.go
@@ -140,16 +140,18 @@ func intfallback() {
 }
 
 // The --entry flag invokes an alternate entry point, for use in subprocess tests.
+var testEntry = flag.String("entry", "", "child process entry-point")
+
 func TestMain(m *testing.M) {
-	var entry string
-	flag.StringVar(&entry, "entry", "", "child process entry-point")
+	// In some build systems, notably Blaze, flag.Parse is called before TestMain,
+	// in violation of the TestMain contract, making this second call a no-op.
 	flag.Parse()
-	switch entry {
+	switch *testEntry {
 	case "":
 		os.Exit(m.Run()) // normal case
 	case "intfallback":
 		intfallback()
 	default:
-		log.Fatalf("unknown entry point: %s", entry)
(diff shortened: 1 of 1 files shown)


adonovan:
> Within google3/blaze, all Go tests have additional initialization logic added by the build system. That causes the flags to be parsed before TestMain runs, resulting in a "flag provided but not defined" error when the test spawns itself from a subshell. Moving the flag definition to package-level variable is enough to fix the initialization order.

I think the bug may be in google3. The [documentation for the testing package](https://pkg.go.dev/testing#hdr-Main) is very clear:

> When TestMain is called, flag.Parse has not been run. If TestMain depends on command-line flags, including those of the testing package, it should call flag.Parse explicitly.

Let's file an internal Blaze+Go bug and see where that process leads.

starlark/int_test.go
```diff
@@ -140,16 +140,16 @@ func intfallback() {
 }
 
 // The --entry flag invokes an alternate entry point, for use in subprocess tests.
+var testEntry = flag.String("entry", "", "child process entry-point")
+
 func TestMain(m *testing.M) {
-	var entry string
-	flag.StringVar(&entry, "entry", "", "child process entry-point")
 	flag.Parse()
```
adonovan, line 146: Let's add a comment:

// In some build systems, notably Blaze), flag.Parse is called before TestMain,
// in violation of the TestMain contract, making this second call a no-op.

adonovan:
> Let's file an internal Blaze+Go bug and see where that process leads.

I don't think the Blaze behavior will be easily fixed, so let's proceed with this.

starlark/int_test.go
WojciechP: Added.

adonovan (approved):

PR #443, opened 2022-12-08
Made JSON encoder a public function

eastside:
This one is pretty simple. I found myself starting to write a bunch of code to encode a Starlark value as JSON, when I realized that oh hey, there's a JSON encoder already in the project. It's just hidden.

### DIFF
--- a/lib/json/json.go
+++ b/lib/json/json.go
@@ -27,29 +27,30 @@ import (
 
 // Module json is a Starlark module of JSON-related functions.
 //
-//   json = module(
-//      encode,
-//      decode,
-//      indent,
-//   )
+//	json = module(
+//	   encode,
+//	   decode,
+//	   indent,
+//	)
 //
 // def encode(x):
 //
 // The encode function accepts one required positional argument,
 // which it converts to JSON by cases:
-// - A Starlark value that implements Go's standard json.Marshal
-//   interface defines its own JSON encoding.
-// - None, True, and False are converted to null, true, and false, respectively.
-// - Starlark int values, no matter how large, are encoded as decimal integers.
-//   Some decoders may not be able to decode very large integers.
-// - Starlark float values are encoded using decimal point notation,
-//   even if the value is an integer.
-//   It is an error to encode a non-finite floating-point value.
-// - Starlark strings are encoded as JSON strings, using UTF-16 escapes.
-// - a Starlark IterableMapping (e.g. dict) is encoded as a JSON object.
-//   It is an error if any key is not a string.
-// - any other Starlark Iterable (e.g. list, tuple) is encoded as a JSON array.
-// - a Starlark HasAttrs (e.g. struct) is encoded as a JSON object.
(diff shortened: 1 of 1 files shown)


lib/json/json.go
```diff
@@ -81,12 +82,8 @@ var Module = &starlarkstruct.Module{
...
 }
 
-func encode(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
-	var x starlark.Value
-	if err := starlark.UnpackPositionalArgs(b.Name(), args, kwargs, 1, &x); err != nil {
-		return nil, err
-	}
-
+// Encodes the given Starlark value as a string in JSON format.
+func EncodeJSON(x starlark.Value) (string, error) {
```
adonovan, line 86: Let's call this Encode (since the package name is json) and make the result type a []byte.

lib/json/json.go
```diff
@@ -81,12 +82,8 @@ var Module = &starlarkstruct.Module{
 	},
 }
 
-func encode(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
-	var x starlark.Value
-	if err := starlark.UnpackPositionalArgs(b.Name(), args, kwargs, 1, &x); err != nil {
-		return nil, err
-	}
-
+// Encodes the given Starlark value as a string in JSON format.
```
adonovan, line 85: // Encode returns the JSON encoding of a Starlark value.

adonovan, line 85: And for symmetry, let's have a Decode function.

lib/json/json.go
```diff
@@ -217,21 +217,21 @@ func EncodeJSON(x starlark.Value) (string, error) {
...
 	if err := starlark.UnpackPositionalArgs(b.Name(), args, kwargs, 1, &x); err != nil {
 		return nil, err
 	}
-	s, err := EncodeJSON(x)
+	bytes, err := Encode(x)
 	if err != nil {
 		return nil, fmt.Errorf("%s: %v", b.Name(), err)
 	}
-	return starlark.String(s), nil
+	return starlark.String(string(bytes[:])), nil
```
adonovan, line 234: String(bytes)

eastside:
Two questions:

First, I wasn't sure if Decode should also accept []byte, to have some symmetry with Encode.

Second, I wasn't sure if I should add some tests. I started to add some, but realized I was testing basically everything that's already being tested.

adonovan:
> First, I wasn't sure if Decode should also accept []byte, to have some symmetry with Encode.

Yes, I think it should.

> Second, I wasn't sure if I should add some tests. I started to add some, but realized I was testing basically everything that's already being tested.

You could add one test to ensure that a value can be encoded and then decoded to ensure there are no gross errors, but I wouldn't bother testing all the cases of JSON.

PR #444, opened 2023-01-03
starlark: do not use reflect.Value.MethodByName

aarzilli:
Using MethodByName disables deadcode elimination in the linker (see the
deadcode function in $GOROOT/src/cmd/link/internal/ld/deadcode.go).

Since the only use of MethodByName is with a static string as an
argument it can be easily replaced with a downcast to an interface.


adonovan (approved):
Thanks, I don't know why I wrote it that way in the first place.

Issue #450, opened 2023-01-17
Expose Function defaults.


vtsao:
Would it be possible to expose [Function defaults](https://sourcegraph.com/github.com/google/starlark-go@master/-/blob/starlark/value.go?L662)?

I'm working on an integration where I need to get the default values before executing the function. Exposing it may be as simple as adding a `func (fn *Function) Defaults() Tuple` function. Not sure if defaults weren't exposed on purpose?

adonovan:
That tuple is not a valid Starlark value, as it may contain elements of type [mandatory](https://sourcegraph.com/github.com/google/starlark-go@master/-/blob/starlark/interp.go?L679-681), which are not valid values. Even if that were not the case, it seems like the representation should be encapsulated to permit future evolution.

Why do you need to access the defaults tuple? Isn't that a violation of the function abstraction?

vtsao:
> Why do you need to access the defaults tuple?

I'm integrating Starlark functions with a command line tool we own and I want to get the types of the function parameters so I can automatically generate flags for the command. When registering flags, I need to know the type of the flag, like is it a number, list, map, etc. So I was hoping to use param defaults for that.

adonovan:
Ok, let's add a `func (fn *Function) ParamDefault(i int) Value` method after [Param](https://sourcegraph.com/github.com/google/starlark-go@fae38c8a6d89dc410be86b76dfff475b29dba878/-/blob/starlark/value.go?L703-715), that returns nil for mandatory parameters and panics unless 0 <= i <= NumParams.

vtsao:
Thanks, I'll create a PR.

PR #452, opened 2023-01-17
Added a ParamDefault() method to Function that returns parameter default values.

vtsao:
#450

### DIFF
--- a/starlark/value.go
+++ b/starlark/value.go
@@ -698,7 +698,6 @@ func (fn *Function) Globals() StringDict { return fn.module.makeGlobalDict() }
 
 func (fn *Function) Position() syntax.Position { return fn.funcode.Pos }
 func (fn *Function) NumParams() int            { return fn.funcode.NumParams }
-func (fn *Function) NumParamDefaults() int     { return len(fn.defaults) }
 func (fn *Function) NumKwonlyParams() int      { return fn.funcode.NumKwonlyParams }
 
 // Param returns the name and position of the ith parameter,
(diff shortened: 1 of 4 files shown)


vtsao:
@adonovan not sure where the best place is to add a test for this? It doesn't look like any of the other `starlark.Function` methods have tests.

starlark/value.go
```diff
@@ -698,6 +698,7 @@ func (fn *Function) Globals() StringDict { return fn.module.makeGlobalDict() }
 
 func (fn *Function) Position() syntax.Position { return fn.funcode.Pos }
 func (fn *Function) NumParams() int            { return fn.funcode.NumParams }
+func (fn *Function) NumParamDefaults() int     { return len(fn.defaults) }
```
adonovan, line 701: This isn't necessary. Let's keep it simple: the caller can call ParamDefault for each of the NumParams.

starlark/value.go
```diff
@@ -711,6 +712,21 @@ func (fn *Function) Param(i int) (string, syntax.Position) {
 	id := fn.funcode.Locals[i]
 	return id.Name, id.Pos
 }
+
+// ParamDefault returns the default value the ith optional parameter, where
```
adonovan, line 716: Let's say: "ParamDefault returns the default value of the specified parameter (0 <= i < NumParams()), or nil if the parameter is not optional."

(and update the implementation to correspond)

adonovan:
> @adonovan not sure where the best place is to add a test for this? It doesn't look like any of the other `starlark.Function` methods have tests.

Yeah, there isn't much there yet. See the TODO in testdata/function.star:
```
# TODO(adonovan):
# - add some introspection functions for looking at function values
#   and test that functions have correct position, free vars, names of locals, etc.
```

If you're feeling diligent, you could add a built-in function following the paradigm of hasFields or struct that returns a dict of function attributes (e.g. name, pos, defaults) and then write assertions in Starlark. But if you don't feel like it that's fine too.

starlark/value.go
vtsao: The length of `NumParams()` and `fn.defaults` may be different though right? `fn.defaults` only has values for parameters that have a default (and keyword-only parameters).

adonovan: Exactly, the representation omits the longest prefix that would otherwise contain all "mandatory" elements. But we shouldn't expose that in the API.

starlark/value.go
vtsao: Done.

starlark/value.go
vtsao: Gotcha, done.

vtsao:
> If you're feeling diligent, you could add a built-in function following the paradigm of hasFields or struct that returns a dict of function attributes (e.g. name, pos, defaults) and then write assertions in Starlark. But if you don't feel like it that's fine too.

I added a test in `value_test.go` for this that I think is more direct as it tests `ParamDefault()` directly. Hope this is okay, even though it's a bit different than the other tests.

vtsao:
@adonovan friendly ping - how does this all look?

vtsao:
Hey @adonovan, following up again - any chance we can get this merged? Let me know if you'd like to see anything else in this PR. Thanks!

adonovan (approved):
Sorry for the delay. Change looks good. Many thanks!

Issue #455, opened 2023-01-21
starlark: TestProfile allocates too much


adonovan:
The test of the profiler generates the fibonacci sequence, which rapidly allocates integers that are over 20,000 digits long, a total of about 2GB of actual RAM. If the POSIX int64 optimization is enabled, and the virtual address space has been artificially limited to only slightly more than then 4GB consumed by the optimization, then there isn't enough virtual address space left for actual RAM, so allocation fails.

We should change the test to allocate less.

PR #456, opened 2023-01-22
starlark: optimize TestProfile

adonovan:
Previously, the test would compute the first 100,000 elements of the Fibonacci sequence, holding them all in an array. Since elements may have over 20,000 digits, this allocates a lot of memory, which, under tight ulimits of virtual address space, in conjunction with the mmap int optimization, may cause the process to have insufficient virtual address space to allocate the necessary memory.

This change causes it to compute but not retain the elements of the sequence.

Fixes google/starlark-go#455

### DIFF
--- a/starlark/profile_test.go
+++ b/starlark/profile_test.go
@@ -31,10 +31,10 @@ func TestProfile(t *testing.T) {
 
 	const src = `
 def fibonacci(n):
-	res = list(range(n))
-	for i in res[2:]:
-		res[i] = res[i-2] + res[i-1]
-	return res
+	x, y = 1, 1
+	for i in range(n):
+		x, y = y, x+y
+	return y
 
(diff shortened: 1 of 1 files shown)


starlark/profile_test.go
```diff
@@ -76,7 +77,7 @@ fibonacci(100000)
 			t.Errorf("output did not contain %q", want)
 		}
 	}
-	if t.Failed() {
+	if true || t.Failed() {
```
brandjon, line 80: Leftover from debugging?

brandjon (approved):

starlark/profile_test.go
adonovan: Indeed, thanks. But it's close to midnight on a Saturday, when all right-thinking people are out dancing. Step away your from keyboard, man.

PR #457, opened 2023-02-27
Adding new ThreeWayComparable interface

SamWheating:
Re: https://github.com/google/starlark-go/issues/356

This adds a new `TotallyOrdered` interface, similar to `Comparable` which returns an integer representing the <, ==, > relation between values. The `CompareDepth` function has been to updated check for this interface and handle this return style if applicable.

Also refactors `int`, `float`, `time.Time` and `time.Duration` types to use the new interface.

The previous `Comparable` interface will have to be maintained, not just for API compatibility but also for types which define equality/inequality but not ordering (for example, [structs](https://github.com/google/starlark-go/blob/c52844e64a10fb0a67ae99619ee98a49de4f8210/starlarkstruct/struct.go#L247))

#### Open questions:

 - Is the ThreeWayCompareSameType name too verbose? I was thinking about overloading the name CompareSameType but that felt like it would get confusing. -> Yes, change to `TotallyOrdered`
 - Should I also update the other orderable types (float, str, list, etc) to use this interface in this PR? -> Yes.

### DIFF
--- a/lib/time/time.go
+++ b/lib/time/time.go
@@ -212,13 +212,12 @@ func (d Duration) AttrNames() []string {
 // CompareSameType implements comparison of two Duration values. required by
 // starlark.Comparable interface.
 func (d Duration) CompareSameType(op syntax.Token, v starlark.Value, depth int) (bool, error) {
-	cmp := 0
 	if x, y := d, v.(Duration); x < y {
-		cmp = -1
+		return -1, nil
 	} else if x > y {
-		cmp = 1
+		return 1, nil
 	}
-	return threeway(op, cmp), nil
+	return 0, nil
 }
 
 // Binary implements binary operators, which satisfies the starlark.HasBinary
@@ -394,16 +393,15 @@ func (t Time) AttrNames() []string {
 
 // CompareSameType implements comparison of two Time values. required by
 // starlark.Comparable interface.
-func (t Time) CompareSameType(op syntax.Token, yV starlark.Value, depth int) (bool, error) {
+func (t Time) CompareSameType(yV starlark.Value, depth int) (int, error) {
 	x := time.Time(t)
 	y := time.Time(yV.(Time))
-	cmp := 0
 	if x.Before(y) {
-		cmp = -1
+		return -1, nil
 	} else if x.After(y) {
-		cmp = 1
+		return 1, nil
 	}
-	return threeway(op, cmp), nil
+	return 0, nil
 }
 
 // Binary implements binary operators, which satisfies the starlark.HasBinary
@@ -484,24 +482,4 @@ func builtinAttrNames(methods map[string]builtinMethod) []string {
 	}
 	sort.Strings(names)
 	return names
-}
-
-// Threeway interprets a three-way comparison value cmp (-1, 0, +1)
-// as a boolean comparison (e.g. x < y).
-func threeway(op syntax.Token, cmp int) bool {
-	switch op {
-	case syntax.EQL:
-		return cmp == 0
-	case syntax.NEQ:
-		return cmp != 0
-	case syntax.LE:
-		return cmp <= 0
-	case syntax.LT:
-		return cmp < 0
-	case syntax.GE:
-		return cmp >= 0
-	case syntax.GT:
-		return cmp > 0
-	}
-	panic(op)
 }
--- a/lib/time/time.go
+++ b/lib/time/time.go
@@ -209,8 +209,8 @@ func (d Duration) AttrNames() []string {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


lib/time/time.go
```diff
@@ -211,14 +211,13 @@ func (d Duration) AttrNames() []string {
 
 // CompareSameType implements comparison of two Duration values. required by
 // starlark.Comparable interface.
-func (d Duration) CompareSameType(op syntax.Token, v starlark.Value, depth int) (bool, error) {
-	cmp := 0
+func (d Duration) ThreeWayCompareSameType(v starlark.Value, depth int) (int, error) {
 	if x, y := d, v.(Duration); x < y {
-		cmp = -1
+		return -1, nil
```
adonovan, line 216: For all integers you can use this trick:
        
{ return x > y - x < y, nil } // see Hacker's Delight §2.9

starlark/value.go
```diff
@@ -131,15 +131,36 @@ type Comparable interface {
 	CompareSameType(op syntax.Token, y Value, depth int) (bool, error)
 }
 
+// Alternate, simplified comparison interface
+type ThreeWayComparable interface {
```
adonovan, line 135: What do think of naming this TotallyOrdered.Cmp after the Python three-valued comparison function, cmp?

Let's document it as follows:

// A TotallyOrder is a type whose values form a total order:
// if x and y are of the same TotallyOrdered type, then x must be less than y,
// greater than y, or equal to y.
//
// It is simpler than Comparable and should be preferred in new code,
// but if a type implements both interfaces, Comparable takes precedence.
type ...

starlark/value.go
```diff
@@ -131,15 +131,36 @@ type Comparable interface {
...
+	Value
+	// ThreeWayCompareSameType compares one value to another of the same orderable
```
adonovan, line 137: // Cmp compares two values x and y of the same totally ordered type.
// It returns negative if x < y, positive if x > y, and zero if the values are equal.

(I don't think we should specify that the values are exactly +/- 1.)

starlark/value.go
```diff
@@ -131,15 +131,36 @@ type Comparable interface {
...
 var (
-	_ Comparable = Int{}
-	_ Comparable = False
-	_ Comparable = Float(0)
-	_ Comparable = String("")
-	_ Comparable = (*Dict)(nil)
-	_ Comparable = (*List)(nil)
-	_ Comparable = Tuple(nil)
-	_ Comparable = (*Set)(nil)
+	_ ThreeWayComparable = Int{}
```
adonovan, line 156: Floats are totally ordered too.

(So are Strings but the implementation of strings.Compare is obtusely inefficient.)

adonovan (commented):
Thanks for this contribution!

lib/time/time.go
SamWheating: Hmmm, I can see how this would work in some languages but go isn't as generous with the boolean arithmetic.

In python:
```python
>>> x, y = 1, 2
>>> (x > y) - (x < y)
-1
```
 vs. Go:
 ```go
x, y := 1, 2
(x > y) - (x < y)
	
`operator - not defined on (x > y) (untyped bool value)`
 ```
 
 Any suggestions for how this should be done in go? We could do some weird mapping / casting of bool to int but I think that overall the current if/else implementation would be cleaner.

lib/time/time.go
```diff
@@ -393,7 +393,7 @@ func (t Time) AttrNames() []string {
 
 // CompareSameType implements comparison of two Time values. required by
```
adonovan, line 394: Cmp
Required
TotallyOrdered

starlark/int.go
```diff
@@ -191,7 +191,7 @@ func (i Int) Hash() (uint32, error) {
 	return 12582917 * uint32(lo+3), nil
 }
 
-func (x Int) ThreeWayCompareSameType(v Value, depth int) (int, error) {
+func (x Int) Cmp(v Value, depth int) (int, error) {
 	y := v.(Int)
 	xSmall, xBig := x.get()
 	ySmall, yBig := y.get()
```
adonovan, line 197: Could you also add this comment to L201? Thanks.

return signum64(xSmall - ySmall), nil // safe: int32 operands

(It's not safe to use signum(x-y) in this way in general.)

adonovan (commented):
A few more minor comments.

lib/time/time.go
adonovan: > Hmmm, I can see how this would work in some languages but go isn't as generous with the boolean arithmetic.

Quite right; never mind.

adonovan (approved):
Looks good. Many thanks again!

PR #458, opened 2023-02-27
Disambiguate a resolve message

filmil:
I had a duplicate kwargs named 'disabled'. This
ended up with a message 'keyword argument disabled repeated', which is really difficult to parse.

Hope to clarify things by changing that text a bit.

### DIFF
--- a/resolve/resolve.go
+++ b/resolve/resolve.go
@@ -746,7 +746,7 @@ func (r *resolver) expr(e syntax.Expr) {
 				}
 				x := binop.X.(*syntax.Ident)
 				if seenName[x.Name] {
-					r.errorf(x.NamePos, "keyword argument %s repeated", x.Name)
+					r.errorf(x.NamePos, "keyword argument %q is repeated", x.Name)
 				} else {
 					if seenName == nil {
 						seenName = make(map[string]bool)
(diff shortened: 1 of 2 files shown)


filmil:
I would ask to keep as is.

This messaging helps in cases where different people added the same attribute in close proximity in time. It can be confusing if you see 'repeated', but you know you haven't repeated anything.

I just had a brief run-around because of this, and hence this change. The current wording could have helped.

adonovan:
> This messaging helps in cases where different people added the same attribute in close proximity in time. It can be confusing if you see 'repeated', but you know you haven't repeated anything.

Like all compiler errors, this is a static check of the code at a particular state. I don't really see how time (either execution time or codebase evolution) enters the picture. If a user sees this error, it means the string `a=` appears twice (i.e. is repeated) in the same function call. The error seems pretty clear to me.

filmil:
> Like all compiler errors, this is a static check of the code at a particular state. I don't really see how time (either execution time or codebase evolution) enters the picture. [...] 

I understand. Happy to change to what is more acceptable to you. Upcoming.

> The error seems pretty clear to me.

I understand that too.

I was noting down a user's perspective, where the message proved to be confusing. Again, happy to shorten the message.

filmil:
Is done now. PTAL.

adonovan (approved):
Thanks again!

PR #466, opened 2023-04-28
Allow json.decode to return a default value if decoding fails

### LINKED ISSUE #465: FR: json.valid
As Starlark doesn't include `try` and `except`, it would be valuable to have a `json.valid` or similar function to enable custom handling (overwrite, `fail`, etc) of invalid json encoding.

tetromino:
This is the starlark-go counterpart to bazelbuild/bazel@71fb1e4

It is useful to be able to decode a potentially invalid JSON string without
failing Starlark evaluation. Typically, in this case, we would want to return
a default value - either a reasonable safe default or some domain-specific
representation of "we encountered an error".

Note that the original behavior of json.decode failing on invalid JSON input
still remains (for lack of a better word) the default: the `default` value,
even if it is `None`, must be specified explicitly.

Fixes #465

### DIFF
--- a/lib/json/json.go
+++ b/lib/json/json.go
@@ -290,6 +290,11 @@ func decode(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, k
 	if err := starlark.UnpackArgs(b.Name(), args, kwargs, "x", &s, "default?", &d); err != nil {
 		return nil, err
 	}
+	if len(args) < 1 {
+		// "x" parameter is positional only; UnpackArgs does not allow us to
+		// directly express def(x, *, default)
+		return nil, fmt.Errorf("%s: unexpected keyword argument x", b.Name())
+	}
 
 	// The decoder necessarily makes certain representation choices
 	// such as list vs tuple, struct vs dict, int vs float.
--- a/lib/json/json.go
+++ b/lib/json/json.go
@@ -502,7 +502,7 @@ func decode(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, k
 		switch x := x.(type) {
 		case failure:
 			if d != nil {
-			    v = d
+				v = d
 			} else {
 				err = fmt.Errorf("json.decode: at offset %d, %s", i, x)
 			}
--- a/lib/json/json.go
+++ b/lib/json/json.go
@@ -62,11 +62,9 @@ import (
 //   contain a decimal point.
 // - JSON objects are parsed as new unfrozen Starlark dicts.
 // - JSON arrays are parsed as new unfrozen Starlark lists.
-// If `x` is not a valid JSON string and the optional, positional-only `default`
-// parameter is specified (including specified as `None`), the decode function
-// returns the `default` value.
-// If `x` is not a valid JSON string and the optional, positional-only `default`
-// parameter is not specified, the decode function fails.
+// If x is not a valid JSON string, the behavior depends on the "default"
+// positional-only parameter: if present, Decode returns its value; otherwise,
+// Decode fails.
 //
 // def indent(str, *, prefix="", indent="\t"):
 //
--- a/lib/json/json.go
+++ b/lib/json/json.go
@@ -63,8 +63,7 @@ import (
 // - JSON objects are parsed as new unfrozen Starlark dicts.
 // - JSON arrays are parsed as new unfrozen Starlark lists.
 // If x is not a valid JSON string, the behavior depends on the "default"
-// positional-only parameter: if present, Decode returns its value; otherwise,
-// Decode fails.
+// parameter: if present, Decode returns its value; otherwise, Decode fails.
 //
 // def indent(str, *, prefix="", indent="\t"):
 //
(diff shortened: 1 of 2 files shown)


lib/json/json.go
```diff
@@ -54,15 +54,19 @@ import (
...
 //
-// The decode function accepts one positional parameter, a JSON string.
+// The decode function has one required positional parameter, a JSON string.
 // It returns the Starlark value that the string denotes.
 // - Numbers are parsed as int or float, depending on whether they
 //   contain a decimal point.
 // - JSON objects are parsed as new unfrozen Starlark dicts.
 // - JSON arrays are parsed as new unfrozen Starlark lists.
-// Decoding fails if x is not a valid JSON string.
+// If `x` is not a valid JSON string and the optional, positional-only `default`
```
adonovan, line 65: More concisely:
```
// If x is not a valid JSON string the behavior depends
// on the "default" parameter: if present, Decode
// returns its value; otherwise Decode fails.
```

Don't use backquotes in Go doc comments. If the result would be ambiguous (as with "the default parameter") then use double quotes around the name.

lib/json/json.go
```diff
@@ -54,15 +54,19 @@ import (
...
+// parameter is specified (including specified as `None`), the decode function
+// returns the `default` value.
+// If `x` is not a valid JSON string and the optional, positional-only `default`
```
adonovan, line 68: Why is it positional only? Keyword-only would be clearer and more extensible.

If the answer is: for consistency with dict.pop, I'm not sure that's a compelling reason. That function is tiny and very familiar, any needs to avoid a dict allocation. None of that is true here.

lib/json/json.go
```diff
@@ -496,18 +501,22 @@ func decode(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, k
 		x := recover()
 		switch x := x.(type) {
 		case failure:
-			err = fmt.Errorf("json.decode: at offset %d, %s", i, x)
+			if d != nil {
```
adonovan, line 504: One thing to be wary of in this sentinel-based error reporting is the potential for inadvertently creating an exception handler in the language. Imagine a variant of decode that under some circumstance called a user-supplied function. By calling `decode(fn, default=123)`, the user could call fn, handle any error that it raises, and replace it with a sentinel value. This could potentially be used to continue after a function that leaves the thread in a bad state, or to continue and perhaps retry after some error (such as a one-shot equivalent of the limit on computation steps).

I don't see any problem with this logic as it stands, but if we use this approach more widely we will need to think carefully about each case.

lib/json/json.go
tetromino: brandjon@ convinced me to do it positional-only in the Java Starlark version (which we already committed) for consistency with getattr, dict.get, and dict.pop. Arguably, consistency is as important as clarity.

As for extensibility, I was worried about that too until I realized that having an optional positional-only parameter does not preclude us from adding keyword-only parameters later if needed.

lib/json/json.go
tetromino: Acknowledged - but in this case, we are matching that the failure type is `failure` which is declared locally and not exported, so any other source of panic would fall through to

```
		default:
			panic(x) // unexpected panic
```

I will add a comment to make this more clear.

lib/json/json.go
adonovan: The Java version could be quickly updated if we decide to change things here. I'll let you and Jon decide.

brandjon: Discussed with @tetromino offline. I'm pretty ambivalent. Consistency with `dict.get` and `getattr` is nice. On the other hand there may be other library APIs that need a default and for which a keyword is more clearly appropriate, so we may be sacrificing consistency with those. So I'll defer to @tetromino on this.

> having an optional positional-only parameter does not preclude us from adding keyword-only parameters later if needed

If you really mean keyword-*only*, we'd have to break code relying on the positional convention at some point.

tetromino: I would resolve the bike-shed color question by making the `default` param keyword or positional. (And will make the corresponding change in Java.)

tetromino:
@adonovan - as side effect, this makes the `x` argument of json.decode positional or keyword. Currently `starlark.UnpackArgs` doesn't allow marking args as keyword-only. Would you want me to fix that in a followup, or before merging this PR, or it's a nit that does not matter?

adonovan:
> as side effect, this makes the x argument of json.decode positional or keyword. Currently starlark.UnpackArgs doesn't allow marking args as keyword-only. Would you want me to fix that in a followup, or before merging this PR, or it's a nit that does not matter?

I think it's important to be consistent across implementations, and consistent with spec, as many bugs in the past have resulted from differences this small. I don't have a strong feeling about which semantics you choose. 

BTW you can use define an all-parameters-keyword-only function using UnpackArgs by first rejecting the len(args) > 0 case.

tetromino:
> BTW you can use define an all-parameters-keyword-only function using UnpackArgs by first rejecting the len(args) > 0 case.

Good point. Done.

PR #468, opened 2023-05-30
lib/json: allow encoding dicts with int keys

andreimatei:
Before this patch, only dicts with string keys were supported. Now, integer keys are also permitted. Dicts with int keys should be common. This matches Go's json.Marshal() and Python's json.dumps(), which also permit int keys (in addition to other things).

### DIFF
--- a/lib/json/json.go
+++ b/lib/json/json.go
@@ -155,19 +155,45 @@ func encode(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, k
 			// e.g. dict (must have string keys)
 			buf.WriteByte('{')
 			items := x.Items()
+			stringKeys := false
+			intKeys := false
 			for _, item := range items {
-				if _, ok := item[0].(starlark.String); !ok {
-					return fmt.Errorf("%s has %s key, want string", x.Type(), item[0].Type())
+				_, isString := item[0].(starlark.String)
+				_, isInt := item[0].(starlark.Int)
+				if !isString && !isInt {
+					return fmt.Errorf("%s has %s key, want string or int", x.Type(), item[0].Type())
 				}
+				stringKeys = stringKeys || isString
+				intKeys = intKeys || isInt
 			}
-			sort.Slice(items, func(i, j int) bool {
-				return items[i][0].(starlark.String) < items[j][0].(starlark.String)
-			})
+			// Sort the keys. This is useful if they come from a dict, which presents
+			// its items unordered. However, we only sort if all the keys have the
+			// same type.
+			if !(stringKeys && intKeys) {
+				sort.Slice(items, func(i, j int) bool {
+					// Compare as strings if all the keys are strings, or as ints if all
+					// the keys are ints.
+					if stringKeys {
+						return items[i][0].(starlark.String) < items[j][0].(starlark.String)
+					}
+					cmp, err := items[i][0].(starlark.Int).Cmp(items[j][0], 0 /* depth */)
+					if err != nil {
+						panic(fmt.Errorf("unexpected failure to compare ints: %w", err))
+					}
+					return cmp == -1
+				})
+			}
 			for i, item := range items {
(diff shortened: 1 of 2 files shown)


lib/json/json.go
```diff
@@ -155,19 +155,45 @@ func encode(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, k
...
-					return fmt.Errorf("%s has %s key, want string", x.Type(), item[0].Type())
+				_, isString := item[0].(starlark.String)
+				_, isInt := item[0].(starlark.Int)
+				if !isString && !isInt {
+					return fmt.Errorf("%s has %s key, want string or int", x.Type(), item[0].Type())
 				}
+				stringKeys = stringKeys || isString
+				intKeys = intKeys || isInt
+			}
+			// Sort the keys. This is useful if they come from a dict, which presents
```
andreimatei, line 169: I was wondering what this key sorting business is about, and whether it should be pushed into the different implementations of the `IterableMapping` that want it?
I'm thinking that by default a random impl does *not* want it (i.e. it wants the keys rendered in the order it produces them in `Items()`), but notable `starlark.Dict` probably wants it (because I think it currently produces its items in a random order).

adonovan (commented):
Thanks for this contribution, Andrei.

lib/json/json.go
andreimatei: Makes sense about the key sorting. I wasn't particularly interested in doing anything about it.
As you can see, the patch kept the sorting, although not for the case of mixed int/string keys. I'm happy to do whatever in that case - sort as strings, or even reject the whole marshalling.

> Supporting int keys is also a behavior change that needs discussion.

So you're saying I should necessarily discuss the int keys support on https://github.com/bazelbuild/starlark/ for this patch to move forward (i.e. "extras" in the Go impl are not kosher)? In the happy scenario that the proposal gets accepted, would a Java implementation also be required? Because that might be beyond personal interest :)

adonovan: >  I should necessarily discuss the int keys support on https://github.com/bazelbuild/starlark/ for this patch to move forward...?

Yes, the two implementations should be consistent in the behavior of all their core functions. (This is not a theoretical concern: at Google we have tools in Go and Java that process Bazel/Blaze BUILD files, and any divergence in their behaviors is problematic.)

> would a Java implementation also be required?

No. Someone else will implement that if the proposal is accepted.

andreimatei: I'm happy to discuss the `int` keys as a wider Starlark proposal, assuming it has your support in principle.
But just to check - I've found this recent comment on the
[starlark](https://github.com/bazelbuild/starlark) repo that seems to say that the JSON library is not under their purview -
https://github.com/bazelbuild/starlark/issues/253#issuecomment-1568684017
Should I not read too much into it?

adonovan: > I'm happy to discuss the int keys as a wider Starlark proposal, assuming it has your support in principle.

Personally I don't see much value in json.encode supporting dicts with int keys, since you can't round-trip encode+decode them: since JSON doesn't support objects with int keys, they end up being coerced to strings. And while it's true the Go's json.Marshal supports this behavior (again by coercing to strings), I've never seen anyone use it, and I can't really imagine why you'd want to. Plus it complicates the sorting. But feel free to propose it if you like.


> Should I not read too much into it?

Yep, don't worry about that comment.

andreimatei: If you don't care about it, I probably will not propose anything; I'm not blocked on this.
But,

> since you can't round-trip encode+decode

They do round-trip, though, at least in Go:
https://go.dev/play/p/v07cIJJNZLA
Let me know if this changes your interest; if it doesn't, I'll close this PR.

andreimatei:
Closing this for now as it does not look like an immediate accept. Might reopen later in the context of a broader Starlark proposal.

PR #367, opened 2021-04-08
refactor(json): move go.starlark.net/starlarkjson -> go.starlark.net/lib/json

b5:
and keep an alias package in place for backwards compatibility

PR #486, opened 2023-07-17
Fix typo in import comment of json.go

nozzy123nozzy:
Hi,
I experienced a problem in which the go-list command couldn't find go.starlark.net/stalarkjson when making the Debian package of starlark-go.

The root cause is the typo in the [import comments](https://pkg.go.dev/cmd/go#hdr-Import_path_checking) of starlarkjson/json.go file caused by #367. This typo makes the go-list command under GO111MODULE=off to give up finding go files of starlark-go. As a result, the packaging in Debian failed.

Would you please accept the PR to fix that typo?

### DIFF
--- a/starlarkjson/json.go
+++ b/starlarkjson/json.go
@@ -6,7 +6,7 @@
 // backwards compatibility
 //
 // Deprecated: use go.starlark.net/lib/json instead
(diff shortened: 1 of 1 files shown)


adonovan (approved):
Well spotted! Thanks for the fix.

PR #489, opened 2023-07-21
Fix build flags for s390x, ppc64, and ppc64le

alexsaezm:
According to the [documentation](https://pkg.go.dev/cmd/go#hdr-Build_constraints), ppc64x is not a valid flag, making `TestIntFallback` fail. s390x is also missing and it works fine.

```
--- FAIL: TestIntFallback (0.01s)
    int_test.go:130: expected warning was not printed. Output=<<>>
```


adonovan (approved):
Thank you!

PR #491, opened 2023-07-25
lib/proto: add support for instantiating proto map fields

tazjin:
With this change, it becomes possible to instantiate `map<k,v>` type fields in Protobuf messages from Starlark.

Maps can be constructed from any Starlark type that implements starlark.IterableMapping.

Protobuf messages can have most types as keys/values, so the type conformity is checked individually for each key/value pair (as the Starlark side of things is dynamically typed).

This has been tested against fairly complex proto messages containing maps.

Map operations apart from construction are not supported in this CL.

### DIFF
--- a/lib/proto/proto.go
+++ b/lib/proto/proto.go
@@ -395,7 +395,6 @@ func setField(msg protoreflect.Message, fdesc protoreflect.FieldDescriptor, valu
 		}
 		defer iter.Done()
 
-		// TODO(adonovan): handle maps
 		list := msg.Mutable(fdesc).List()
 		list.Truncate(0)
 		var x starlark.Value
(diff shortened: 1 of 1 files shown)


tazjin:
Fixed author email address for CLA.

lib/proto/proto.go
```diff
@@ -409,6 +408,45 @@ func setField(msg protoreflect.Message, fdesc protoreflect.FieldDescriptor, valu
...
+		}
+
+		iter := mapping.Iterate()
+		defer iter.Done()
+
+		// Each value is converted using toProto as usual, passing the key/value
+		// field descriptors to check their types.
+		mutMap := msg.Mutable(fdesc).Map()
+		var k starlark.Value
+		for iter.Next(&k) {
```
adonovan, line 424: This might be cleaner if you  called mapping.Items() (though of course that does allocate 2*len values temporarilly).

adonovan (approved):
Thanks for this contribution!

lib/proto/proto.go
tazjin: I decided to go for the iterator route, as the list code is also using iterators and I wanted to avoid the extra allocations (Rust-brain shining through :)).

The difference is not that large, and it doesn't reduce the verbosity inside the loop, so I would prefer to leave it as is. If you feel strongly the other way, I'm happy to change it.

lib/proto/proto.go
```diff
@@ -409,6 +408,45 @@ func setField(msg protoreflect.Message, fdesc protoreflect.FieldDescriptor, valu
 		return nil
 	}
 
+	if fdesc.IsMap() {
+		mapping, ok := value.(starlark.IterableMapping)
+		if !ok {
+			return fmt.Errorf("in map field %s: expected mappable type, but got %s", value.Type())
```
adonovan, line 414: "...", fdesc.Name(), value.Type()

lib/proto/proto.go
```diff
@@ -409,6 +408,45 @@ func setField(msg protoreflect.Message, fdesc protoreflect.FieldDescriptor, valu
...
+				return fmt.Errorf("in key of map field %s: %w", fdesc.Name(), err)
+			}
+
+			// `found` is discarded, as the presence of the key in the
+			// iterator guarantees the presence of some value (even if it is
+			// starlark.None). Mismatching values will be caught in toProto
+			// below.
+			v, _, err := mapping.Get(k)
+			if err != nil {
+				return fmt.Errorf("in map field %s, at key: %w", fdesc.Name(), k.String(), err)
```
adonovan, line 436: "..., at key %v: %w"

lib/proto/proto.go
```diff
@@ -409,6 +408,45 @@ func setField(msg protoreflect.Message, fdesc protoreflect.FieldDescriptor, valu
...
+			}
+
+			vproto, err := toProto(fdesc.MapValue(), v)
+			if err != nil {
+				return fmt.Errorf("in map field %s, at key: %w", fdesc.Name(), k.String(), err)
```
adonovan, line 441: ", at key %v: %w"

tazjin:
Fixed those, thanks.

adonovan (approved):
Looks good! Thanks again.

PR #494, opened 2023-07-31
Replace `ctx.new_file` with `ctx.actions.declare_file`

mai93:
This will change the output directory from `blaze-genfiles` to `blaze-bin`


tetromino (approved):

adonovan:
FYI: this change has no significant effect, since the file being changed is only used as test input data to the Starlark scanner; it may look like Bazel configuration, but it is never used by Bazel.

Issue #492, opened 2023-07-26
FR: Support More Methods for type `set`


ohmyx:
As per PYthon doc : https://docs.python.org/3.10/library/stdtypes.html#set-types-set-frozenset

The five methods should be added:
```
add(elem)
Add element elem to the set.

remove(elem)
Remove element elem from the set. Raises KeyError if elem is not contained in the set.

discard(elem)
Remove element elem from the set if it is present.

pop()
Remove and return an arbitrary element from the set. Raises KeyError if the set is empty.

clear()
Remove all elements from the set.
```

Now it's too hard to use, for example, add() can only be implemented via helper like:

```python
def set_add(s, x):
    l = list(s)
    l.append(x)
    return set(l)
```

qms95:
+1
当前set太鸡肋了！

adonovan:
These operations are all compatible with Python and consistent with starlark-go's notion of `set`, and they are useful but awkward or inefficient to implement in Starlark itself, so I would accept a PR to add all of these methods. (`set` is a starlark-go-specific extension, so there's no need for a proposal.)

SamWheating:
I'm looking for something to work on, so I've started implementing this. Feel free to assign this issue to me and I'll try to have a PR open soon.

adonovan:
The `|`, `&`, `^` operators on sets are already supported, but the`-`, `<=`, and `>=` operators are missing. The `union` method is present but the others are missing. All seem like reasonable features to add.

ohmyx:
> The `|`, `&`, `^` operators on sets are already supported, but the`-`, `<=`, and `>=` operators are missing. The `union` method is present but the others are missing. All seem like reasonable features to add.

Yes. It would be nice to have those methods, 'cause the symbols are initially elusive in understanding. @adonovan 

```
Welcome to Starlark (go.starlark.net)
>>> x = set([1,2,3])
>>> y = set([3,5,6])
>>> x & y
set([3])
>>> x | y
set([1, 2, 3, 5, 6])
>>> x - y
Traceback (most recent call last):
  <stdin>:1:3: in <expr>
Error: unknown binary op: set - set
>>> x ^ y
set([1, 2, 5, 6])
>>> x <= y
Traceback (most recent call last):
  <stdin>:1:3: in <expr>
Error: set <= set not implemented
>>> x >= y
Traceback (most recent call last):
  <stdin>:1:3: in <expr>
Error: set >= set not implemented
```

SamWheating:
[...]

PR #497, opened 2023-08-14
Additional methods for Set type

SamWheating:
Re: https://github.com/google/starlark-go/issues/492

Added `set.add(x)`, `set.pop()`, `set.remove(x)`, `set.discard(x)`, `set.clear()`, as well as updates to the Spec and tests.

Demo:
```python
>>> a = set([1,2,3])
>>> a.add(4)
>>> a
set([1, 2, 3, 4])
>>> a.remove(1)
>>> a
set([2, 3, 4])
>>> a.remove(1)
  <stdin>:1:9: in <expr>
Error in remove: remove: missing key
>>> a.discard(2)
>>> a
set([3, 4])
>>> a.discard(2)
>>> a.pop()
3
>>> a
set([4])
>>> a.pop()
4
>>> a.pop()
  <stdin>:1:6: in <expr>
Error in pop: pop: empty set
>>> a.add(1)
>>> a.add(2)
>>> a
set([1, 2])
>>> a.clear()
>>> a
set([])
```


doc/spec.md
```diff
@@ -154,6 +154,11 @@ reproducibility is paramount, such as build tools.
     * [list·pop](#list·pop)
     * [list·remove](#list·remove)
     * [set·union](#set·union)
+    * [set·add](#set·add)
```
adonovan, line 157: Let's keep these in alphabetical order.

doc/spec.md
```diff
@@ -966,7 +971,15 @@ Sets are instantiated by calling the built-in `set` function, which
 returns a set containing all the elements of its optional argument,
 which must be an iterable sequence.  Sets have no literal syntax.
 
-The only method of a set is `union`, which is equivalent to the `|` operator.
+A set has these methods:
+
+* [`union`](#set·union)
```
adonovan, line 976: ditto

doc/spec.md
```diff
@@ -3755,6 +3768,68 @@ y = set([2, 3])
...
+```python
+x = set([1, 2])
+x.add(3)                             # None (x == set([1, 2, 3]))
+x.add(3)                             # None (x == set([1, 2, 3]))
+```
+
+<a id='set·pop'></a>
+### set·pop
+
+`S.pop()` removes an arbitrary item from the set and returns it.
```
adonovan, line 3785: There is no "arbitrary" in Starlark, so let's describe this as the "first inserted element".

doc/spec.md
```diff
@@ -3755,6 +3768,68 @@ y = set([2, 3])
...
+
+`pop` fails if the set is empty.
+
+```python
+x = set([1, 2, 3])
+x.pop()                              # 1 (x == set([2, 3]))
```
adonovan, line 3791: Clearer perhaps as:
```
x = set([1, 2])
x.pop() # 1
x.pop() # 2
x.pop() # error: element not found
```

doc/spec.md
```diff
@@ -3755,6 +3768,68 @@ y = set([2, 3])
...
+```
+
+<a id='set·remove'></a>
+### set·remove
+
+`S.remove(x)` removes `x` from the set and returns None.
+
+`remove` fails if the set does not contain `x` or is frozen.
```
adonovan, line 3799: All the mutating methods should ideally mention what happens when the set is frozen, especially given the subtlety that `add` may not always attempt to mutate the set (if the element is already present), so it does not always fail on a frozen dict. The spec should probably take a position on this, by stating that the lookup precedes the attempt at mutation (that's what dict.setdefault does, although the document is not particularly explicit).

doc/spec.md
```diff
@@ -3755,6 +3768,68 @@ y = set([2, 3])
...
+
+```python
+x = set([1, 2, 3])
+x.remove(2)                             # None (x == set([1, 3]))
```
adonovan, line 3803: Split this into two lines:

```
x.remove(2) # None
x                    # set([1, 3])

File: doc/spec.md [hunk 7]
```diff
@@ -3755,6 +3768,68 @@ y = set([2, 3])
...
+x.remove(2)                             # error: element not found
+```
+
+<a id='set·discard'></a>
+### set·discard
+
+`S.discard(x)` removes `x` from the set and returns None.
```
adonovan, line 3810: Given my mutation comment above, it might be easier to express by cases. Also, when the return value is trivial (None) it can be relegated to a secondary paragraph. Something like:

> If x is an element of set S, S.discard(x) removes x from the set, or fails if the set is frozen. If x is not an element of the set, discard has no effect.

> It returns None.

starlark/library.go
```diff
@@ -140,7 +140,12 @@ var (
 	}
 
 	setMethods = map[string]*Builtin{
-		"union": NewBuiltin("union", set_union),
+		"union":   NewBuiltin("union", set_union),
```
adonovan, line 143: sort

starlark/library.go
```diff
@@ -2183,6 +2188,74 @@ func set_union(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error)
 	return union, nil
 }
 
+// https://github.com/google/starlark-go/blob/master/doc/spec.md#set·add.
+func set_add(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error) {
+	var elem Value
+	if err := UnpackPositionalArgs(b.Name(), args, kwargs, 0, &elem); err != nil {
```
adonovan, line 2194: Shouldn't this be 1 (otherwise set.add() will insert a nil Value)?

starlark/library.go
```diff
@@ -2183,6 +2188,74 @@ func set_union(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error)
...
+}
+
+// https://github.com/google/starlark-go/blob/master/doc/spec.md#set·remove.
+func set_remove(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error) {
+	var k Value
+	if err := UnpackPositionalArgs(b.Name(), args, kwargs, 1, &k); err != nil {
+		return nil, err
+	}
+	if found, err := b.Receiver().(*Set).Delete(k); err != nil {
+		return nil, nameErr(b, err) // dict is frozen or key is unhashable
```
adonovan, line 2211: If key is unhashable then it can't be a member of the set. Shouldn't it simply return in that case? I wonder what Python does.

starlark/library.go
```diff
@@ -2183,6 +2188,74 @@ func set_union(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error)
...
+}
+
+// https://github.com/google/starlark-go/blob/master/doc/spec.md#set·discard.
+func set_discard(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error) {
+	var k Value
+	if err := UnpackPositionalArgs(b.Name(), args, kwargs, 1, &k); err != nil {
+		return nil, err
+	}
+	if _, err := b.Receiver().(*Set).Delete(k); err != nil {
+		return nil, nameErr(b, err) // dict is frozen or key is unhashable
```
adonovan, line 2225: ditto

starlark/testdata/set.star
```diff
@@ -116,3 +116,27 @@ assert.eq(iter(), [1, 2, 3])
...
+assert.true(4 not in x)
+assert.fails(lambda: x.remove(4), "remove: missing key")
+
+# discard
+x.add(4)
+x.discard(4)
+assert.true(4 not in x)
+assert.eq(x.discard(4), None)
+
+# pop
```
adonovan, line 135: Add a test that pop removes the first inserted element.

starlark/testdata/set.star
```diff
@@ -116,3 +116,27 @@ assert.eq(iter(), [1, 2, 3])
 
 # sets are not indexable
 assert.fails(lambda : x[0], "unhandled.*operation")
+
+# adding and removing
+x.add(4)
```
adonovan, line 121: Add `x = set()` before each of these stanzas so that there is no residual effect across them.

starlark/testdata/set.star
```diff
@@ -116,3 +116,27 @@ assert.eq(iter(), [1, 2, 3])
...
+assert.true(4 not in x)
+assert.eq(x.discard(4), None)
+
+# pop
+z = set([1])
+assert.eq(z.pop(), 1)
+assert.fails(lambda: z.pop(), "pop: empty set")
+
+# clear
+x.clear()
```
adonovan, line 141: Can you add tests of each operation on a frozen set (see dict.star for the template), including the edge cases where a potentially mutating operation does not attempt to mutate the set?

adonovan (commented):
Hi Sam, many thanks for this contribution. I've made a number of comments but the necessary changes should all be fairly straightforward.

starlark/library.go
SamWheating: Good point - looks like Python 3.7.13 will throw a typeError when trying to remove (or discard) an unhashable type from a set.
```
>>> a = set([1,2,3])
>>> a.remove(1)
>>> a.remove([4,5,6])
  File "<stdin>", line 1, in <module>
TypeError: unhashable type: 'list'
```
 I think that this makes sense, especially in the case of `remove` where a successful return indicates that the item was found and removed. Thoughts?

doc/spec.md
SamWheating: Good point, all of the frozen checks defer to the underlying hashtable so I wasn't thinking about this case. I'll verify and document this behaviour. 

Related - how do you feel about calling `clear()` on an empty, frozen dict or set? Should that succeed?

starlark/testdata/set.star
SamWheating: Starlark doesn't permit this because of global variable reassignment, but I will just create a unique set for each block.

doc/spec.md
SamWheating: Landed on the following semantics around frozen sets:
 - `set.clear()` will succeed on an empty frozen set
 - `set.discard(x)` will succeed on a frozen set which does not contain `x`
 - `set.add(x)` will succeed on a frozen set which already contains `x`
 - `pop()` and `remove()` will always fail on frozen sets with `cannot delete from frozen hash table`. 
 
 Two edge cases to consider - 
  - Should `pop()` on an empty frozen set fail because its empty, or because its frozen?
  - Should `remove(x)` on a frozen set not containing x fail because of missing key, or because its frozen?

doc/spec.md
```diff
@@ -3740,6 +3753,79 @@ x.remove(2)                             # None (x == [1, 3])
...
+x.discard(2)                             # None
+x                                        # set([1, 3])
+```
+
+<a id='set·pop'></a>
+### set·pop
+
+`S.pop()` removes the first inserted item from the set and returns it.
+
+`pop` fails if the set is empty or is frozen.
```
adonovan, line 3806: empty or frozen

starlark/library.go
```diff
@@ -2168,6 +2173,87 @@ func string_splitlines(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value
...
+	}
+	return None, nil
+}
+
+// https://github.com/google/starlark-go/blob/master/doc/spec.md#set·clear.
+func set_clear(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error) {
+	if err := UnpackPositionalArgs(b.Name(), args, kwargs, 0); err != nil {
+		return nil, err
+	}
+	if b.Receiver().(*Set).Len() == 0 { // clear on an empty set is non-mutating
```
adonovan, line 2199: Factor the returns:

```
if b.Receiver().(*Set).Len() > 0 {	
	if err := b.Receiver().(*Set).Clear(); err != nil {
		return nil, nameErr(b, err)
	}
}
```

The intent of the code is self-explanatory so it doesn't need a comment.

starlark/library.go
```diff
@@ -2168,6 +2173,87 @@ func string_splitlines(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value
...
+}
+
+// https://github.com/google/starlark-go/blob/master/doc/spec.md#set·discard.
+func set_discard(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error) {
+	var k Value
+	if err := UnpackPositionalArgs(b.Name(), args, kwargs, 1, &k); err != nil {
+		return nil, err
+	}
+	if found, err := b.Receiver().(*Set).Has(k); err != nil {
+		return nil, nameErr(b, err)
```
adonovan, line 2216: For discard, unhashable => not found, so I think this should not be an error. Can we add a test case?

starlark/library.go
```diff
@@ -2168,6 +2173,87 @@ func string_splitlines(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value
...
+	} else if !found {
+		return None, nil
+	}
+	if _, err := b.Receiver().(*Set).Delete(k); err != nil {
+		return nil, nameErr(b, err) // dict is frozen or key is unhashable
```
adonovan, line 2221: `// frozen`

(We've already checked that the key is hashable.)

adonovan (commented):
Sorry for the delay. More nitpicks, but we're nearly there. Thanks again.

starlark/library.go
SamWheating: Had some thoughts about this here:
https://github.com/google/starlark-go/pull/497#discussion_r1294783317

If we're aiming for consistency with the python implementation we should be throwing an error on an unhashable discard. Thoughts?

SamWheating:
Thanks for the review, got everything addressed aside from a few open discussions.

While you're here - how would you feel about adding built-in set parsing and set comprehension? I'd be interested in working on the implementation but also think this might be too much divergence from the official spec.

starlark/library.go
adonovan: That seems fine. Thanks for checking.

doc/spec.md
adonovan: > Should pop() on an empty frozen set fail because its empty, or because its frozen?
> Should remove(x) on a frozen set not containing x fail because of missing key, or because its frozen?

Generally we don't specify which error you get when more than one applies. What's more important is that we specify whether successful shortcuts trump slow-path errors (as in the case of discarding an unhashable from an empty set: does it fail because unhashable or succeed because empty?). So what you have looks fine.

adonovan:
> how would you feel about adding built-in set parsing and set comprehension? I'd be interested in working on the implementation but also think this might be too much divergence from the official spec.

I would rather not implement those since they change the dialect accepted by the parser. If you can persuade the Bazel folks to accept set as a standard feature, then these expressions would be appropriate, but I don't think they feel much need for it.


Thanks again for contributing this, and for your patience during the review.

adonovan (approved):

PR #510, opened 2023-09-22
Make resolve follow FileOptions.TopLevelControl

marco6:
The new struct `FileOptions` allows for fine grained control over language feature.

However, it didn't respect `FileOptions.TopLevelControl` flag, as the only way to actually allow if/for/while in the top level context was to enable `FileOptions.GlobalReassign` flag.

### DIFF
--- a/resolve/resolve.go
+++ b/resolve/resolve.go
@@ -505,7 +505,7 @@ func (r *resolver) stmt(stmt syntax.Stmt) {
 		}
 
 	case *syntax.IfStmt:
-		if !r.options.GlobalReassign && r.container().function == nil {
+		if !r.options.GlobalReassign && !r.options.TopLevelControl && r.container().function == nil {
 			r.errorf(stmt.If, "if statement not within a function")
 		}
 		r.expr(stmt.Cond)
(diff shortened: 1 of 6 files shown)


resolve/resolve.go
```diff
@@ -505,7 +505,7 @@ func (r *resolver) stmt(stmt syntax.Stmt) {
 		}
 
 	case *syntax.IfStmt:
-		if !r.options.GlobalReassign && r.container().function == nil {
+		if !r.options.GlobalReassign && !r.options.TopLevelControl && r.container().function == nil {
```
adonovan, line 508: I think these three should just be `if !r.options.TopLevelControl ...`, since the GlobalReassign flag was split into two fine-grained options fields.

resolve/resolve_test.go
```diff
@@ -16,14 +16,14 @@ import (
 
 // A test may enable non-standard options by containing (e.g.) "option:recursion".
 func getOptions(src string) *syntax.FileOptions {
-	// TODO(adonovan): use new fine-grained names.
```
adonovan, line 19: The intent of this TODO was to get rid of the test option `globalreassign` and express it in terms of the fine-grained flags. When the TODO is done, every field in getOptions' FileOptions will be set by exactly one `option(src, "...")` call. Would you care to make that change?

adonovan (commented):
Thanks for this contribution!

resolve/resolve_test.go
marco6: Ok, I think I did that. I removed the new tests from `resolve.star` as that functionality was already tested in others. I updated the flags in the tests according to this suggestion.

resolve/resolve.go
marco6: Done.

adonovan (approved):
Looks great, thank you!

Issue #381, opened 2021-08-23
proto: Support map fields


euroelessar:
There is a TODO in the codebase, so I assume the support is desired.
https://github.com/google/starlark-go/blob/master/lib/proto/proto.go#L84

There are two primary questions in regard of the possible implementation:
1. Ensuring deterministic ordering is tricky, the only feasible way is to sort all keys before the iteration (meaning lexicographical order as in opposite to the insertion one in regular dicts).
2. Are there any tests for the package? Where should they be put?

tazjin:
I'm very interested in sorting this out at the moment, at least for struct initialisation. @adonovan, would you accept a PR for this? (CLA is on file, and I'm a [Xoogler](http://epitaphs/tazjin)).

tazjin:
Turned out to be fairly easy, map construction is here: https://github.com/google/starlark-go/pull/491

It seems like there isn't great test coverage of the lib/proto functionality at the moment. Didn't touch that too much. This code exists in a private fork, too, and has been tested against some fairly complex protos with map fields there.

PR #511, opened 2023-10-02
Support setting and iterating over map keys

negz:
Fixes https://github.com/google/starlark-go/issues/381

This PR introduces a new `MapField` type, which can be used to set and access map field values by key.

Without this PR I get a panic in `toStarlark1` where `x.Message()` is called on a map field value. (Perhaps because my proto has some `map<string, Message>` fields):

```
panic: type mismatch: cannot convert map to message
```

I've also updated the nascent `proto.star` unit tests to use their own proto file, to test repeated and map fields.

Separately from the main change, I've included a commit that adds a list-like `append` method to `RepeatedField`. I can break that out into a separate PR if you'd prefer.


negz:
@adonovan Gentle ping. 🙂 Is this a contribution you'd be interested in accepting?

adonovan:
Sorry for the very long delay. I just spent 20m playing with this and it looks like the right approach. It'll take me a little longer to review the code change properly but I'll try to get to it this week.

Thanks for your patience.

lib/proto/proto.go
```diff
@@ -969,6 +1040,145 @@ func (it *repeatedFieldIterator) Done() {
...
+	})
+
+	return out
+}
+
+func (mf *MapField) Len() int { return mf.mp.Len() }
+
+func (mf *MapField) String() string {
+	// We want to use {k1: v1, k2: v2} notation, like a dict.
+	buf := new(bytes.Buffer)
```
adonovan, line 1142: Use strings.Builder to save an allocation, since the result is desired as a string.

starlark/eval_test.go
```diff
@@ -115,7 +117,23 @@ func TestExecFile(t *testing.T) {
 	testdata := starlarktest.DataFile("starlark", ".")
 	thread := &starlark.Thread{Load: load}
 	starlarktest.SetReporter(thread, t)
-	proto.SetPool(thread, protoregistry.GlobalFiles)
+
+	// This proto is used for the proto.star tests. It's generated by running:
```
adonovan, line 121: Nice.

starlark/testdata/proto.star
```diff
@@ -3,12 +3,27 @@
...
+
+assert.eq(list(m.repeated_field), ["a", "b", "c"])
+m.repeated_field = ["d", "e"]
+assert.eq(list(m.repeated_field), ["d", "e"])
+m.repeated_field.append("f")
+assert.eq(list(m.repeated_field), ["d", "e", "f"])
+
+assert.eq(dict(m.map_field), {"a": "A", "b": "B"})
+m.map_field["c"] = "C"
+assert.eq(dict(m.map_field), {"a": "A", "b": "B", "c": "C"})
```
adonovan, line 29: Could we have some tests of the various failure modes?

starlark/testdata/proto/test.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
```
adonovan, line 1: // This proto schema is used for tests in proto.star.
// Re-generate the binary FileDescriptorSet file by 
// running: protoc ...(etc).
// (Requires the "protobuf" brew/apt package.)

adonovan (commented):
Thanks for this contribution, and sorry it took so long to review. Overall it looks good. I've left a lot of comments but don't be discouraged--they're mostly pretty superficial.

negz:
Unfortunately I'm unlikely to find time to work on this again soon. I'd still appreciate the change, but it's not a priority. I'll close this PR.

adonovan:
If you don't mind, I'll reopen it just so we don't forget that it exists. You or I or someone else may pick it up when the muse strikes.

adonovan:
Re-closing this now that #581 has superseded it.

PR #515, opened 2023-10-16
Limit CPU and memory usage for int-to-float conversion

marco6:
This PR adds a fast path for very big int to float conversion that always get converted to `±Inf`.


adonovan:
> This PR adds a fast path for very big int to float conversion that always get converted to ±Inf.

Was this change opportunistic, or did the logic appear in a profile? I'm curious what workload would make heavy use of numbers larger than the universe.

marco6:
Opportunistic: we had to do that for concerns not related to performance and it felt in line with the previous optimization.

adonovan (approved):

Issue #516, opened 2023-10-27
time: allow to inject per-module or per-thread `now()`


ash2k:
This is a follow up for https://github.com/google/starlark-go/issues/435.

Global var is still there in the `time` module:

https://github.com/google/starlark-go/blob/22325403fcb3696ebcc542d0e7cc75156d1fdca3/lib/time/time.go#L74

I'd like to be able to customize what `now()` returns on a per-thread basis. Alternatively, I'd like to be able to instantiate a new `time` module with injected `now()` (or both).

Thank you!

adonovan:
I would accept a PR to make the `now` function consult a `starlark.Thread`-local variable (of type `func() time.Now`) before falling back to the standard `time.Now` implementation.

ash2k:
Thank you. I'll open a PR soon.

Do you want me to keep the existing package var for backward compatibility or remove it?

adonovan:
We can't remove it, but we can deprecate it and recommend that people use the new mechanism instead, which avoids global state.

PR #517, opened 2023-10-29
time: per-thread time.now() function

ash2k:
Closes https://github.com/google/starlark-go/issues/516.

p.s. there are quite a few changes that go fmt wants to make. I kept them out of this PR.

### DIFF
--- a/lib/time/time.go
+++ b/lib/time/time.go
@@ -6,6 +6,7 @@
 package time // import "go.starlark.net/lib/time"
 
 import (
+	"errors"
 	"fmt"
 	"sort"
 	"time"
@@ -68,10 +69,28 @@ var Module = &starlarkstruct.Module{
 	},
 }
 
-// NowFunc is a function that generates the current time. Intentionally exported
+// NowFunc is a function that reports the current time. Intentionally exported
 // so that it can be overridden, for example by applications that require their
 // Starlark scripts to be fully deterministic.
+//
+// Deprecated: avoid updating this global variable
+// and instead use SetNow on each thread to set its clock function.
 var NowFunc = time.Now
+
+const contextKey = "time.now"
+
+// SetNow sets the thread's optional clock function.
+// If non-nil, it will be used in preference to NowFunc when the
+// thread requests the current time by executing a call to time.now.
+func SetNow(thread *starlark.Thread, nowFunc func() (time.Time, error)) {
+	thread.SetLocal(contextKey, nowFunc)
+}
+
+// Now returns the clock function previously associated with this thread.
+func Now(thread *starlark.Thread) func() (time.Time, error) {
+	nowFunc, _ := thread.Local(contextKey).(func() (time.Time, error))
+	return nowFunc
+}
 
 func parseDuration(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
 	var d Duration
@@ -129,7 +148,19 @@ func fromTimestamp(thread *starlark.Thread, _ *starlark.Builtin, args starlark.T
 }
 
 func now(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
-	return Time(NowFunc()), nil
+	nowErrFunc := Now(thread)
+	if nowErrFunc != nil {
+		t, err := nowErrFunc()
+		if err != nil {
+			return nil, err
+		}
+		return Time(t), nil
+	}
+	nowFunc := NowFunc
+	if nowFunc == nil {
+		return nil, errors.New("time.now() is not available")
+	}
+	return Time(nowFunc()), nil
 }
 
 // Duration is a Starlark representation of a duration.
--- a/lib/time/time_test.go
+++ b/lib/time/time_test.go
@@ -1,0 +1,82 @@
+package time
+
+import (
+	"errors"
+	"testing"
+	"time"
+
+	"go.starlark.net/starlark"
+)
+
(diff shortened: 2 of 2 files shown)


lib/time/time.go
```diff
@@ -71,8 +71,25 @@ var Module = &starlarkstruct.Module{
...
 // Starlark scripts to be fully deterministic.
+//
+// Deprecated: use SetNow() to set per-thread clock.
 var NowFunc = time.Now
 
+const contextKey = "time.Now"
+
+// SetNow associates with the specified Starlark thread a function that generates the current time.
+// Can be used for example by applications that require their
+// Starlark scripts to be fully deterministic.
```
adonovan, line 82: Let's phrase this as:

// SetNow sets the thread's optional clock function.
// If non-nil, it will be used in preference to NowFunc when the
// thread requests the current time by executing a call to time.now.

lib/time/time.go
```diff
@@ -71,8 +71,25 @@ var Module = &starlarkstruct.Module{
 // NowFunc is a function that generates the current time. Intentionally exported
 // so that it can be overridden, for example by applications that require their
 // Starlark scripts to be fully deterministic.
+//
+// Deprecated: use SetNow() to set per-thread clock.
 var NowFunc = time.Now
 
+const contextKey = "time.Now"
```
adonovan, line 78: now (lowercase) since that's the Starlark name.

lib/time/time.go
```diff
@@ -71,8 +71,25 @@ var Module = &starlarkstruct.Module{
 // NowFunc is a function that generates the current time. Intentionally exported
```
adonovan, line 71: s/generates/reports/ would be better.

lib/time/time.go
```diff
@@ -71,8 +71,25 @@ var Module = &starlarkstruct.Module{
...
 // so that it can be overridden, for example by applications that require their
 // Starlark scripts to be fully deterministic.
+//
+// Deprecated: use SetNow() to set per-thread clock.
```
adonovan, line 75: // Deprecated: avoid updating this global variable
// and instead use SetNow on each thread to set its clock function.

lib/time/time.go
```diff
@@ -71,8 +71,25 @@ var Module = &starlarkstruct.Module{
...
+const contextKey = "time.Now"
+
+// SetNow associates with the specified Starlark thread a function that generates the current time.
+// Can be used for example by applications that require their
+// Starlark scripts to be fully deterministic.
+func SetNow(thread *starlark.Thread, nowFunc func() time.Time) {
+	thread.SetLocal(contextKey, nowFunc)
+}
+
+// Now returns the function to generate the current time previously associated with this thread.
```
adonovan, line 87: ...the clock function previously...

adonovan (commented):
Thanks! Code looks good.

lib/time/time.go
```diff
@@ -129,7 +147,11 @@ func fromTimestamp(thread *starlark.Thread, _ *starlark.Builtin, args starlark.T
 }
 
 func now(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
-	return Time(NowFunc()), nil
+	nowFunc := Now(thread)
```
adonovan, line 150: It occurred to me that it might be important for deterministic applications to be able to prevent access to the clock entirely, not just fix it at some particular time. If so, then the signature of the per-thread time function should be func() (time.Time, error), and the logic here should return a Starlark error if the per-thread now func returns an error, or if the NowFunc var is nil (instead of panicking).  What do you think?

adonovan (approved):

adonovan:
> p.s. there are quite a few changes that go fmt wants to make. I kept them out of this PR.

Thanks. I've posted https://github.com/google/starlark-go/pull/519.

lib/time/time.go
ash2k: I think this is an excellent observation! I'll push an update in a minute.

lib/time/time.go
```diff
@@ -68,11 +69,29 @@ var Module = &starlarkstruct.Module{
...
+
+// SetNow sets the thread's optional clock function.
+// If non-nil, it will be used in preference to NowFunc when the
+// thread requests the current time by executing a call to time.now.
+func SetNow(thread *starlark.Thread, nowFunc func() (time.Time, error)) {
+	thread.SetLocal(contextKey, nowFunc)
+}
+
+// Now returns the clock function previously associated with this thread.
+func Now(thread *starlark.Thread) func() (time.Time,error) {
```
adonovan, line 90: Be sure to `go fmt` the code before publishing.

lib/time/time_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"testing"
+	"time"
+
+	"go.starlark.net/starlark"
+)
+
+func TestPerThreadNowReturnsCorrectTime(t *testing.T) {
+	th := &starlark.Thread{}
+	date := time.Date(1, 2, 3, 4, 5, 6, 7, time.UTC)
+	SetNow(th, func() (time.Time, error) {
```
adonovan, line 14: We should defer a second call to SetNow to restore the old value of the global variable, to avoid perturbing other tests.

adonovan (commented):
Thanks for the tests!

lib/time/time.go
ash2k: Ah, I disabled autoformatting in my IDE as it messes with all the comments in the file. Fixed.

lib/time/time_test.go
ash2k: But this is per-thread and we throw that thread away anyway?

adonovan: Of course. Never mind!

PR #520, opened 2023-11-20
Remove 'too many arguments' error when interpolating over mappings

TheSignPainter98:
When no conversions are specified, interpolation over an empty tuple is accepted, but interpolation over an empty dict is rejected:
```
>>> 'foo' % ()
"foo"
>>> 'foo' % {}
  <stdin>:1:7: in <expr>
Error: too many arguments for format string
```
For context, extra dict arguments are not rejected when a conversion is present:
```
>>> '%(a)d' % {'a': 1, 'unused': 2}
"1"
```
This PR relaxes the 'too many arguments' check for mappings and adds tests to ensure consistency between tuples and dicts

### DIFF
--- a/starlark/eval.go
+++ b/starlark/eval.go
@@ -1659,11 +1659,14 @@ func interpolate(format string, x Value) (Value, error) {
 		index++
 	}
 
-	if _, ok := x.(Mapping); !ok {
-		if index < nargs {
-			return nil, fmt.Errorf("too many arguments for format string")
-		}
+	if index < nargs && !is[Mapping](x) {
+		return nil, fmt.Errorf("too many arguments for format string")
 	}
 
 	return String(buf.String()), nil
+}
+
+func is[T any](x any) bool {
+	_, ok := x.(T)
+	return ok
 }
--- a/starlark/eval.go
+++ b/starlark/eval.go
@@ -1659,8 +1659,10 @@ func interpolate(format string, x Value) (Value, error) {
 		index++
 	}
 
-	if index < nargs {
-		return nil, fmt.Errorf("too many arguments for format string")
+	if _, ok := x.(Mapping); !ok {
+		if index < nargs {
+			return nil, fmt.Errorf("too many arguments for format string")
+		}
 	}
 
 	return String(buf.String()), nil
--- a/starlark/library.go
+++ b/starlark/library.go
@@ -40,7 +40,7 @@ func init() {
 		"True":      True,
 		"False":     False,
 		"abs":       NewBuiltin("abs", abs),
-		"any":       NewBuiltin("any", any),
+		"any":       NewBuiltin("any", any_),
 		"all":       NewBuiltin("all", all),
 		"bool":      NewBuiltin("bool", bool_),
 		"bytes":     NewBuiltin("bytes", bytes_),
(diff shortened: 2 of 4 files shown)


starlark/eval.go
```diff
@@ -1659,8 +1659,10 @@ func interpolate(format string, x Value) (Value, error) {
 		index++
 	}
 
-	if index < nargs {
-		return nil, fmt.Errorf("too many arguments for format string")
+	if _, ok := x.(Mapping); !ok {
```
adonovan, line 1662: Let's check for Mapping only in the uncommon case:
```
if index < nargs && !is[Mapping](x) { ... }

func is[T any](x T) bool {
    _, ok := x.(T)
    return ok
}
```

adonovan (approved):
Thanks for this contribution! Code change looks sound.

Could you also add this line to the doc/spec.md file? Thanks.

After this paragraph:

> If the "%" is immediately followed by "(key)", ...

add a new paragraph:

> If the format string contains no conversions, the operand must be an empty tuple.

Thanks.

starlark/eval.go
TheSignPainter98: Sounds good, I'm just adding this. There's an unfortunate name clash with the `any` builtin, would it be better to rename this to `any_` or to use `interface{}` for `is` as I see more of the pre-1.18 in the codebase?

TheSignPainter98:
Adding the line to the spec, I think the suggestion isn't quite right as an empty mapping would also be acceptable. Perhaps something like this would work?
> If the format string contains no conversions, the operand must be a `Mapping` or an empty tuple.

starlark/eval.go
adonovan: > There's an unfortunate name clash with the any builtin, would it be better to rename this to any_ or to use interface{} for is as I see more of the pre-1.18 in the codebase?

What do you mean by name clash? The `is` function I suggested above (which should be placed at top level at the end of the file) intentionally refers to the `any` built-in. (You could say interface{} but that is almost never necessary since `any` is in scope nearly everywhere.)

TheSignPainter98: Ah I think we're referring to different things. I was referring to the implementation of Starlark's built-in `any(iter)` function. Currently, I get the following compilation error on `is`:
```
starlark/eval.go:1669:11: any (value of type func(thread *Thread, _ *Builtin, args Tuple, kwargs []Tuple) (Value, error)) not a type
```
My question is whether the implementation function for Starlark's `any()` should be renamed to allow Go's `any` built-in to appear in the signature of `is` however I don't see any use of Go's `any` in the codebase at the moment

starlark/library.go
```diff
@@ -210,7 +210,7 @@ func all(thread *Thread, _ *Builtin, args Tuple, kwargs []Tuple) (Value, error)
 }
 
 // https://github.com/google/starlark-go/blob/master/doc/spec.md#any
-func any(thread *Thread, _ *Builtin, args Tuple, kwargs []Tuple) (Value, error) {
+func any_(thread *Thread, _ *Builtin, args Tuple, kwargs []Tuple) (Value, error) {
```
adonovan, line 213: Ah, now I understand. Thanks for this fix!

adonovan (approved):
Thanks again!

PR #524, opened 2024-01-23
Remove debug clock from Time.String()

marco6:
`Time.String()` states in the documentation that it formats time according to the string `2006-01-02 15:04:05.999999999 -0700 MST`, however it uses `time.Time.String()` methods which, if available, adds to that string the monotonic clock, leading to inconsistent output. For example:

```
>>> time.now()
2024-01-23 14:47:50.352396992 +0100 CET m=+210.079650729
>>> time.parse_time("2024-01-23T14:47:50.352396992+01:00")
2024-01-23 14:47:50.352396992 +0100 CET
```

This PR uses a format string instead, so that the output does never contain the monotonic clock.

### DIFF
--- a/lib/time/time.go
+++ b/lib/time/time.go
@@ -19,37 +19,37 @@ import (
 // Module time is a Starlark module of time-related functions and constants.
 // The module defines the following functions:
 //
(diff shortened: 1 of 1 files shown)


adonovan (approved):
Good catch! Thanks for contributing a fix.

PR #525, opened 2024-02-06
feat: add a Makefile for easy building

HarikrishnanBalagopal:


.gitignore
```diff
@@ -0,0 +1,24 @@
+# If you prefer the allow list template instead of the deny list, see community template:
```
adonovan, line 1: This is your personal configuration; it belongs in $HOME/.gitignore, not this project.

Makefile
```diff
@@ -0,0 +1,5 @@
+BIN_DIR='bin/'
+
+.PHONY: build
+build:
+	mkdir -p ${BIN_DIR} && go build -o ${BIN_DIR} ./...
```
adonovan, line 5: Again, this is your personal configuration. Most contributors use 'go test ./...' or "go install ./cmd/starlark".

.gitignore
HarikrishnanBalagopal: Yes it's opinionated, I just copied it from https://github.com/github/gitignore/blob/main/Go.gitignore

Makefile
HarikrishnanBalagopal: Can we document this is the README? I added the minimal Makefile because there was the README talks about installing the CLI globally instead of building the repo locally.

adonovan: The install command you quoted is already document in the main README. We can add a note to say run `go test ./...`, but I think for most Go users that's the first thing they would try

PR #527, opened 2024-02-19
starlark: new API for go1.23 push iterators

adonovan:
starlark: new API for go1.23 push iterators
    
    This change defines two helpers, Elements(seq)
    and Entries(mapping), that adapt the Iterable and
    IterableMapping interfaces to one- and two-variable
    go1.23 range loops. The new syntax is more concise
    and frees the caller from worrying about Iterator.Done.
    
    We do not yet update any calls to use them, so that
    the project can continue to be build with pre-go1.23
    releases of Go.
    
    Also, define Elements methods on {*List,Tuple,*Set}
    and an Entries method on *Dict. These optimized iterators
    (which can fetch both k and v in one go) will be
    used by the Elements and Entries standalone functions
    when the operand supports it. (User-defined types
    can implement these methods too, although the
    interface isn't currently documented.)
    
    Also, a go1.23-only test of the new iteration.

### DIFF
--- a/starlark/hashtable.go
+++ b/starlark/hashtable.go
@@ -416,6 +416,16 @@ func (it *keyIterator) Done() {
 	}
 }
 
+// entries is a go1.23 iterator over the entries of the hash table.
+func (ht *hashtable) entries(yield func(k, v Value) bool) {
+	if !ht.frozen {
+		ht.itercount++
+		defer func() { ht.itercount-- }()
+	}
+	for e := ht.head; e != nil && yield(e.key, e.value); e = e.next {
+	}
+}
+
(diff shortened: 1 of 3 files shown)


jayconrod (commented):
This is cool. No real comments, since I've only loosely followed the discussion: this is pretty new to me.

Guarding new functionality behind a `1.23` build constraint seems fine.

adonovan (commented):
I've tidied things up and removed the breaking changes (they can happen later), so this is ready for proper review now.

jayconrod (approved):
Nice. This new syntax is quite elegant.

Issue #544, opened 2024-05-06
UnpackArgs has different behavior for `args` vs `kwargs`


nicks:
**Repro steps:**

1. Define a built-in function with UnpackArgs used like this:

```
if err := starlark.UnpackArgs("unpack", args, kwargs,
    "x", &x, "y?", &y, "z", &z); err != nil {
  t.Errorf("UnpackArgs failed: %v", err)
}
```

2. Pass two arguments to this function as named args:

```
f(1, 2)
```

This is OK.

3. Pass two arguments to this function as positional args:

```
f(x=1, y=2)
```

**Expected result:**

This should succeed. `y` is optional, so `z` should also be optional

**Actual result:**

```
UnpackArgs failed: unpack: missing argument for z
```

**Additional info**

I have a test in a branch that demonstrates the problem, might also open a PR if the fix is easy.

adonovan:
Did you mean `starlark.UnpackArgs("unpack", args, kwargs, ...`?

nicks:
oops, yes, fixed!

PR #545, opened 2024-05-06
starlark: tweak how UnpackArgs enforces optional args

nicks:
Fixes https://github.com/google/starlark-go/issues/544

### DIFF
--- a/starlark/eval_test.go
+++ b/starlark/eval_test.go
@@ -1109,3 +1109,36 @@ f(1)
 		t.Errorf("env() returned %s, want %s", got, want)
 	}
 }
+
+func TestUnpackArgsOptionalInference(t *testing.T) {
+	// success
+	kwargs := []starlark.Tuple{
+		{starlark.String("x"), starlark.MakeInt(1)},
+		{starlark.String("y"), starlark.MakeInt(2)},
+	}
+	var x, y, z int
+	if err := starlark.UnpackArgs("unpack", nil, kwargs,
+		"x", &x, "y?", &y, "z", &z); err != nil {
+		t.Errorf("UnpackArgs failed: %v", err)
+	}
+	got := fmt.Sprintf("x=%d y=%d z=%d", x, y, z)
+	want := "x=1 y=2 z=0"
+	if got != want {
+		t.Errorf("got %s, want %s", got, want)
+	}
+
+	// success
+	args := starlark.Tuple{starlark.MakeInt(1), starlark.MakeInt(2)}
+	x = 0
+	y = 0
+	z = 0
+	if err := starlark.UnpackArgs("unpack", args, nil,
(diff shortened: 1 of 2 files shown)


starlark/unpack.go
```diff
@@ -44,23 +44,23 @@ type Unpacker interface {
 //
 // Examples:
 //
-//      var (
-//          a Value
-//          b = MakeInt(42)
-//          c Value = starlark.None
-//      )
+//	var (
```
nicks, line 47: fyi -- these indentation changes come from go fmt in Go 1.22. I can separate them out into a separate pr if that's easier.

starlark/eval_test.go
```diff
@@ -1109,3 +1109,44 @@ f(1)
...
+	kwargs := []starlark.Tuple{
+		{starlark.String("x"), starlark.MakeInt(1)},
+		{starlark.String("y"), starlark.MakeInt(2)},
+	}
+	var x, y, z int
+	if err := starlark.UnpackArgs("unpack", nil, kwargs,
+		"x", &x, "y?", &y, "z", &z); err != nil {
+		t.Errorf("UnpackArgs failed: %v", err)
+	}
+	if x != 1 {
```
adonovan, line 1124: You can achieve higher vertical density (and, in this case, readability) in both the code and the test log by combining all three in one line:

got := fmt.Sprint("x=%d y=%d z=%d", x, y, z)
want := "x=1 y=2 z=0"
...assert...

starlark/unpack.go
adonovan: Don't worry, it's fine.

adonovan: While you're here, would you mind fixing these preexisting things?

L28: s/integer types uses/integer types use/
L36:  s/where or not/whether or not/

s/interface{}/any/g throughout this function

thanks

starlark/unpack.go
```diff
@@ -164,12 +163,15 @@ kwloop:
 	}
 
 	// Check that all non-optional parameters are defined.
-	// (We needn't check the first len(args).)
-	for i := len(args); i < nparams; i++ {
+	for i := 0; i < nparams; i++ {
 		name := pairs[2*i].(string)
 		if strings.HasSuffix(name, "?") {
 			break // optional
```
adonovan, line 169: The doc comment says "If a parameter is marked optional, then all following parameters are implicitly optional whether or not they are marked." In hindsight we should probably have disallowed non-optionals after optionals, but that would be a breaking change now.

adonovan (commented):
Thanks for the fix!

starlark/unpack.go
nicks: agreed

adonovan (approved):
Hi Nick, thanks for sending this fix. LGTM.

PR #551, opened 2024-05-16
starlark: require go1.23 for new iterators

adonovan:
The correct type for new iterators is iter.Seq, which depends on a go1.23-only package, so we must limit the new functions to new toolchains.

In principle this is (yet another) breaking change, but I suspect only users of (pre-)go1.23 will notice, and they are living on the bleeding edge by choice. (Even then, if the iterators are used only as `range` operands, they will be fine.)

Also, fix a couple of doc links.

### DIFF
--- a/starlark/iter.go
+++ b/starlark/iter.go
@@ -1,0 +1,118 @@
+// Copyright 2024 The Bazel Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build go1.23
+
+package starlark
+
+import (
+	"fmt"
+	"iter"
+)
+
+func (d *Dict) Entries() iter.Seq2[Value, Value] { return d.ht.entries }
+
+// Elements returns a go1.23 iterator over the elements of the list.
+//
+// Example:
+//
+//	for elem := range list.Elements() { ... }
+func (l *List) Elements() iter.Seq[Value] {
+	return func(yield func(Value) bool) {
+		if !l.frozen {
+			l.itercount++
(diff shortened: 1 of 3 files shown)


adonovan:
Thanks for reminding me to git add the new file! ;-)

starlark/iter.go
```diff
@@ -0,0 +1,118 @@
...
+)
+
+func (d *Dict) Entries() iter.Seq2[Value, Value] { return d.ht.entries }
+
+// Elements returns a go1.23 iterator over the elements of the list.
+//
+// Example:
+//
+//	for elem := range list.Elements() { ... }
+func (l *List) Elements() func(yield func(Value) bool) {
```
jayconrod, line 21: Should it return `iter.Seq[Value]` now?

starlark/iter.go
```diff
@@ -0,0 +1,118 @@
...
+	return func(yield func(Value) bool) {
+		for _, x := range t {
+			if !yield(x) {
+				break
+			}
+		}
+	}
+}
+
+func (s *Set) Elements() func(yield func(k Value) bool) {
```
jayconrod, line 50: `iter.Seq[Value]` here too?

jayconrod (approved):
Ah, that makes more sense, looks good. And `iterator_test.go` already has `//go:build go1.23` so nothing changes there.

starlark/iter.go
adonovan: Yep! Good catch.

starlark/iter.go
adonovan: ditto

PR #558, opened 2024-07-04
Fix overflow checking for repeat operation

marco6:
The current overflow check for repeat operation fails to check when a multiplication wraps to a positive number.

On 64 bits build it is hard to get one (as you would need a tuple/string with more than 2^32 element - but it's still possible).

On a 32 bits build however it is quite trivial:

```python
# tuples and lists return the wrong count
len((1,) * 641 * 6700417) # returns 1
len([1] * 641 * 6700417) # returns 1

# strings/bytes cause panic: strings: Repeat count causes overflow
"a" * 641 * 6700417
```

This PR uses `bits.Mul` to make sure that this does not happen.

### DIFF
--- a/starlark/eval.go
+++ b/starlark/eval.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"log"
 	"math/big"
+	"math/bits"
 	"sort"
 	"strings"
 	"sync/atomic"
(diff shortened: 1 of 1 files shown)


adonovan (approved):
Well spotted! Many thanks for contributing this fix.

PR #560, opened 2024-07-25
Removed dead code for RPAREN.

haberman:
This appears to be dead code, as we have just checked for RPAREN previously


adonovan (approved):
Well spotted! Thanks for this simplification.

PR #567, opened 2024-11-25
Don't crash when freezing a struct with a circular reference

### LINKED ISSUE #566: starlarkstruct Freeze() can cause a stack overflow if circular references are involved
If `starlarkstruct.Struct` is exposed to the starlark environment, and one creates a struct with a circular reference, and causes it to be frozen, then there will be a stack overflow.

A module with this Starlark content should reproduce the issue, but I haven't double-checked that it actually does:

```python
def make_struct():
    items = []
    foo = struct(items=items)
    foo.items.append(foo)
    return foo

toplevel = make_struct()
```

A fix for this issue could be done as follows:

```diff
--- a/starlarkstruct/struct.go
+++ b/starlarkstruct/struct.go
@@ -99,6 +99,7 @@ func FromStringDict(constructor starlark
 type Struct struct {
        constructor starlark.Value
        entries     entries // sorted by name
+       frozen      bool
 }
 
 // Default is the default constructor for structs.
@@ -172,8 +173,11 @@ func (s *Struct) Hash() (uint32, error) 
        return x, nil
 }
 func (s *Struct) Freeze() {
-       for _, e := range s.entries {
-               e.value.Freeze()
+       if !s.frozen {
+               s.frozen = true
+               for _, e := range s.entries {
+                       e.value.Freeze()
+               }
        }
 }
 ```

### PR DESCRIPTION
**Pull request by oprypin**
The Freeze method needs to check for a `frozen` condition just like all the other types of values, otherwise there can be infinite recursion.

The test that I'm adding here would crash without the fix.

* Fixes #566

### DIFF
--- a/starlarkstruct/struct.go
+++ b/starlarkstruct/struct.go
@@ -99,6 +99,7 @@ func FromStringDict(constructor starlark.Value, d starlark.StringDict) *Struct {
 type Struct struct {
 	constructor starlark.Value
 	entries     entries // sorted by name
+	frozen      bool
 }
 
 // Default is the default constructor for structs.
@@ -172,8 +173,11 @@ func (s *Struct) Hash() (uint32, error) {
 	return x, nil
 }
 func (s *Struct) Freeze() {
-	for _, e := range s.entries {
-		e.value.Freeze()
+	if !s.frozen {
+		s.frozen = true
+		for _, e := range s.entries {
+			e.value.Freeze()
+		}
 	}
 }
 
--- a/starlark/value.go
+++ b/starlark/value.go
@@ -95,6 +95,11 @@ type Value interface {
 	// structure through this API will fail dynamically, making the
 	// data structure immutable and safe for publishing to other
 	// Starlark interpreters running concurrently.
+	//
+	// Implementations of Freeze must be defensive against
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: starlarkstruct/struct.go [hunk 1]
```diff
@@ -172,8 +173,11 @@ func (s *Struct) Hash() (uint32, error) {
 	return x, nil
 }
 func (s *Struct) Freeze() {
-	for _, e := range s.entries {
-		e.value.Freeze()
+	if !s.frozen {
```
adonovan, line 176: In value.go, could you also make this change:

```diff
	// Freeze causes the value, and all values transitively
	// reachable from it through collections and closures, to be
	// marked as frozen.  All subsequent mutations to the data
	// structure through this API will fail dynamically, making the
	// data structure immutable and safe for publishing to other
	// Starlark interpreters running concurrently.
	//
+	// Implementations of Freeze must be defensive against
+	// reference cycles; this can be achieved by first checking
+	// the value's frozen state, then setting it, and only then
+	// visiting any other values that it references.
	Freeze()
```

Thanks.

adonovan (commented):
Thanks! I checked that no other Freeze implementations have this problem.

adonovan (approved):
Thanks again!

Issue #569, opened 2024-12-04
Add support for "_" special variable in REPL mode to store last expression value


EastSPIRA:
Similar to Python's REPL where "_" holds the value of the last expression, it would be valuable to have this functionality in Starlark's REPL mode.

Example of desired behavior:
```python
>>> 2 + 3
5
>>> _
5
>>> _ * 2
10
>>> _
10
```

Benefits:
- Improves interactive debugging and experimentation workflow
- Provides consistency with Python's REPL behavior, making it more intuitive for Python developers
- Useful for exploring computation results without explicitly assigning variables

Implementation considerations:
- Only needs to be active in REPL mode
- Should be updated after each complete expression evaluation
- Should not affect regular script execution outside REPL

The change would primarily affect the REPL implementation in cmd/starlark/main.go while maintaining Starlark's core principles of being a configuration language.

adonovan:
I had never noticed that `_` worked that way in the Python REPL.

Sure, send me a PR.

vt128:
@adonovan @EastSPIRA Please take a look at my PR #570

adonovan:
Thanks!

PR #570, opened 2024-12-26
Add support for "_" special variable in REPL mode to store last expression value

vt128:
This Pull Request implements the feature request to support the `_` special variable in REPL mode, which stores the value of the last evaluated expression. This enhancement aligns the Starlark REPL behavior more closely with Python's REPL, improving the interactive experience for users. Fixed issue #569.

**Changes Made:**

- In `repl.go`, after each successful expression evaluation, the result is stored in the `_` variable within the `globals` dictionary.
- The `_` variable is only updated when an expression is evaluated, not after executing statements.
- If an error occurs during expression evaluation, the `_` variable remains unchanged.
- No changes are made to script execution outside of REPL mode.

**Benefits:**

- Improves interactive debugging and experimentation workflow.
- Provides consistency with Python's REPL behavior, making it more intuitive for Python developers.
- Allows users to easily access the result of the last expression without explicitly assigning it to a variable.

**Example Usage:**

```
>>> 2 + 3
5
>>> _
5
>>> _ * 2
10
>>> _
10
```

**Testing:**

- Verified that the `_` variable correctly stores the value of the last evaluated expression.
- Ensured that `_` remains unchanged after executing statements or if an error occurs during evaluation.
- Confirmed that accessing `_` before any expression has been evaluated results in a `NameError`, consistent with Python's REPL.

### DIFF
--- a/repl/repl.go
+++ b/repl/repl.go
@@ -123,6 +123,9 @@ func rep(opts *syntax.FileOptions, rl *readline.Instance, thread *starlark.Threa
 			return nil
 		}
 
+		// store the result in "_" variable to hold the value of last expression, similar to Python REPL
+		globals["_"] = v
+
 		// print
 		if v != starlark.None {
 			fmt.Println(v)


adonovan (approved):
LGTM. Thanks!

PR #573, opened 2025-01-27
Implement set.update method

andponlin-canva:
The method [update](https://github.com/bazelbuild/starlark/blob/master/spec.md#setupdate) on the type `set` is available in the Bazel implementation of Starlark, but unfortunately not this Go implementation. This PR will augment the `set` implementation to also support the `update` method.

### DIFF
--- a/doc/spec.md
+++ b/doc/spec.md
@@ -3774,6 +3774,19 @@ y = set([2, 3])
 x.union(y)                              # set([1, 2, 3])
 ```
 
+<a id='set·update'></a>
+### set·update
+
+`S.update(iterable)` will have the effect of adding the members
+of the supplied iterable to the set. If no argument is provided
+then the method call will have no effect. The method will return
+`None`.
+
+```python
+x = set([1, 2])
+x.update([2, 3])
+```
+
 <a id='string·elem_ords'></a>
 ### string·elem_ords
 
--- a/doc/spec.md
+++ b/doc/spec.md
@@ -839,7 +839,7 @@ A set has these methods:
 * [`remove`](#set·remove)
 * [`symmetric_difference`](#set·symmetric_difference)
 * [`union`](#set·union)
-
+* [`update`](#set·update)
 
 A set used in a Boolean context is considered true if it is non-empty.
 
@@ -3777,14 +3777,12 @@ x.union(y)                              # set([1, 2, 3])
 <a id='set·update'></a>
 ### set·update
 
-`S.update(iterable)` will have the effect of adding the members
-of the supplied iterable to the set. If no argument is provided
-then the method call will have no effect. The method will return
-`None`.
+`S.update(iterable...)` adds to S each element of the iterable
+sequences. The method will return `None`.
 
 ```python
 x = set([1, 2])
-x.update([2, 3])
+x.update([2, 3], [4, 5])
 ```
 
 <a id='string·elem_ords'></a>
--- a/starlark/library.go
+++ b/starlark/library.go
@@ -151,6 +151,7 @@ var (
 		"remove":               NewBuiltin("remove", set_remove),
 		"symmetric_difference": NewBuiltin("symmetric_difference", set_symmetric_difference),
 		"union":                NewBuiltin("union", set_union),
+		"update":               NewBuiltin("update", set_update),
 	}
 )
 
@@ -2347,6 +2348,22 @@ func set_union(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error)
 		return nil, nameErr(b, err)
 	}
 	return union, nil
+}
+
+// https://github.com/google/starlark-go/blob/master/doc/spec.md#set·update.
+func set_update(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error) {
+	var iterable Iterable
+	if err := UnpackPositionalArgs(b.Name(), args, kwargs, 0, &iterable); err != nil {
+		return nil, err
+	}
+	if iterable != nil {
+		iter := iterable.Iterate()
+		defer iter.Done()
+		if err := b.Receiver().(*Set).InsertAll(iter); err != nil {
+			return None, nameErr(b, err)
+		}
+	}
+	return None, nil
 }
 
 // Common implementation of string_{r}{find,index}.
--- a/starlark/library.go
+++ b/starlark/library.go
@@ -2358,25 +2358,17 @@ func set_update(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error
 
 	receiverSet := b.Receiver().(*Set)
 
-	// factored out into a function so that the `defer` is scoped to this iterable
-	// rather than all iterables in the for loop below.
-	insertIterable := func(iterable Iterable) error {
-		iter := iterable.Iterate()
-		defer iter.Done()
-		if err := receiverSet.InsertAll(iter); err != nil {
-			return nameErr(b, err)
-		}
-		return nil
-	}
-
 	for i, arg := range args {
-		switch arg := arg.(type) {
-		case Iterable:
-			if err := insertIterable(arg); err != nil {
-				return nil, nameErr(b, err)
-			}
-		default:
+		iterable, ok := arg.(Iterable)
+		if !ok {
 			return nil, fmt.Errorf("update: argument #%d is not iterable: %s", i+1, arg.Type())
+		}
+		if err := func() error {
+			iter := iterable.Iterate()
+			defer iter.Done()
+			return receiverSet.InsertAll(iter)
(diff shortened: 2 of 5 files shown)


doc/spec.md
```diff
@@ -3774,6 +3774,19 @@ y = set([2, 3])
 x.union(y)                              # set([1, 2, 3])
 ```
 
+<a id='set·update'></a>
+### set·update
+
+`S.update(iterable)` will have the effect of adding the members
```
adonovan, line 3780: Present tense: "has" (throughout)

Shorter: "update adds the members..."

Even more concisely: "S.update(iterable) adds to S each element of the iterable sequence."

But I think the behavior should be: "S.update(iterable...) adds to S all the elements of each iterable sequence provided as a positional argument"

That's what Python3 does, and Starkark/Java, and it's consistent with dict.update.

starlark/library.go
```diff
@@ -2349,6 +2350,22 @@ func set_union(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error)
 	return union, nil
 }
 
+// https://github.com/google/starlark-go/blob/master/doc/spec.md#set·update.
+func set_update(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error) {
+	var iterable Iterable
+	if err := UnpackPositionalArgs(b.Name(), args, kwargs, 0, &iterable); err != nil {
```
adonovan, line 2356: I think update should accept any number of Iterable arguments. So, reject kwargs, then loop over args testing that each one is iterable; lines 2360--2364 will form the loop body.

starlark/library.go
```diff
@@ -2349,6 +2350,22 @@ func set_union(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error)
...
+		return nil, err
+	}
+	if iterable != nil {
+		iter := iterable.Iterate()
+		defer iter.Done()
+		if err := b.Receiver().(*Set).InsertAll(iter); err != nil {
+			return None, nameErr(b, err)
```
adonovan, line 2363: return nil, nameErr

adonovan (commented):
Thanks for this fix!

doc/spec.md
andponlin-canva: I've updated that and also covered that the method would accept a number of positional arguments which are iterable.

starlark/library.go
```diff
@@ -2352,17 +2352,25 @@ func set_union(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error)
 
 // https://github.com/google/starlark-go/blob/master/doc/spec.md#set·update.
 func set_update(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error) {
-	var iterable Iterable
-	if err := UnpackPositionalArgs(b.Name(), args, kwargs, 0, &iterable); err != nil {
-		return nil, err
+	if len(kwargs) > 0 {
+		return nil, nameErr(b, "kwargs disallowed")
```
adonovan, line 2356: Please use wording consistent with other functions:

		return nil, fmt.Errorf("update does not accept keyword arguments")

starlark/library.go
```diff
@@ -2352,17 +2352,25 @@ func set_union(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error)
...
-		defer iter.Done()
-		if err := b.Receiver().(*Set).InsertAll(iter); err != nil {
-			return None, nameErr(b, err)
+
+	for argIdx, arg := range args {
+		switch v := arg.(type) {
+		case Iterable:
+			iter := v.Iterate()
+			defer iter.Done()
+			if err := b.Receiver().(*Set).InsertAll(iter); err != nil {
```
adonovan, line 2364: Let's hoist `recv := b.Receiver().(*Set)` outside the loop.

starlark/library.go
andponlin-canva: Thanks for pointing that out; it was actually covered in the original [spec](https://github.com/bazelbuild/starlark/blob/master/spec.md#setupdate) as well. I've updated the logic to allow for an arbitrary number of positional arguments which must be iterable.

starlark/library.go
```diff
@@ -2352,17 +2352,25 @@ func set_union(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error)
...
+				return nil, nameErr(b, err)
+			}
+		default:
+			return nil, nameErr(b, fmt.Sprintf(
```
adonovan, line 2368: return nil, fmt.Errorf("update: argument #%d is not iterable: %s", i+1, arg.Type())

starlark/library.go
```diff
@@ -2352,17 +2352,25 @@ func set_union(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error)
...
+	if len(kwargs) > 0 {
+		return nil, nameErr(b, "kwargs disallowed")
 	}
-	if iterable != nil {
-		iter := iterable.Iterate()
-		defer iter.Done()
-		if err := b.Receiver().(*Set).InsertAll(iter); err != nil {
-			return None, nameErr(b, err)
+
+	for argIdx, arg := range args {
```
adonovan, line 2359: for i, arg

starlark/library.go
```diff
@@ -2352,17 +2352,25 @@ func set_union(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error)
...
+		switch v := arg.(type) {
```
adonovan, line 2360: arg := arg

starlark/library.go
andponlin-canva: Good point. I have also factored out the logic from this case branch into a separate function because the `defer` is scoped to the `for` loop.

starlark/library.go
```diff
@@ -2353,21 +2353,30 @@ func set_union(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error)
...
-	for argIdx, arg := range args {
-		switch v := arg.(type) {
+	receiverSet := b.Receiver().(*Set)
+
+	// factored out into a function so that the `defer` is scoped to this iterable
+	// rather than all iterables in the for loop below.
+	insertIterable := func(iterable Iterable) error {
+		iter := iterable.Iterate()
+		defer iter.Done()
+		if err := receiverSet.InsertAll(iter); err != nil {
```
adonovan, line 2366: return receiverSet.InsertAll(iter)

Otherwise nameErr will be applied twice.

starlark/library.go
```diff
@@ -2353,21 +2353,30 @@ func set_union(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error)
...
+		return nil
+	}
+
+	for i, arg := range args {
+		switch arg := arg.(type) {
 		case Iterable:
-			iter := v.Iterate()
-			defer iter.Done()
-			if err := b.Receiver().(*Set).InsertAll(iter); err != nil {
+			if err := insertIterable(arg); err != nil {
```
adonovan, line 2375: You can define the function inline, so that the flow of control is easier to read:

```
	for i, arg := range args {
		iterable, ok := arg.(Iterable)
		if !ok {
			return nil, fmt.Errorf("update: argument #%d is not iterable: %s", i+1, arg.Type())
		}
	 	if err := func() error {
			iter := iterable.Iterate()
			defer iter.Done()
			return receiverSet.InsertAll(iter)
		}(); err != nil {
			return nil, nameErr(b, err)
		}
	}
```

(It's also not necessary to use a switch for a single case.)

starlark/library.go
```diff
@@ -2353,21 +2353,30 @@ func set_union(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error)
...
 	if len(kwargs) > 0 {
-		return nil, nameErr(b, "kwargs disallowed")
+		return nil, nameErr(b, "update does not accept keyword arguments")
 	}
 
-	for argIdx, arg := range args {
-		switch v := arg.(type) {
+	receiverSet := b.Receiver().(*Set)
+
+	// factored out into a function so that the `defer` is scoped to this iterable
```
adonovan, line 2361: Good point; but the comment isn't necessary as the code speaks for itself.

adonovan (commented):
Thanks again.

starlark/library.go
andponlin-canva: OK; I've simply shifted to that structure. above.

starlark/library.go
andponlin-canva: The fact that the `defer` is in the function is obvious but the rationale is probably not so obvious.

adonovan: It's fairly common to call a function literal to reduce the scope of a defer.

adonovan (approved):

PR #574, opened 2025-01-28
correct set.union method

andponlin-canva:
The `union` function on `set` is inconsistent with the behaviour of `update` as it does not support multiple iterable positional arguments as is the case in the Bazel specification of the language. This PR will align `starlark-go` with the Bazel spec.

Note that the `set.union` method no longer uses the `Union` function defined in `value.go` in order to avoid making a new `set` instance for each interable processed.

### DIFF
--- a/doc/spec.md
+++ b/doc/spec.md
@@ -3765,7 +3765,8 @@ x.symmetric_difference([3, 4, 5])         # set([1, 2, 4, 5])
 `S.union(iterable...)` returns a new set into which have been inserted
 all the elements of set S and each element of the iterable sequences.
 
-`union` fails if any element of the iterable is not hashable.
+`union` fails if any argument is not an iterable sequence, or if any
+sequence element is not hashable.
 
 ```python
 x = set([1, 2])
--- a/doc/spec.md
+++ b/doc/spec.md
@@ -3762,9 +3762,8 @@ x.symmetric_difference([3, 4, 5])         # set([1, 2, 4, 5])
 <a id='set·union'></a>
 ### set·union
 
-`S.union(iterable)` returns a new set into which have been inserted
-all the elements of set S and all the elements of the argument, which
-must be iterable.
+`S.union(iterable...)` returns a new set into which have been inserted
+all the elements of set S and each element of the iterable sequences.
 
 `union` fails if any element of the iterable is not hashable.
 
--- a/starlark/library.go
+++ b/starlark/library.go
@@ -2337,17 +2337,27 @@ func set_symmetric_difference(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple)
 
 // https://github.com/google/starlark-go/blob/master/doc/spec.md#set·union.
 func set_union(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error) {
-	var iterable Iterable
-	if err := UnpackPositionalArgs(b.Name(), args, kwargs, 0, &iterable); err != nil {
-		return nil, err
+	if len(kwargs) > 0 {
+		return nil, nameErr(b, "update does not accept keyword arguments")
 	}
-	iter := iterable.Iterate()
-	defer iter.Done()
-	union, err := b.Receiver().(*Set).Union(iter)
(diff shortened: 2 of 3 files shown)


starlark/library.go
```diff
@@ -2337,17 +2337,27 @@ func set_symmetric_difference(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple)
 
 // https://github.com/google/starlark-go/blob/master/doc/spec.md#set·union.
 func set_union(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error) {
-	var iterable Iterable
-	if err := UnpackPositionalArgs(b.Name(), args, kwargs, 0, &iterable); err != nil {
-		return nil, err
+	if len(kwargs) > 0 {
+		return nil, nameErr(b, "update does not accept keyword arguments")
```
adonovan, line 2341: "union..."

starlark/library.go
```diff
@@ -2337,17 +2337,27 @@ func set_symmetric_difference(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple)
...
-	union, err := b.Receiver().(*Set).Union(iter)
-	if err != nil {
-		return nil, nameErr(b, err)
+
+	resultSet := b.Receiver().(*Set).clone()
+
+	for i, arg := range args {
+		iterable, ok := arg.(Iterable)
+		if !ok {
+			return nil, fmt.Errorf("update: argument #%d is not iterable: %s", i+1, arg.Type())
```
adonovan, line 2349: union

starlark/library.go
```diff
@@ -2337,17 +2337,27 @@ func set_symmetric_difference(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple)
...
+		}
+		if err := func() error {
+			iter := iterable.Iterate()
+			defer iter.Done()
+			return resultSet.InsertAll(iter)
+		}(); err != nil {
+			return nil, err
```
adonovan, line 2356: This should return nameErr, like set_update.

But instead let's factor the common code of set_update and set_union since they only differ by whether the destination is a clone:

```
func set_union(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error) {
	if len(kwargs) > 0 {
		return nil, nameErr(b, "union does not accept keyword arguments")
	}
	res := b.Receiver().(*Set).clone()
  	if err := setUpdate(res, args); err != nil {
  	  	return nil, nameErr(b, err)
  	}
  	return res, nil
}

func set_update(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error) {
	if len(kwargs) > 0 {
		return nil, nameErr(b, "update does not accept keyword arguments")
	}
	recv := b.Receiver().(*Set)
   	if err := setUpdate(res, args); err != nil {
  	  	return nil,nameErr(b, err)
  	}
  	return None, nil
}
```

doc/spec.md
```diff
@@ -3762,9 +3762,8 @@ x.symmetric_difference([3, 4, 5])         # set([1, 2, 4, 5])
 <a id='set·union'></a>
 ### set·union
 
-`S.union(iterable)` returns a new set into which have been inserted
-all the elements of set S and all the elements of the argument, which
-must be iterable.
+`S.union(iterable...)` returns a new set into which have been inserted
+all the elements of set S and each element of the iterable sequences.
 
 `union` fails if any element of the iterable is not hashable.
```
adonovan, line 3768: I think this should  read "if any argument is not an iterable sequence, or if any sequence element is not hashable."

adonovan (commented):
Well spotted! Thanks for the fix.

starlark/library.go
andponlin-canva: Ah true; I will do that...

andponlin-canva: I've also factored-out the check for the `kwargs` because the `update` / `union` prefix can be added later albeit with an additional colon in the message. This seems OK to me; let me know what you think.

adonovan (approved):
LGTM. Thanks!

PR #576, opened 2025-02-05
internal/compile: build index of local vars

adonovan:
This eliminates the linear scan over locals in the interpreter when processing named arguments. This reduces the new benchmark from 1081ns to 843ns (-22%).

(This was inspired by recent work in the Java-based implementation, but findParams seemed to be a bigger hit in that case, possibly because of the two extra levels of indirection involved in the length comparison (hot path) of a Java array of Java strings, versus Go where they are all in the same cache line.)

### DIFF
--- a/internal/compile/compile.go
+++ b/internal/compile/compile.go
@@ -345,6 +345,8 @@ type Funcode struct {
 
 	lntOnce sync.Once
 	lnt     []pclinecol // decoded line number table
+
+	localsMap map[string]int
 }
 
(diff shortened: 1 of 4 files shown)


tetromino (commented):
LGTM for the idea (just yesterday I thought of adding the same map to the java starlark interpreter in MethodDescriptor :) but I'm probably not qualified to judge the go implementation

jayconrod (approved):

Issue #584, opened 2025-03-18
Bring sets in line with language spec and enable by default


tetromino:
Now that sets are part of the Starlark language spec (see https://github.com/bazelbuild/starlark/pull/290), we should bring starlark-go's implementation in line with the spec and enable them by default.

Specifically:

* disable comparison operator for sets (or make it flag-guarded?)
* require both sides of `|`, `&`, `-`, and `^` operators (and their augmented forms) to be sets if lhs is a set
* add multi-argument form of `union()`, `intersection()`, `difference()`
* add `update()`, `isdisjoint()`, `intersection_update()`, `difference_update()`, `symmetric_difference_update()`

@aranguyen @brandjon FYI

adonovan:
I agree, and have been meaning to do this (though would welcome a CL). However, I'm worried that it might be a breaking change for some users, so I propose that we use the existing `-set` flag to mean "legacy set behavior" and the absence of that flag to mean "standard set behavior". Then we can deprecate the flag.

tetromino:
I'd suggest flag-guarding the differences in operator behavior (comparison, `|`, etc.), but not the differences in behavior for no-op mutation attempts on frozen sets (which seems like an edge case).

adonovan:
I agree the frozen thing is an edge case we can ignore.

After further reflection, it's not clear to me how to implement flag guarding for the relevant operators since the flag is conceptually a compile-time (and thus per-file) setting, not a per data structure setting nor a per thread setting. Operators don't have access to the files from which they came (nor the thread that calls them, for that matter); also, binary operators and functions have to make a choice about what to do in the (admittedly unlikely) event that exactly one but not both of their operands is covered by the flag semantics. (The interesting case of `x|y` is when only the left operand is a set, perhaps finessing that question.)

I'm not sure we can do much more than emit a dynamic warning message: the first time a deprecated operation is called we log a message.

brandjon:
In Bazel this is gated with `BuildLanguageOptions` / `StarlarkSemantics`, which is more or less global to the invocation. Why can't the Go interpreter access this as global (per-interpreter) state? Or in the worst case, truly global (per-binary) state? That would be enough to guard against backsliding after an LSC within Google. Not sure what the backward compatibility concerns are for other applications.

If you really do need to make it per-file or per-data structure, what about creating an internal second variant of the set datatype and having the flag gate which version you get via set-producing operations (in particular, the `set()` builtin)?

adonovan:
[...]

PR #586, opened 2025-03-26
Support complete collection of set methods, including multi-argument forms

tetromino:
Also update the docs and tests.

Working towards https://github.com/google/starlark-go/issues/584

### DIFF
--- a/starlark/library.go
+++ b/starlark/library.go
@@ -2223,7 +2223,7 @@ func set_clear(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error)
 // https://github.com/google/starlark-go/blob/master/doc/spec.md#set·difference.
 func set_difference(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error) {
 	accum := b.Receiver().(*Set).clone()
-	if err := forIterableArgsDo(args, kwargs, accum.DifferenceUpdate); err != nil {
+	if err := forIterableArgsIterateAndDo(args, kwargs, accum.DifferenceUpdate); err != nil {
 		return nil, nameErr(b, err)
 	}
 	return accum, nil
@@ -2236,7 +2236,7 @@ func set_difference_update(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (V
 		// even if args is empty
 		return nil, nameErr(b, err)
 	}
-	if err := forIterableArgsDo(args, kwargs, recv.DifferenceUpdate); err != nil {
+	if err := forIterableArgsIterateAndDo(args, kwargs, recv.DifferenceUpdate); err != nil {
 		return nil, nameErr(b, err)
 	}
 	return None, nil
@@ -2245,15 +2245,7 @@ func set_difference_update(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (V
 // https://github.com/google/starlark-go/blob/master/doc/spec.md#set_intersection.
 func set_intersection(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error) {
 	accum := b.Receiver().(*Set).clone()
-	// TODO: use an intersection primitive that preserves lhs iteration order
-	if err := forIterableArgsDo(args, kwargs, func(other Iterator) error {
-		result, error := accum.Intersection(other)
-		if error != nil {
-			return error
-		}
-		accum = result.(*Set)
-		return nil
-	}); err != nil {
+	if err := forIterableArgsDo(args, kwargs, accum.IntersectionUpdate); err != nil {
 		return nil, nameErr(b, err)
 	}
 	return accum, nil
@@ -2266,26 +2258,8 @@ func set_intersection_update(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) 
 		// even if args is empty
 		return nil, nameErr(b, err)
 	}
-	var accum *Set = nil
-	// TODO: use an intersection primitive that preserves lhs iteration order
-	if err := forIterableArgsDo(args, kwargs, func(other Iterator) error {
-		if accum == nil {
-			accum = recv
-		}
-		result, error := accum.Intersection(other)
-		if error != nil {
-			return error
-		}
-		accum = result.(*Set)
-		return nil
-	}); err != nil {
+	if err := forIterableArgsDo(args, kwargs, recv.IntersectionUpdate); err != nil {
 		return nil, nameErr(b, err)
-	}
-	if accum != nil {
-		recv.Clear()
-		accumIter := accum.Iterate()
-		defer accumIter.Done()
-		recv.InsertAll(accumIter)
 	}
 	return None, nil
 }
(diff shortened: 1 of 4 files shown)


doc/spec.md
```diff
@@ -3691,15 +3713,45 @@ x                                        # set([1, 3])
...
+### set·intersection\_update
+
+`S.intersection_update(*others)` removes from the set `S` any elements not found
+in at least one of the iterable sequences `*others`. It returns `None`.
+
+It is permissible to call `intersection_update` without any arguments; this
+leaves the set `S` unchanged.
+
+`intersection_update` fails if the set `S` is frozen or has active iterators, or
+if any element of any of the `*others` is unhashable.
```
adonovan, line 3739: For the record, this precludes a short cut when S is empty.

starlark/library.go
```diff
@@ -2463,20 +2542,24 @@ func updateDict(dict *Dict, updates Tuple, kwargs []Tuple) error {
 	return nil
 }
 
-func setUpdate(s *Set, args Tuple, kwargs []Tuple) error {
+// forIterableArgsDo is a helper for implementing methods which perform the
+// same operation on an iteratover over each iterable argument, and don't
```
adonovan, line 2546: iterator

starlark/library.go
```diff
@@ -2463,20 +2542,24 @@ func updateDict(dict *Dict, updates Tuple, kwargs []Tuple) error {
...
+// allow keyword arguments - for example, set.update(*others),
+// set.union(*others), etc.
+func forIterableArgsDo(operation func(other Iterator) error, args Tuple, kwargs []Tuple) error {
```
adonovan, line 2549: It's common style in Go to put the func parameter last so that it can be called like so:
```
forIterableArgsDo(args, kwarks, func() {
     ...
})
```

Of course, the error result makes this slightly less tidy, but I think it's still the preferred order.

starlark/value.go
```diff
@@ -1221,20 +1221,18 @@ func (s *Set) clone() *Set {
...
-			return nil, err
-		}
+	if err := set.InsertAll(other); err != nil {
+		return nil, err
 	}
 	return set, nil
 }
 
-func (s *Set) InsertAll(iter Iterator) error {
+// TODO: rename to Update()
```
adonovan, line 1232: We can't rename it, but you can add Update, make this function a wrapper around it, and add a `//go:fix inline` annotation on InsertAll, which will nudge people to use Update.

starlark/value.go
```diff
@@ -1275,7 +1294,10 @@ func (s *Set) IsSubset(other Iterator) (bool, error) {
 	}
 }
 
+// Warning: the resulting set has the iteration order of other.
```
adonovan, line 1297: Rather than add these Warnings and TODOs, we should fix the underlying problem in this CL.

starlark/value.go
```diff
@@ -1295,17 +1317,33 @@ func (s *Set) Intersection(other Iterator) (Value, error) {
 
 func (s *Set) SymmetricDifference(other Iterator) (Value, error) {
 	diff := s.clone()
+	if err := diff.symmetricDifferenceUpdate(s, other); err != nil {
+		return nil, err
+	}
+	return diff, nil
+}
+
+// s and sReadonlyCopy must contain the same elements; s will be modified,
```
adonovan, line 1326: It's unfortunate that you need to clone s just to be able to iterate over it at the same time as (one of the callers) wishes to update s. Can `set_symmetric_difference_update` not simply call `s.symmetricDifferenceUpdate(s, iter)`? Is the awkwardness designed to handle the edge case `x.op_update(x)` for nonempty sets x? It might be simpler to shortcut in that case: x.symmetric_difference_update(x) becomes just x.clear().

We should add tests of such cases.

starlark/library.go
```diff
@@ -2218,34 +2222,89 @@ func set_clear(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error)
...
+		return nil, nameErr(b, err)
+	}
+	return None, nil
 }
 
 // https://github.com/google/starlark-go/blob/master/doc/spec.md#set_intersection.
 func set_intersection(_ *Thread, b *Builtin, args Tuple, kwargs []Tuple) (Value, error) {
-	// TODO: support multiple others: s.difference(*others)
+	accum := b.Receiver().(*Set).clone()
+	// TODO: use an intersection primitive that preserves lhs iteration order
```
adonovan, line 2248: We should address this in this CL. I think we should explicitly guarantee the insertion order of all these operations.

adonovan (commented):
Thanks for contributing these operators.

starlark/value.go
tetromino: The problem is that `other` can be a list/tuple containing duplicates. Consider something like `set([x]).symmetric_difference([x, x])` - if we didn't have sReadonlyCopy, we would naively iterate over `[x, x]`, delete `x`, re-insert `x`, and end up with `set([x])` as the wrong answer.

So we have two possibilities: either make a temporary copy of the lhs set `s`, as I did here; or else make a temporary set copy of the rhs list `other` to deduplicate.

With the second approach, it makes more sense to pass `other` as a `Value` rather than an iterator, so that we can avoid making a copy if `other` is already deduplicated (e.g. already a set). And that approach also can work for order-preserving intersection. WDYT?

starlark/library.go
tetromino: Thanks for the suggestion, makes the code more readable!

starlark/library.go
tetromino: Fixed

starlark/value.go
adonovan: Ah, interesting example (and definitely worthy of a comment). Still I don't understand how the readonly copy of s solves that problem: we would need to first accumulate the iterator into a set s', and then compute sΔs', otherwise we would need to remember information (parity of occurrence) about all iterator values already seen.

> So we have two possibilities: either make a temporary copy of the lhs set s, as I did here; or else make a temporary set copy of the rhs list other to deduplicate.

The second approach seems conceptually simpler.

starlark/value.go
tetromino: Done.

starlark/value.go
tetromino: Done.

starlark/value.go
tetromino: Done.

starlark/library.go
tetromino: Done.

doc/spec.md
tetromino: That seems reasonable? We usually prohibit no-op mutation attempts - for example, if a set is immutable, we don't allow trying to update it with an element that it already contains.

tetromino:
I've added a new Set.IntersectionUpdate() which preserves lhs iteration order - please take a look. Notably, since IntersectionUpdate needs to delete from the lhs hash table whilst iterating over it, it needs to iterate manually instead of using a normal iterator. I believe this is sound as long as the entry being deleted is an entry that has already been examined, and as long as we save the next pointer (since deletion will mutate the entry struct).

PR #587, opened 2025-04-02
fix: prevent race condition in List.Slice method

thevilledev:
## Description

This PR addresses a race condition in the list slicing operation (`list[start:end:step]`). When multiple Starlark threads operate concurrently on the same shared, mutable list, a slicing operation can lead to unexpected panics when another thread modifies the list during slicing.

Feedback welcome!

## Problem

Starlark's specification and implementation documentation explicitly specify that collections should be protected from concurrent modification during iteration. As noted in `doc/impl.md` under the "Fail-fast iterators" section:

> Instead of mutation of the collection invalidating the iterator, the act of iterating makes the collection temporarily immutable, so that an attempt to, say, delete a dict element while looping over the dict, will fail.

> This is implemented by having each mutable iterable value record a counter of active iterators. Starting a loop increments this counter, and completing a loop decrements it. A collection with a nonzero counter behaves as if frozen.

While this protection mechanism is correctly implemented for most operations, the List.Slice method was missing this crucial protection. Since slicing iterates through list elements, it should follow the same "fail-fast iterator" pattern as other iteration operations.

## Fix

This PR modifies the List.Slice method to:

- Increment the list's itercount at the start of slicing operations
- Decrement it when the operation completes (using a deferred function)

This makes slicing operations consistent with other list operations that iterate over elements, ensuring that attempts to modify a list while it's being sliced will fail with a clear error message instead of potentially causing a panic.

## References

I have a simple test application to validate this at https://go.dev/play/p/1CBhyyfJom0.

- Output without the proposed fix: `PANIC in Thread A (Slicing): runtime error: index out of range [724994] with length 1`
- Output with the proposed fix: `Thread B returned error: Error in clear: clear: cannot clear list during iteration`

Originally reported to Google OSS VRP.

### DIFF
--- a/starlark/value.go
+++ b/starlark/value.go
@@ -952,15 +952,32 @@ func (l *List) Len() int              { return len(l.elems) }
 func (l *List) Index(i int) Value     { return l.elems[i] }
 
 func (l *List) Slice(start, end, step int) Value {
+	// For slice operations, we don't need to check if the list is "mutable"
+	// since slice doesn't modify the original list. However, we need to
+	// increment the itercount to prevent other threads from modifying the list
+	// while we're iterating through it.
+	if !l.frozen {
+		l.itercount++
+		defer func() { l.itercount-- }()
+	}
+
+	// Create a defensive copy to safely access elements
+	elems := make([]Value, len(l.elems))
+	copy(elems, l.elems)
+
 	if step == 1 {
-		elems := append([]Value{}, l.elems[start:end]...)
-		return NewList(elems)
+		resultSlice := elems[start:end]
+		resultCopy := make([]Value, len(resultSlice))
+		copy(resultCopy, resultSlice)
+		return NewList(resultCopy)
 	}
 
 	sign := signum(step)
 	var list []Value
 	for i := start; signum(end-i) == sign; i += step {
-		list = append(list, l.elems[i])
+		if i < len(elems) { // Add bounds check for extra safety
+			list = append(list, elems[i])
+		}
 	}
 	return NewList(list)
 }
--- a/starlark/value_test.go
+++ b/starlark/value_test.go
@@ -8,7 +8,9 @@ package starlark_test
 
 import (
 	"fmt"
+	"sync"
 	"testing"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
 	"go.starlark.net/starlark"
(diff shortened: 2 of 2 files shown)


adonovan:
> ## Description
> This PR addresses a race condition in the list slicing operation (`list[start:end:step]`). When multiple Starlark threads operate concurrently on the same shared, mutable list, a slicing operation can lead to unexpected panics when another thread modifies the list during slicing.

By design, you should never get to a situation in which two Starlark threads have access to the same mutable value and that value is not frozen. An application that reaches this situation is using Starlark incorrectly. 

> ## Problem
> Starlark's specification and implementation documentation explicitly specify that collections should be protected from concurrent modification during iteration. As noted in `doc/impl.md` under the "Fail-fast iterators" section: [...]

The quoted section relates to single-threaded programs, and disallows mutating a list that is currently being iterated by that same thread, similar to the dynamic checks in Java's java.util package that result in a ConcurrentModificationException. (The word concurrent in this context is misleading because it refers to two operations within a single thread.)


> While this protection mechanism is correctly implemented for most operations, the List.Slice method was missing this crucial protection. Since slicing iterates through list elements, it should follow the same "fail-fast iterator" pattern as other iteration operations.

List.Slice is implemented correctly. It's a read-only operation, and thus safe for concurrent use with other read-only operations. Starlark data structures are mutable only while confined to their originating thread, so data races are eliminated by design.

thevilledev:
Thanks for the feedback. I understand the intended concurrency model relies on freezing post-initialization, as outlined in the docs (`spec.md`/`impl.md`).

The core issue this PR addresses is that `List.Slice` currently causes a Go panic if that model is violated by the host app.

You're correct that `impl.md` ("Fail-fast iterators") describes the `itercount` mechanism primarily in the context of preventing mutation by the *same* thread during iteration. My rationale is that `List.Slice` performs an iteration over list elements based on its length at the start. Applying the *existing* `itercount` protection mechanism to `Slice` makes it consistent with other iterative operations and robust against concurrent modification *from any source* during that iteration. In return, a more resilient library.

If it is considered "by design", could we consider a documentation update? We could explicitly add a note to `spec.md` or `impl.md` warning against sharing unfrozen mutable values between threads.

adonovan:
> The core issue this PR addresses is that List.Slice currently causes a Go panic if that model is violated by the host app.

I'm glad to hear that it panics, and thus helped reveal a data race in your application. Silently doing something different or returning an error would have made your bug harder to find.

> My rationale is that List.Slice performs an iteration over list elements based on its length at the start.

List.Slice is just an accessor that looks at the current value of the array and returns a new value; in that respect it's not fundamentally different from List.Index, or any number of trivial accessors. The only difference is that it reads l.elems multiple times, so if you have a data race, you are more likely to notice it. But racing calls to List.Index are also not safe: for example, you could read the two words of a single array element (Value) at different moments, giving you a corrupted value.

This code doesn't involve Starlark iterators at all (or Go iterators, for that matter), so itercount is entirely the wrong mechanism. It's just making the problem harder to see.

>  Applying the existing itercount protection mechanism to Slice makes it consistent with other iterative operations and robust against concurrent modification from any source during that iteration. In return, a more resilient library.

Neither of these is true, as I've explained above.

> If it is considered "by design", could we consider a documentation update? We could explicitly add a note to spec.md or impl.md warning against sharing unfrozen mutable values between threads.

I'd be happy to clarify the docs, but I'm not sure what information is missing. Here's the current wording:

https://github.com/google/starlark-go/blob/master/doc/spec.md#freezing-a-value:
> Freezing a value
> Starlark has a feature unusual among imperative programming languages: a mutable value may be frozen so that all subsequent attempts to mutate it fail with a dynamic error; the value, and all other values reachable from it, become immutable.
> 
> Immediately after execution of a Starlark module, all values in its top-level environment are frozen. Because all the global variables of an initialized Starlark module are immutable, the module may be published to and used by other threads in a parallel program without the need for locks. For example, the Bazel build system loads and executes BUILD and .bzl files in parallel, and two modules being executed concurrently may freely access variables or call functions from a third without the possibility of a race condition.

and, as you point out, doc/impl says:

> Freezing
> 
> All mutable values created during module initialization are _frozen_ upon its completion. It is this property that permits a Starlark module to be referenced by two Starlark threads running concurrently (such as the initialization threads of two other modules) without the possibility of a data race.
>
> The Go implementation supports freezing by storing an additional "frozen" Boolean variable in each mutable object. Once this flag is set, all subsequent attempts at mutation fail. Every value defines a Freeze method that sets its own frozen flag if not already set, and calls Freeze for each value that it contains. For example, when a list is frozen, it freezes each of its elements; when a dictionary is frozen, it freezes each of its keys and values; and when a function value is frozen, it freezes each of the free variables and parameter default values implicitly referenced by its closure. Application-defined types must also follow this discipline.
> 
> The freeze mechanism in the Go implementation is finer grained than in the Java implementation: in effect, the latter has one "frozen" flag per module, and every value holds a reference to the frozen flag of its module. This makes setting the frozen flag more efficient---a simple bit flip, no need to traverse the object graph---but coarser grained. Also, it complicates the API slightly because to construct a list, say, requires a reference to the frozen flag it should use.

If you have specific suggestions for improvements, feel free to report a doc issue and I'll fix it. Thanks.

PR #588, opened 2025-04-17
Export UnpackOneArg

misberner:
This PR turns `unpackOneArg` into an exported function.

The intended use case is leveraging the unpacking logic in custom `Unpacker` implementations (e.g., unpacking into a `string`-based enum with validation by first unpacking into a `*string` using the default logic, then validating the value before converting to a typed enum). Without this, one would either have to re-implement the relevant parts of the `unpackOneArg` logic, or use the workaround of `UnpackPositionalArgs("", starlark.Tuple{arg}, nil, 1, target)` and then stripping the `": for parameter 1: "` prefix of any error returned (which obviously is ugly).

### DIFF
--- a/starlark/unpack.go
+++ b/starlark/unpack.go
@@ -119,7 +119,7 @@ func UnpackArgs(fnname string, args Tuple, kwargs []Tuple, pairs ...any) error {
 				continue
 			}
 		}
-		if err := UnpackOneArg(arg, pairs[2*i+1]); err != nil {
+		if err := UnpackArg(arg, pairs[2*i+1]); err != nil {
 			return fmt.Errorf("%s: for parameter %s: %s", fnname, name, err)
 		}
 	}
@@ -144,7 +144,7 @@ kwloop:
 				}
 
 				ptr := pairs[2*i+1]
-				if err := UnpackOneArg(arg, ptr); err != nil {
+				if err := UnpackArg(arg, ptr); err != nil {
 					return fmt.Errorf("%s: for parameter %s: %s", fnname, name, err)
 				}
 				continue kwloop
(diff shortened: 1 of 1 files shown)


starlark/unpack.go
```diff
@@ -209,14 +209,24 @@ func UnpackPositionalArgs(fnname string, args Tuple, kwargs []Tuple, min int, va
...
-		if err := unpackOneArg(arg, vars[i]); err != nil {
+		if err := UnpackOneArg(arg, vars[i]); err != nil {
 			return fmt.Errorf("%s: for parameter %d: %s", fnname, i+1, err)
 		}
 	}
 	return nil
 }
 
-func unpackOneArg(v Value, ptr any) error {
+// UnpackOneArg unpacks a single argument, represented as a Value, into
```
adonovan, line 219: More concisely:

// UnpackArg unpacks a Value v into the variable pointed to by ptr.
// See [UnpackArgs] for details, including which types of variable are supported.

starlark/unpack.go
```diff
@@ -209,14 +209,24 @@ func UnpackPositionalArgs(fnname string, args Tuple, kwargs []Tuple, min int, va
...
+// the given target variable. The target must be a pointer to a supported
+// type, or implement the Unpacker interface (see UnpackArgs for more
+// details, including what types are supported).
+//
+// UnpackOneArg is used by both UnpackArgs and UnpackPositionalArgs to handle
```
adonovan, line 224: // This function defines the unpack operation for a single value.
// The implementations of most built-in functions use [UnpackArgs] and 
// [UnpackPositionalArgs] to unpack a sequence of positional and/or
// keyword arguments.

adonovan (commented):
Thanks, this looks good. Just a few comment tweaks.

misberner:
Thanks for the quick turnaround! Addressed comments

adonovan (approved):
Thanks again.

PR #589, opened 2025-04-28
feat: add method `CancelWithError()`

jcmfernandes:
I'm interested in knowing if the execution was cancelled, and I don't want to rely on string matching the error message to know it.

### DIFF
--- a/starlark/eval.go
+++ b/starlark/eval.go
@@ -5,6 +5,7 @@
 package starlark
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"log"
@@ -60,7 +61,7 @@ type Thread struct {
 	Steps, maxSteps uint64
 
 	// cancelReason records the reason from the first call to Cancel.
-	cancelReason *string
+	cancelReason error
 
 	// locals holds arbitrary "thread-local" Go values belonging to the client.
 	// They are accessible to the client but not to any Starlark program.
@@ -92,6 +93,10 @@ func (thread *Thread) Uncancel() {
 	atomic.StorePointer((*unsafe.Pointer)(unsafe.Pointer(&thread.cancelReason)), nil)
 }
 
+var (
+	ErrExecutionCancelled = errors.New("Starlark computation cancelled")
+)
+
 // Cancel causes execution of Starlark code in the specified thread to
 // promptly fail with an EvalError that includes the specified reason.
 // There may be a delay before the interpreter observes the cancellation
@@ -102,8 +107,17 @@ func (thread *Thread) Uncancel() {
 // Unlike most methods of Thread, it is safe to call Cancel from any
 // goroutine, even if the thread is actively executing.
 func (thread *Thread) Cancel(reason string) {
+	heapErr := new(error)
+	*heapErr = fmt.Errorf("%w: %s", ErrExecutionCancelled, reason)
 	// Atomically set cancelReason, preserving earlier reason if any.
-	atomic.CompareAndSwapPointer((*unsafe.Pointer)(unsafe.Pointer(&thread.cancelReason)), nil, unsafe.Pointer(&reason))
+	atomic.CompareAndSwapPointer((*unsafe.Pointer)(unsafe.Pointer(&thread.cancelReason)), nil, unsafe.Pointer(heapErr))
+}
+
+func (thread *Thread) CancelWithError(err error) {
+	heapErr := new(error)
+	*heapErr = fmt.Errorf("%w: %w", ErrExecutionCancelled, err)
+	// Atomically set cancelReason, preserving earlier reason if any.
+	atomic.CompareAndSwapPointer((*unsafe.Pointer)(unsafe.Pointer(&thread.cancelReason)), nil, unsafe.Pointer(heapErr))
 }
 
 // SetLocal sets the thread-local value associated with the specified key.
--- a/starlark/eval_test.go
+++ b/starlark/eval_test.go
@@ -946,11 +946,17 @@ func TestCancel(t *testing.T) {
 		if fmt.Sprint(err) != "Starlark computation cancelled: nope" {
 			t.Errorf("execution returned error %q, want cancellation", err)
 		}
+		if !errors.Is(err, starlark.ErrExecutionCancelled) {
+			t.Errorf("execution didn't return error starlark.ErrExecutionCancelled")
+		}
 
 		// cancellation is sticky
 		_, err = starlark.ExecFile(thread, "precancel.star", `x = 1//0`, nil)
 		if fmt.Sprint(err) != "Starlark computation cancelled: nope" {
 			t.Errorf("execution returned error %q, want cancellation", err)
+		}
+		if !errors.Is(err, starlark.ErrExecutionCancelled) {
+			t.Errorf("execution didn't return error starlark.ErrExecutionCancelled")
 		}
 	}
 	// A thread cancelled during a built-in executes no more code.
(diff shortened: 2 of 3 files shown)


starlark/eval.go
```diff
@@ -102,8 +107,17 @@ func (thread *Thread) Uncancel() {
...
 // goroutine, even if the thread is actively executing.
 func (thread *Thread) Cancel(reason string) {
+	heapErr := new(error)
+	*heapErr = fmt.Errorf("%w: %s", ErrExecutionCancelled, reason)
+	// Atomically set cancelReason, preserving earlier reason if any.
+	atomic.CompareAndSwapPointer((*unsafe.Pointer)(unsafe.Pointer(&thread.cancelReason)), nil, unsafe.Pointer(heapErr))
+}
+
+func (thread *Thread) CancelWithError(err error) {
+	heapErr := new(error)
```
adonovan, line 117: Simpler:

err := fmt.Errorf(...)
...&err...

but the fmt.Errorf wrapping is needed only if `!errors.Is(err, ErrCanceled)`.

Also, let's document:

// Wrap the user's error so that errors.Is will
// find both the user's error and ErrCanceled as causes.

starlark/interp.go
```diff
@@ -105,8 +105,8 @@ loop:
 				thread.Cancel("too many steps")
 			}
 		}
-		if reason := atomic.LoadPointer((*unsafe.Pointer)(unsafe.Pointer(&thread.cancelReason))); reason != nil {
-			err = fmt.Errorf("Starlark computation cancelled: %s", *(*string)(reason))
+		if reasonErr := atomic.LoadPointer((*unsafe.Pointer)(unsafe.Pointer(&thread.cancelReason))); reasonErr != nil {
```
adonovan, line 108: cancelErr.Load() etc

starlark/eval.go
```diff
@@ -92,6 +93,10 @@ func (thread *Thread) Uncancel() {
 	atomic.StorePointer((*unsafe.Pointer)(unsafe.Pointer(&thread.cancelReason)), nil)
 }
 
+var (
+	ErrExecutionCancelled = errors.New("Starlark computation cancelled")
```
adonovan, line 97: Let's call it just ErrCanceled (note US spelling), and document it:
```
// When a Starlark function call is canceled by a call to
// [Thread.Cancel] or [Thread.CancelWithError],
// the EvalError returned by [Call] will respond to
// errors.Is(err, ErrCanceled).
```

starlark/eval_test.go
```diff
@@ -966,6 +972,62 @@ func TestCancel(t *testing.T) {
...
+			t.Errorf("execution didn't return expected error")
+		}
+
+		// cancellation is sticky
+		_, err = starlark.ExecFile(thread, "precancel.star", `x = 1//0`, nil)
+		if fmt.Sprint(err) != "Starlark computation cancelled: nope" {
+			t.Errorf("execution returned error %q, want cancellation", err)
+		}
+		if !errors.Is(err, starlark.ErrExecutionCancelled) {
+			t.Errorf("execution didn't return error starlark.ErrExecutionCancelled")
```
adonovan, line 1004: Let's make the assertions more specific about what exactly they tested and what they got, e.g. `"ExecFIle: !error.Is(ErrCanceled): got %T, %v", err, err`

adonovan (commented):
Thanks for this PR.

It is crucial that cancellation preserve the backtrace of the Starlark call stack so that the user can tell what function was interrupted by cancellation. The existing code in Call will cause the leaf error (cancellation) to be wrapped by EvalError, if it's not already an EvalError, so I think it should work correctly. But can you add a test (or modify your current test) so that a stack of calls f->g->h is cancelled while h is running (e.g. counting to infinity), and assert that the resulting error has both (a) a call stack showing f, g, h and responds to errors.Is(err, Cancelled)?

starlark/eval.go
jcmfernandes: I must confess that I wrote "cancelled" because the message says "cancelled". I don't think we want to change the message, as I'm sure that people are string matching at this point... but tell me if you think otherwise.

starlark/eval_test.go
jcmfernandes: Will do so.

starlark/eval.go
```diff
@@ -102,8 +110,17 @@ func (thread *Thread) Uncancel() {
 // Unlike most methods of Thread, it is safe to call Cancel from any
 // goroutine, even if the thread is actively executing.
 func (thread *Thread) Cancel(reason string) {
+	thread.CancelWithError(errors.New(reason))
+}
+
+// Wrap the user's error so that errors.Is will find both the user's error and
```
adonovan, line 116: This comment belongs within the method, since it's an implementation detail.

starlark/eval_test.go
```diff
@@ -946,12 +946,18 @@ func TestCancel(t *testing.T) {
 		if fmt.Sprint(err) != "Starlark computation cancelled: nope" {
 			t.Errorf("execution returned error %q, want cancellation", err)
 		}
+		if !errors.Is(err, starlark.ErrCanceled) {
+			t.Errorf("execution didn't return error starlark.ErrExecutionCancelled")
```
adonovan, line 950: ErrCanceled

(throughout)

starlark/interp.go
```diff
@@ -105,8 +103,8 @@ loop:
 				thread.Cancel("too many steps")
 			}
 		}
-		if reason := atomic.LoadPointer((*unsafe.Pointer)(unsafe.Pointer(&thread.cancelReason))); reason != nil {
-			err = fmt.Errorf("Starlark computation cancelled: %s", *(*string)(reason))
+		if reasonErr := thread.cancelReason.Load(); reasonErr != nil {
+			err = *(*error)(reasonErr)
```
adonovan, line 107: `err = *reasonErr`

starlark/eval.go
adonovan: I think we can afford to change it. We never promised any spelling, and if this causes people to have to update their tests, they might also switch to the more robust approach of using `errors.Is(ErrCanceled, err)`.

starlark/eval_test.go
jcmfernandes: Done.

starlark/eval.go
jcmfernandes: Done.

jcmfernandes:
Hey @adonovan, thanks for the review. I addressed your feedback. Let me know what you think.

starlark/eval_test.go
```diff
@@ -963,8 +971,168 @@ func TestCancel(t *testing.T) {
...
+		if !ok {
+			t.Errorf("ExecFile: expected *starlark.EvalError; got %T, %v", err, err)
+		}
+		expectedBacktace := `
+  cancel.star:12:2: in <toplevel>
+  cancel.star:3:3: in f
+  cancel.star:6:3: in g
+`
+		if !strings.Contains(evalErr.Backtrace(), expectedBacktace) || !regexp.MustCompile(`cancel\.star:(10:\d+|9:\d+): in h`).MatchString(evalErr.Backtrace()) {
+			t.Errorf("ExecFile: expected backtrace to include calls to f(), g(), and h(); got %s", evalErr.Backtrace())
```
jcmfernandes, line 1135: We can, alternatively, inspect the call frames instead of the backtrace string. As you prefer.

starlark/eval_test.go
```diff
@@ -993,7 +1161,7 @@ func TestExecutionSteps(t *testing.T) {
 	// Exceeding the step limit causes cancellation.
 	thread.SetMaxExecutionSteps(1000)
 	_, err = countSteps(1000)
-	if fmt.Sprint(err) != "Starlark computation cancelled: too many steps" {
+	if fmt.Sprint(err) != "Starlark computation canceled: too many steps" {
```
jcmfernandes, line 1164: And maybe here we can check for `ErrCanceled` instead.

starlark/eval_test.go
```diff
@@ -963,8 +971,168 @@ func TestCancel(t *testing.T) {
...
+
+def h():
+	for x in fibonacci:
+		ready()
+
+f()
+`
+		cancelErr := errors.New("nope")
+		go func() {
+			for !ready.Load() {
```
adonovan, line 1111: This is not how to wait for an event; use a semaphore (sync.WaitGroup) or just a channel.

But why not just have the `ready` function cancel its own Thread? The concurrency of polling is not what's being tested here. The Starlark program then reduces to:
```
def f(): g()
def g(): h()
def h(): cancel()
```
and the cancel builtin function becomes:
```
"cancel": starlark.NewBuiltin(...) {
		thread.CancelWithError(cancelErr)
		return starlark.None, nil
}
```

starlark/eval_test.go
adonovan: This looks fine.

starlark/eval_test.go
jcmfernandes: I see value in testing internally (sync?) and externally (async?) triggered cancellations, but I'm happy to change it.

jcmfernandes:
> Please discuss bugs and features in the issue tracker before sending code changes. Thanks.

Can you please clarify what you expect at this point?

adonovan:
> > Please discuss bugs and features in the issue tracker before sending code changes. Thanks.
> Can you please clarify what you expect at this point?

Sure: I expect that before non-trivial code changes, we open an issue in this GitHub repo's issue tracker, describe the problem and the proposed solution, flag any potential difficulties, and come to a consensus on whether and how to fix it. This avoids wasting time in code review, and forms an easily searchable record of what problems exist or have been fixed.

jcmfernandes:
> > > Please discuss bugs and features in the issue tracker before sending code changes. Thanks.
> > > Can you please clarify what you expect at this point?
> 
> Sure: I expect that before non-trivial code changes, we open an issue in this GitHub repo's issue tracker, describe the problem and the proposed solution, flag any potential difficulties, and come to a consensus on whether and how to fix it. This avoids wasting time in code review, and forms an easily searchable record of what problems exist or have been fixed.

Will do so.

PR #591, opened 2025-05-29
Print token value for illegal tokens

misberner:
If an illegal token (typically a Python keyword that isn't supported in Starlark) is encountered, the parser prints it as just `illegal token` in the error message. Without looking at the precise line:col position, this can be very confusing, especially for people coming from a Python background that are new to Starlark.

This PR ensures that in the illegal token case, the token value is printed alongside the `illegal token` string.

Before:
```
print(None is None)  # <script>:1:14: got illegal token, want ','
```
After:
```
print(None is None)  # <script>:1:14: got illegal token "is", want ','
```

### DIFF
--- a/syntax/scan.go
+++ b/syntax/scan.go
@@ -22,7 +22,7 @@ import (
 type Token int8
 
 const (
-	ILLEGAL Token = iota
+	ILLEGAL Token = iota // deprecated: illegal tokens are at the bottom of this block
 	EOF
 
 	NEWLINE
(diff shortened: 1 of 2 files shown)


adonovan (commented):
Thanks for this UI improvement. I wonder if it might not be simpler just to name all the Python tokens and add a comment `// reserved` next to each one. Then the existing token String logic just works.

adonovan:
I agree that some kind of invalid zero value is best, even if we don't refer to it. I don't see your latest patch set though.

misberner:
Sorry. Too many remotes, pushed to the right one now

syntax/scan.go
```diff
@@ -22,7 +22,7 @@ import (
 type Token int8
 
 const (
-	ILLEGAL Token = iota
+	ILLEGAL Token = iota // deprecated: illegal tokens are at the bottom of this block
```
adonovan, line 25: It's not deprecated: it's still the name for zero value of Token, which indicates some kind of mistake. So let's remove this comment.

syntax/scan.go
```diff
@@ -98,6 +98,26 @@ const (
 	RETURN
 	WHILE
 
+	// Illegal tokens (reserved Python keywords that aren't part of the Starlark language)
```
adonovan, line 101: Let's avoid confusion with the name ILLEGAL above and document this simply as:

// Reserved words (following Python); unused in Starlark

then the tokens below don't need individual comments.

adonovan (commented):
Looks good, but a couple of style nits.

misberner:
done

PR #592, opened 2025-05-30
update go to 1.21, deps to latest, and modernize

adonovan:


go.mod
```diff
@@ -1,12 +1,12 @@
 module go.starlark.net
 
-go 1.19
+go 1.23.0
```
jayconrod, line 3: Should be `1.21.0`?

jayconrod (approved):

go.mod
adonovan: Indeed it should; done.

fenollp (approved):

adonovan:
This change is on hold because updating module dependencies causes us to drop support for various go1.x releases, contra our documentation. I think that's an inevitable consequence of the [extended forward compatibility](https://github.com/golang/go/issues/57001) and the fact that all [golang.org/x repos closely follow the latest release](https://github.com/golang/go/issues/69095), but it should not a problem for most clients since the toolchain will transparently do the right thing. But we should probably announce the policy change rather than sneaking it in here.

adonovan (commented):
Thanks for this PR, and sorry for neglecting it for so long. Since running 'go get -u' updates the minimum go version required by dependencies, in this case to go 1.23.0, it violated the support policy in the README file, which promises four versions. I don't think that policy is tenable, so I've relaxed it, applied `go get -u`, and run `go fix`, in https://github.com/google/starlark-go/pull/635.

I'll revert this PR now. But with the new policy, similar PRs will be immediately accepted in future. Thanks again.

PR #595, opened 2025-06-18
starlark: Add (*Function).Predeclared

### LINKED ISSUE #594: Add (*starlark.Function).Predeclared()
👋 I'm writing an interactive debugger for Starlark. I've found some of the previous PRs in the area pretty helpful, including #539. However, it doesn't make sense to me why `(*starlark.Function).Globals()` [exists](https://pkg.go.dev/go.starlark.net/starlark#Function.Globals) but not  `(*starlark.Function).Predeclared()`.

While it's true that the embedder of Starlark provides the predeclared dict in the first place, Starlark allows using a different predeclared dict for every module. It would be a bit onerous to thread the predeclared dict for every module through to the debugger, when the debugger technically already has access to the dict through `thread.DebugFrame(i).Callable().(*starlark.Function).module.predeclared`.

In terms of what this `Predeclared` method should return, I considered returning `fn.module.predeclared`, but that dictionary is the actual dictionary used for execution, and modification would be ill-advised. Since StringDict has no shallow-freeze mechanism, it's probably better for `Predeclared` to return a clone of `fn.module.predeclared`.

Related to #304.

timothyg-stripe:
This PR adds a Predeclared method to Function, similar to the existing Globals method.

I considered returning just fn.module.predeclared, but that dictionary is the actual dictionary used for execution, and any inadvertent modification of that dict by the caller of Predeclared would be ill-advised. Since StringDict has no shallow-freeze mechanism, it's probably better for Predeclared to return a clone.

Also, edit the doc comment for Globals to remove some confusing "unfrozen" wording: a StringDict itself can never be frozen anyway.

Fixes #594

### DIFF
--- a/starlark/eval_test.go
+++ b/starlark/eval_test.go
@@ -828,6 +828,35 @@ def somefunc():
 	}
 }
 
+func TestFunctionGlobals(t *testing.T) {
+	predeclared := starlark.StringDict{
+		"predecl": starlark.String("foo"),
+	}
+	globals, _ := starlark.ExecFile(&starlark.Thread{}, "func.star", `
+def somefunc():
+	return predecl
+`, predeclared)
+
+	fn := globals["somefunc"].(*starlark.Function)
+	fnGlobals := fn.Globals()
+	if !reflect.DeepEqual(fnGlobals, globals) {
+		t.Errorf("somefunc.Globals(): got %q, want %q", fnGlobals, globals)
+	}
+	fnGlobals["shouldnt_appear"] = starlark.String("something_new")
+	if globals.Has("shouldnt_appear") {
+		t.Errorf("somefunc.Globals(): Globals() reflected onto ExecFile's returned globals")
+	}
+
+	fnPredeclared := fn.Predeclared()
+	if !reflect.DeepEqual(fnPredeclared, predeclared) {
+		t.Errorf("somefunc.Predeclared(): got %q, want %q", fnPredeclared, predeclared)
+	}
+	fnPredeclared["shouldnt_appear"] = starlark.String("something_new")
+	if predeclared.Has("shouldnt_appear") {
+		t.Errorf("somefunc.Predeclared(): Predeclared() reflected onto original predeclared")
+	}
+}
+
 func TestFrameLocals(t *testing.T) {
 	// trace prints a nice stack trace including argument
 	// values of calls to Starlark functions.
--- a/starlark/value.go
+++ b/starlark/value.go
@@ -710,7 +710,7 @@ type module struct {
 	constants   []Value
 }
 
-// makeGlobalDict returns a new, unfrozen StringDict containing all global
+// makeGlobalDict returns a new StringDict containing all global
 // variables so far defined in the module.
 func (m *module) makeGlobalDict() StringDict {
 	r := make(StringDict, len(m.program.Globals))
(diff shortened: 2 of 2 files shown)


starlark/value.go
```diff
@@ -730,10 +730,20 @@ func (fn *Function) String() string        { return toString(fn) }
...
 func (fn *Function) Truth() Bool           { return true }
 
-// Globals returns a new, unfrozen StringDict containing all global
+// Globals returns a new StringDict containing all global
 // variables so far defined in the function's module.
 func (fn *Function) Globals() StringDict { return fn.module.makeGlobalDict() }
 
+// Predeclared returns a new StringDict containing the
+// predeclared names in the function's module.
+func (fn *Function) Predeclared() StringDict {
```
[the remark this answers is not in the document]
timothyg-stripe: Done, PTAL.

starlark/value.go
```diff
@@ -696,43 +696,61 @@ func (*stringCodepointsIterator) Done() {}
...
 
-// A module is the dynamic counterpart to a Program.
-// All functions in the same program share a module.
-type module struct {
-	program     *compile.Program
+// A Module represents an evaluated Starlark module.
+// It is the dynamic counterpart to a [Program].
+// All functions in the same program share a Module.
+type Module struct {
+	program     *Program
```
timothyg-stripe, line 708: This type change make it so that `prog.InitModule().Program() == prog`.

starlark/value.go
```diff
@@ -696,43 +696,61 @@ func (*stringCodepointsIterator) Done() {}
...
 func (fn *Function) Truth() Bool           { return true }
+func (fn *Function) Module() *Module       { return fn.module }
 
-// Globals returns a new, unfrozen StringDict containing all global
+// Globals returns a new StringDict containing all global
 // variables so far defined in the function's module.
-func (fn *Function) Globals() StringDict { return fn.module.makeGlobalDict() }
+//
+// fn.Globals() is equivalent to fn.Module().Globals().
+// We maintain this function for compatibility.
```
adonovan, line 752: You can delete this line.

starlark/value.go
```diff
@@ -696,43 +696,61 @@ func (*stringCodepointsIterator) Done() {}
...
+	r := make(StringDict, len(m.program.compiled.Globals))
+	for i, id := range m.program.compiled.Globals {
 		if v := m.globals[i]; v != nil {
 			r[id.Name] = v
 		}
 	}
 	return r
 }
 
+// Predeclared returns the [StringDict] containing the
```
adonovan, line 731: How about:

// Predeclared returns the predeclared environment used
// to construct this module.

starlark/value.go
```diff
@@ -696,43 +696,61 @@ func (*stringCodepointsIterator) Done() {}
...
+// All functions in the same program share a Module.
+type Module struct {
+	program     *Program
 	predeclared StringDict
 	globals     []Value
 	constants   []Value
 }
 
-// makeGlobalDict returns a new, unfrozen StringDict containing all global
+// Program returns the static Starlark program corresponding to this module.
```
adonovan, line 714: // Program returns the program from which this module was constructed.

starlark/eval.go
```diff
@@ -435,13 +435,21 @@ func CompiledProgram(in io.Reader) (*Program, error) {
 // executes the toplevel code of the specified program,
 // and returns a new, unfrozen dictionary of the globals.
 func (prog *Program) Init(thread *Thread, predeclared StringDict) (StringDict, error) {
-	toplevel := makeToplevelFunction(prog.compiled, predeclared)
+	module, err := prog.InitModule(thread, predeclared)
```
adonovan, line 438: Let's revert these changes (LL438-452) as there's no need to publish InitModule.

starlark/eval_test.go
```diff
@@ -828,6 +828,61 @@ def somefunc():
 	}
 }
 
+func TestFunctionGlobals(t *testing.T) {
```
adonovan, line 831: TestFunction_Globals (since it's a method)

starlark/eval_test.go
```diff
@@ -828,6 +828,61 @@ def somefunc():
...
+	if !reflect.DeepEqual(fnGlobals, globals) {
+		t.Errorf("somefunc.Globals(): got %q, want %q", fnGlobals, globals)
+	}
+	fnGlobals["shouldnt_appear"] = starlark.String("something_new")
+	if globals.Has("shouldnt_appear") {
+		t.Errorf("somefunc.Globals(): Globals() reflected onto ExecFile's returned globals")
+	}
+}
+
+func TestFunctionModule(t *testing.T) {
```
adonovan, line 851: TestFunction_Module

adonovan (commented):
Test looks good; thanks.

timothyg-stripe:
fixed all the comments

adonovan (approved):
LGTM, thanks again.

PR #597, opened 2025-07-01
Add a MembershipCheckable interface and use that for the IN check.

### LINKED ISSUE #596: `in` operator not implementable without mapping interface
From looking at [the code](https://github.com/google/starlark-go/blob/a079b1f113da51a439427f62f21380bef60e67e4/starlark/eval.go#L1053), it seems that the only way of supporting the `in` operator for a custom type/`Value` implementation is by implementing the `Mapping` interface.

This seems too limiting. `in` has different meanings for, e.g., `list` and `dict`. In the current state, it is impossible to implement a list-like value with extras that properly supports both indexed access and `in`. Since the `Mapping` interface takes precedence over `Indexable` for evaluating subscript expressions, it's not possible to support `Get` without either breaking index access, or having `in` checks incorrectly return `True` for indices rather than elements.

This seems to be easily fixable in a backwards-compatible way by adding a new interface (e.g., `Container`, `Containsable` etc.) that offers a single `Contains(starlark.Value) (bool, error)` method used to implement `in`/`not in` semantics and takes precedence over the `Mapping` interface for this purpose. `*starlark.List` and `*starlark.Set` could implement that, and no changes would be needed for `*starlark.Dict`, since preserving the `Mapping.Get`-semantics fallback is required for backwards-compatibility anyways.

Happy to send out a PR if this makes sense.

misberner:
Implement `IN` check for `Tuple` and `*List` via that, too.

Fixes #596 .

### DIFF
--- a/starlark/eval.go
+++ b/starlark/eval.go
@@ -1032,7 +1032,7 @@ func Binary(op syntax.Token, x, y Value) (Value, error) {
 
 	case syntax.IN:
 		switch y := y.(type) {
-		case Container: // List, Tuple, Set etc.
+		case Container: // List, Tuple, Set, String, Bytes, rangeValue etc.
 			found, err := y.Has(x)
 			return Bool(found), err
 		case Mapping: // e.g. dict
@@ -1040,31 +1040,6 @@ func Binary(op syntax.Token, x, y Value) (Value, error) {
 			// errors (value cycle, type error) from "key not found".
 			_, found, _ := y.Get(x)
 			return Bool(found), nil
-		case String:
-			needle, ok := x.(String)
-			if !ok {
-				return nil, fmt.Errorf("'in <string>' requires string as left operand, not %s", x.Type())
-			}
-			return Bool(strings.Contains(string(y), string(needle))), nil
-		case Bytes:
-			switch needle := x.(type) {
-			case Bytes:
-				return Bool(strings.Contains(string(y), string(needle))), nil
-			case Int:
-				var b byte
-				if err := AsInt(needle, &b); err != nil {
-					return nil, fmt.Errorf("int in bytes: %s", err)
-				}
-				return Bool(strings.IndexByte(string(y), b) >= 0), nil
-			default:
-				return nil, fmt.Errorf("'in bytes' requires bytes or int as left operand, not %s", x.Type())
-			}
-		case rangeValue:
-			i, err := NumberToInt(x)
-			if err != nil {
-				return nil, fmt.Errorf("'in <range>' requires integer as left operand, not %s", x.Type())
-			}
-			return Bool(y.contains(i)), nil
 		}
 
 	case syntax.PIPE:
--- a/starlark/eval.go
+++ b/starlark/eval.go
@@ -1034,10 +1034,7 @@ func Binary(op syntax.Token, x, y Value) (Value, error) {
 		switch y := y.(type) {
 		case Container: // List, Tuple, Set etc.
 			found, err := y.Has(x)
-			if err != nil {
-				return Bool(found), err
-			}
-			return Bool(found), nil
+			return Bool(found), err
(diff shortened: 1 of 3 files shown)


starlark/eval.go
```diff
@@ -1032,32 +1032,17 @@ func Binary(op syntax.Token, x, y Value) (Value, error) {
...
-			for _, elem := range y {
-				if eq, err := Equal(elem, x); err != nil {
-					return nil, err
-				} else if eq {
-					return True, nil
-				}
+		case MembershipCheckable: // List, Tuple, Set etc.
+			found, err := y.Has(x)
+			if err != nil {
+				return False, nil
```
adonovan, line 1038: Return, don't discard, the error.


return Bool(found), err

starlark/value.go
```diff
@@ -234,6 +234,14 @@ type Sliceable interface {
 	Slice(start, end, step int) Value
 }
 
+// A MembershipCheckable is a value that supports membership testing
+// via the Has method, which is used by the 'in'/'not in' operator.
+type MembershipCheckable interface {
```
adonovan, line 239: Let's call this Container.

// A Container is a value that supports the 'in' operator.

Mention it at L28. (Stupid GitHub PR tool won't let me comment on that line.)


Add `var ( _ Container = (*List)(nil); ...)`  for each known implementation (see L255 for example).

starlark/eval.go
malte-prophet: ah, meant to do that. fixed

starlark/eval.go
```diff
@@ -1032,32 +1032,17 @@ func Binary(op syntax.Token, x, y Value) (Value, error) {
...
-		case Tuple:
-			for _, elem := range y {
-				if eq, err := Equal(elem, x); err != nil {
-					return nil, err
-				} else if eq {
-					return True, nil
-				}
+		case Container: // List, Tuple, Set etc.
+			found, err := y.Has(x)
+			if err != nil {
```
adonovan, line 1037: I meant that you could avoid the if and just say

    return Bool(found), err

to address both cases without a branch.

(What you have is also correct, but stylistically I would suggest `return false, err` in the failure case, since the first result is immaterial.)

starlark/value.go
```diff
@@ -26,6 +26,7 @@
 //
 //	Callable        -- value is callable like a function
 //	Comparable      -- value defines its own comparison operations
+//  Container       -- value supports the 'in' operator
```
adonovan: inconsistent indentation

starlark/eval.go
```diff
@@ -1032,32 +1032,17 @@ func Binary(op syntax.Token, x, y Value) (Value, error) {
...
+			return Bool(found), nil
 		case Mapping: // e.g. dict
 			// Ignore error from Get as we cannot distinguish true
 			// errors (value cycle, type error) from "key not found".
 			_, found, _ := y.Get(x)
 			return Bool(found), nil
-		case *Set:
-			ok, err := y.Has(x)
-			return Bool(ok), err
 		case String:
```
adonovan, line 1046: Should string also be a Container? Since the interface is defined as "supports `in`", nothing more, I think so. Similarly, `bytes` and `rangeValue`. This leads to greater consistency between Starlark's `in` operator and Go code calling `x.(Container).Has(y)` directly.

starlark/value.go
malte-prophet: this was a bad commit caused by IDE auto-format, fixed

starlark/eval.go
malte-prophet: got it now. done

starlark/eval.go
malte-prophet: Made that change

adonovan (approved):
Looks good. Many thanks!

PR #601, opened 2025-08-04
Support `while` statements in `syntax.Walk`

misberner:
`WhileStmt` may occur if allowed via the `syntax.FileOptions`. If it's present in an AST, we should assume it is allowed in the respective context and handle it properly when `Walk`ing.


adonovan (approved):
Good catch! Thanks for the fix.

misberner:
thx for the quick review @adonovan ! just fyi, I am not able to merge this PR myself

PR #615, opened 2025-10-23
Implement json.encode_indent

### LINKED ISSUE #614: Please add json.encode_indent
See Bazel: https://bazel.build/rules/lib/core/json#encode_indent

We've received a bug report internally that some tests which use starlark-go to parse .bzl files are failing because starlark-go doesn't know about json.encode_indent.

CC @tjgq and @hvadehra.

tetromino:
See https://bazel.build/rules/lib/core/json#encode_indent

Fixes #614

### DIFF
--- a/lib/json/json.go
+++ b/lib/json/json.go
@@ -253,8 +253,8 @@ func encodeIndent(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tu
 	if err != nil {
 		return nil, err
 	}
-	buf := new(bytes.Buffer)
-	if err := json.Indent(buf, []byte(str.(starlark.String)), prefix, indent); err != nil {
+	var buf bytes.Buffer
+	if err := json.Indent(&buf, []byte(str.(starlark.String)), prefix, indent); err != nil {
 		return nil, fmt.Errorf("%s: %v", b.Name(), err)
 	}
 	return starlark.String(buf.String()), nil


lib/json/json.go
```diff
@@ -233,6 +239,27 @@ func encode(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, k
...
+	); err != nil {
+		return nil, err
+	}
+	// Rely on encode() to parse the positional parameter (since the signature matches); use our b so
+	// error messages are attributed to json.encode_indent.
+	str, err := encode(thread, b, args, nil)
+	if err != nil {
+		return nil, err
+	}
+	buf := new(bytes.Buffer)
```
adonovan: strings.Builder

will save an allocation.

tetromino: Currently starlark-go relies on `json.Indent` for indenting, which writes its output into a `*bytes.Buffer`: https://pkg.go.dev/encoding/json#Indent

Maybe I'm missing something obvious, but I don't see how it can be efficiently made to write into a `strings.Builder`.

> BTW, allocating `buf` via `new(T)` or `var x T` in fact makes no difference

Thank you for correcting my wrong mental model of Go's allocator!

adonovan: Oh, good point, the standard encoding/json package is indeed a constraint. This is fine.

PR #630, opened 2026-02-10
starlark: apply the [modernize] tool.

happygiraffe:
$ modernize -minmax=false -fix ./...

[modernize]: https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/modernize


adonovan:
Thanks, this looks good. Because the go.mod file is currently 4 Go releases behind, the modernizers are relatively restrained in which fixes they suggest, but it'll be easier to review a second PR of `go fix` changes if we take care of these older ones (in particular the numerous `any` fixes first).

adonovan (approved):

PR #634, opened 2026-03-24
Fix a typo in hashString() that caused poor hash distribution.

### LINKED ISSUE #633: `hashString` yields poorly distributed hash values for long strings
The `hashString` function appears to produce a very non-uniformly distributed hash value for longer strings by bitwise ORing together the upper and lower halves of the 64-bit value from `maphash`. Presumably this was intended to be an XOR. As it stands I believe this will result in unnecessary hash collisions.

```go
// hashString computes the hash of s.
func hashString(s string) uint32 {
	if len(s) >= 12 {
		// Call the Go runtime's optimized hash implementation,
		// which uses the AES instructions on amd64 and arm64 machines.
		h := maphash.String(seed, s)
		return uint32(h>>32) | uint32(h)
	}
	return softHashString(s)
}
```

jgennis:
When reducing the 64-bit hash returned by maphash.String() to the needed 32-bit hash, a bitwise OR was used to combine the upper and lower halves of the 64-bit hash. This commit changes it to use a bitwise XOR instead, which should better preserve the uniform distribution of the hash values.

Fixes https://github.com/google/starlark-go/issues/633

### DIFF
--- a/starlark/hashtable.go
+++ b/starlark/hashtable.go
@@ -434,7 +434,7 @@ func hashString(s string) uint32 {
 		// Call the Go runtime's optimized hash implementation,
 		// which uses the AES instructions on amd64 and arm64 machines.
 		h := maphash.String(seed, s)
-		return uint32(h>>32) | uint32(h)
+		return uint32(h>>32) ^ uint32(h)
 	}
 	return softHashString(s)
 }


adonovan (approved):
LGTM. Well spotted, and thanks again.

PR #636, opened 2026-04-05
starlark: add maxAlloc guard to list, sorted, enumerate

YuvalFradkin1:
list(), sorted(), and enumerate() preallocate a backing slice using
Len(iterable) as the capacity with no upper-bound check. When called
with a large range() value (e.g. list(range(4611686018427387904))),
the resulting make([]Value, 0, 2^62) overflows and the Go runtime panics:

  panic: runtime error: makeslice: cap out of range
  go.starlark.net/starlark.list(...)
      starlark/library.go:689

The panic is unrecovered and kills the host process. Any application
evaluating untrusted Starlark is affected.

The *Repeat helpers in eval.go already define and enforce maxAlloc=1<<30
for string/bytes/tuple repetition. This PR applies the same guard to
list(), sorted(), and enumerate().

Fixes: https://issuetracker.google.com/issues/499639516

### DIFF
--- a/starlark/library.go
+++ b/starlark/library.go
@@ -345,6 +345,9 @@ func enumerate(thread *Thread, _ *Builtin, args Tuple, kwargs []Tuple) (Value, e
 	var x Value
 
 	if n := Len(iterable); n >= 0 {
+		if uint(n) >= maxAlloc/2 {
+			return nil, fmt.Errorf("enumerate: argument too large (%d elements)", n)
+		}
 		// common case: known length
 		pairs = make([]Value, 0, n)
 		array := make(Tuple, 2*n) // allocate a single backing array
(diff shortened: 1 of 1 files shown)


adonovan:
Thanks. The code changes here look fine to me as a means to prevent accidents, but I fear this is at best a band-aid covering the most trivial ways in which it is possible to cause the interpreter to allocate too much memory. A creative adversary can easily find more complicated ways to do it, and I think it's unlikely that there will be a systematic way to avoid such problems by handling Go OOM conditions. (Even if all allocations performed by the interpreter used a different allocator, application-defined built-in functions would not.)

That said, the iterable returned by `range` does make it particularly easy to turn a large number into a large allocation. We should probably also address these operations which materialize an iterable sequence: `tuple`, `reversed`, `zip`.

PR #637, opened 2026-05-21
Use more ergonomic unsafe and atomic wrappers

ash2k:


internal/compile/serial.go
```diff
@@ -317,19 +317,15 @@ func (d *decoder) uint64() uint64 {
...
-		// by unsafely aliasing slice.
-		type string struct {
-			data *byte
-			len  int
-		}
-		ptr := (*string)(unsafe.Pointer(&s))
-		ptr.data = &slice[0]
-		ptr.len = len(slice)
+func (d *decoder) string() string {
+	slice := d.bytes()
```
adonovan: There's no need to treat the zero case specially any more:
```go
	// Avoid a memory allocation for each string
	// by unsafely aliasing slice.
	slice := d.bytes()
	return unsafe.String(unsafe.SliceData(slice), len(slice))
```

adonovan (commented):
Thanks!

internal/compile/serial.go
ash2k: Updated the code. Thank you!

adonovan (approved):
Thanks!

PR #639, opened 2026-06-12
perf: add allocation-free fast path to string_startswith/endswith

jbedard:
In the simple case with a single positional argument (`s.startswith(x)`) the call to `UnpackPositionalArgs` forced `x`, `start`, and `end` to escape to the heap costing four allocations per call (three variables plus the varargs slice).

In a profile of a Starlark-heavy application, these allocations in `string_startswith` accounted for 4.7% of all allocated objects across the entire process.

BenchmarkStarlark/bench_startswith (1000 calls per op):
```
┌────────┬─────────┬─────────┬───────────┐
│        │  ns/op  │  B/op   │ allocs/op │
├────────┼─────────┼─────────┼───────────┤
│ before │ 140,949 │ 128,048 │ 6,001     │
├────────┼─────────┼─────────┼───────────┤
│ after  │ 90,988  │ 80,048  │ 3,001     │
└────────┴─────────┴─────────┴───────────┘
```

### DIFF
--- a/lib/json/json.go
+++ b/lib/json/json.go
@@ -90,8 +90,8 @@ var Module = &starlarkstruct.Module{
 }
 
 func encode(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
-	var x starlark.Value
-	if err := starlark.UnpackPositionalArgs(b.Name(), args, kwargs, 1, &x); err != nil {
+	x, err := starlark.UnpackPositional1(b.Name(), args, kwargs)
+	if err != nil {
 		return nil, err
 	}
 
--- a/lib/json/json.go
+++ b/lib/json/json.go
@@ -301,8 +301,8 @@ func indent(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, k
 	); err != nil {
 		return nil, err
 	}
-	var str string // positional-only
-	if err := starlark.UnpackPositionalArgs(b.Name(), args, nil, 1, &str); err != nil {
+	str, err := starlark.UnpackString1(b.Name(), args, nil)
+	if err != nil {
 		return nil, err
 	}
 
--- a/lib/math/math.go
+++ b/lib/math/math.go
@@ -163,9 +163,8 @@ func log(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, kwar
 }
 
 func ceil(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
-	var x starlark.Value
-
-	if err := starlark.UnpackPositionalArgs("ceil", args, kwargs, 1, &x); err != nil {
+	x, err := starlark.UnpackPositional1("ceil", args, kwargs)
+	if err != nil {
 		return nil, err
 	}
 
@@ -180,9 +179,8 @@ func ceil(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwa
 }
 
 func floor(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
-	var x starlark.Value
-
-	if err := starlark.UnpackPositionalArgs("floor", args, kwargs, 1, &x); err != nil {
+	x, err := starlark.UnpackPositional1("floor", args, kwargs)
+	if err != nil {
 		return nil, err
 	}
 
--- a/lib/proto/proto.go
+++ b/lib/proto/proto.go
@@ -173,8 +173,8 @@ var Module = &starlarkstruct.Module{
 // (Technically a pool may also have many FileDescriptors with the same
 // file name, but this can't happen with a single consistent snapshot.)
 func file(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
-	var filename string
-	if err := starlark.UnpackPositionalArgs(fn.Name(), args, kwargs, 1, &filename); err != nil {
+	filename, err := starlark.UnpackString1(fn.Name(), args, kwargs)
+	if err != nil {
 		return nil, err
 	}
 
--- a/lib/proto/proto.go
+++ b/lib/proto/proto.go
@@ -958,8 +958,8 @@ func (rf *RepeatedField) Attr(name string) (starlark.Value, error) {
 }
 
 func repeatedFieldAppend(_ *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
-	var object starlark.Value
-	if err := starlark.UnpackPositionalArgs(b.Name(), args, kwargs, 1, &object); err != nil {
+	object, err := starlark.UnpackPositional1(b.Name(), args, kwargs)
+	if err != nil {
 		return nil, err
 	}
 	rf := b.Receiver().(*RepeatedField)
@@ -1432,8 +1432,8 @@ func (e EnumDescriptor) Name() string { return string(e.Desc.Name()) } // for Ca
 // The Call method implements the starlark.Callable interface.
 // A call to an enum descriptor converts its argument to a value of that enum type.
 func (e EnumDescriptor) CallInternal(_ *starlark.Thread, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
-	var x starlark.Value
-	if err := starlark.UnpackPositionalArgs(string(e.Desc.Name()), args, kwargs, 1, &x); err != nil {
+	x, err := starlark.UnpackPositional1(string(e.Desc.Name()), args, kwargs)
+	if err != nil {
 		return nil, err
 	}
 	v, err := enumValueOf(e.Desc, x)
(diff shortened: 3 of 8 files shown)


jbedard:
I feel like this may be a common situation and could potentially be expanded beyond just `string_startswith`? But that is the one that was noticeable for me when profiling.

adonovan:
Thanks for investigating. I agree that this is a systematic problem with UnpackArgs. I don't think the solution is to change the starts_with function; I think we need to alter UnpackArgs, or if necessary design an alternative to it that avoids escaping local variables in the common case when the expensive cases (Unpacker and default) are impossible.

jbedard:
I've added a commit switching to an alternative that I think is better, and it can be used across many methods with a minimal diff.

For now I've only adjusted `string_startswith` but this same idea (with a few more util methods) can be used everywhere. WDYT? Shall I do it everywhere?

jbedard:
I added a few more commits removing simple usages of `UnpackPositionalArgs`, but there are plenty more... WDYT of this strategy? Is it worth doing more? Or just start with the first few commits with `UnpackPositional1-3`?

adonovan:
> I added a few more commits removing simple usages of UnpackPositionalArgs, but there are plenty more... WDYT of this strategy? Is it worth doing more? Or just start with the first few commits with UnpackPositional1-3?

My understanding is that the root issue is that two of UnpackArgs cases (Unpacker and default) cause the variable pointed to by 'ptr any' to escape. It's nothing to do with the variadic array in UnpackArgs or UnpackPositionalArgs. So the UnpackString function is an effective workaround, but I expect that the Positional1 etc variants would not address the problem (though I haven't actually checked).

In other words, it's the generality of the operation applied to each argument that is the problem. If you know that you don't need that generality (e.g. your parameters are all string or ints), then you can avoid the escape. So that suggests a parallel set of Unpack*NoEscape functions would work. But that's an ugly and error-prone API, and it still doesn't solve the problem in important cases such as floatOrInt, which needs case Unpacker.

I'm still thinking about this, and I'll discuss with the compiler folks. In the meantime I would accept a narrow fix to just the specific hot function, as in your original CL.

BTW, what's your application?

adonovan:
(See https://github.com/google/starlark-go/issues/641.)

jbedard:
> My understanding is that the root issue is that two of UnpackArgs cases (Unpacker and default) cause the variable pointed to by 'ptr any' to escape

I think the primary performance issue is the use of pointers to local vars requires they are allocated on the heap which is significantly slower (for multiple reasons, including additional GC overhead, maybe copying to/from local stack?). It seems the use of `any` also contributes to this or makes it more likely.

I created `UnpackPositional1`, `UnpackPositional2` etc because using varargs and returning an array I think just leads to same issue. This unfortunately leads to many similar methods, IMO while just having basic unpacking of `starlark.Value` these methods are simple enough it doesn't matter.

The problem is these can only handle generic `Value`, as soon as you want a more specific decoding I think something like your `ArgDecoder` util in #641 is better.

> BTW, what's your application?

Bazel BUILD file generation via starlark "plugins". The [gazelle](https://github.com/bazel-contrib/bazel-gazelle) project allows writing BUILD generators for different languages/ecosystems with a Go API, I have [implemented one of those](https://github.com/aspect-build/aspect-gazelle/tree/main/language/orion) that then delegates to starlark plugins so users don't have to write go. The best case this allows users to just use a prebuilt binary (no compiling go code) and write plugins in starlark to customize BUILD file generation.

It gets interesting when these plugins written in starlark invoke `str.endswith()` 1m times (maybe for every import statement in every .js file in a giant monorepo...), that's when I notice perf concerns within starlark-go...

jbedard:
If we want to start with a minimum just for `string_startswith` then https://github.com/google/starlark-go/pull/639/commits/a1301b566038e1b6cf5cfc81251162e5e7c19a86 (and parent commits) which has only the initial `UnpackPositional3` util method is a good place to start.

IMO the `UnpackPositional{1,2,3}` and a few others are worth it as well but it might depend on your long term plans here which maybe can be discussed in #641 more.

If you have a preference for this initial PR I can update it.

jbedard:
Any preferences here?

adonovan:
> If we want to start with a minimum just for string_startswith then https://github.com/google/starlark-go/commit/a1301b566038e1b6cf5cfc81251162e5e7c19a86 (and parent commits) which has only the initial UnpackPositional3 util method is a good place to start.

The UnpackPositional[123] functions are not the solution: they don't actually unpack any Starlark values into Go variables, they merely pluck elements out of an array. We don't need a helper for that.

If there is a solution here, it is to make UnpackArg et al non-escaping. I have a prototype that does this for the int case (by tweaking AsInt to avoid %T) and the default case (by careful use of non-escaping low-level primitives). But the stubborn case is Unpacker. It is not needed in almost all calls to Unpack from built-ins, which tend to use only well-known types such as bool/string/int/list. However, the mere existence of the Unpacker case in the switch causes UnpackArg and all its callers to escape `*ptr`. So, we would need to define a completely parallel (private) family of Unpack* functions that do not support case Unpacker, and use these functions from within the interpreter. I would not want to expose them more broadly.

The question is: is it worth it? I evaluated this benchmark using a hacked-up non-escaping UnpackArg:
```bash
$ go test -run=none -bench=BenchmarkStarlark/bench_string_startswith ./starlark
```
```python
def bench_string_startswith(b):
    "Benchmark string.startswith"
    for _ in range(b.n):
        "hello".startswith("hell")
```
and observed only a 20% improvement (140ns -> 120ns). Given that this benchmark is almost entirely call overhead, that seems like an upper bound on the possible improvement.

What benchmark are you running? What fraction of time was it spending due to unnecessary escape?

jbedard:
> The UnpackPositional[123] functions are not the solution: they don't actually unpack any Starlark values into Go variables, they merely pluck elements out of an array. We don't need a helper for that.

If we don't need a helper then lets just remove the `UnpackArgs` helper? That `UnpackArgs` helper is causing the fundamental issue (#641). The helpers here are just dumb error checks and defaulting values while plucking out of an array.

> The question is: is it worth it? I evaluated this benchmark using a hacked-up non-escaping UnpackArg:

The overhead is the GC required due to the memory allocation, I don't think simply benchmarking your test will show that? I don't even see this method in my CPU profiling, yet is fairly significant in memory profiling which is what this PR fixes.

adonovan:
> > The UnpackPositional[123] functions are not the solution: they don't actually unpack any Starlark values into Go variables, they merely pluck elements out of an array. We don't need a helper for that.
> 
> If we don't need a helper then lets just remove the `UnpackArgs` helper? That `UnpackArgs` helper is causing the fundamental issue (#641). The helpers here are just dumb error checks and defaulting values while plucking out of an array.

It's not entirely dumb: it checks that the starlark.Value is a String and reports an error if not.

As a performance hack, we could rewrite startswith to do the check directly and avoid UnpackArgs, but why those functions? And why not all others? There's a definite cost to readability. That's why I asked about your benchmark.

> > The question is: is it worth it? I evaluated this benchmark using a hacked-up non-escaping UnpackArg:
> 
> The overhead is the GC required due to the memory allocation, I don't think simply benchmarking your test will show that? I don't even see this method in my CPU profiling, yet is fairly significant in memory profiling which is what this PR fixes.

If you run the benchmark for longer, with a tight memory limit to force garbage collection, and show the allocation stats, you can see that it runs much slower due to GC, but the no-escape version allocates less and runs faster, but still by about the same ratio of about 20%.

```
starlark-go$ GOMEMLIMIT=1MiB go test -run=none -benchtime=10s -benchmem -bench=BenchmarkStarlark/bench_string_startswith ./starlark
goos: darwin
goarch: arm64
pkg: go.starlark.net/starlark
cpu: Apple M1 Pro
Before:
BenchmarkStarlark/bench_string_startswith-8         	11258619	       987.0 ns/op	     128 B/op	       6 allocs/op
After:
BenchmarkStarlark/bench_string_startswith-8         	14362344	       782.2 ns/op	      80 B/op	       3 allocs/op
```

jbedard:
I'm using `startswith` just as an initial example because that's what showed highest in my profiling while also being a simple method. But this seems like a generic issue that could be solved or at least improved in every built function.

The starlark code being run is not written by me but has a lot of things like `p.startswith("src/")` or `p.endswith(".java")` etc. Then invoke a few dozen of those over 1m file paths and the memory allocation of `startswith` stands out more then I would expect. IMO these calls should be essentially free and require zero memory allocation.

Just applying this PR as-is ([f052860](https://github.com/google/starlark-go/pull/639/commits/f0528600fa6be07f1f2261acd3ac49e64fbd9801)) as a patch to my binary caused the `string_startswith` go from 2.5m objects created to no longer showing up in the profile at all. This commit makes the same change to other methods that are also reduced but the `string_startswith` method is still the prime example.

A robot's analysis of before vs after this PR:
```
Allocations attributable to starlark packages:

┌───────────────┬──────────┬──────────┬──────────────────┐
│    Metric     │   Base   │ Patched  │        Δ         │
├───────────────┼──────────┼──────────┼──────────────────┤
│ alloc_space   │ 745.3 MB │ 667.3 MB │ −78 MB (−10.5%)  │
├───────────────┼──────────┼──────────┼──────────────────┤
│ alloc_objects │ 17.60 M  │ 14.65 M  │ −2.94 M (−16.7%) │
└───────────────┴──────────┴──────────┴──────────────────┘

Starlark's share of the whole process dropped from 27.6% → 22.9% of all objects allocated. Total process objects barely moved (63.84 M → 63.98 M), so this is a real reduction, not a reshuffle.

Where it comes from (from the -base diff):

- string_startswith fast path (commit dd571b7): −40.9 MB, −2.56 M objects — the single biggest contributor. The allocation-free path basically eliminates this function's heap traffic.
- (*Function).CallInternal (the UnpackPositional# refactor): −34 MB flat / −93 MB cumulative, −3.1 M cumulative objects — less garbage during argument unpacking.
- Small regression introduced: (*Builtin).BindReceiver now shows +8.4 MB / +0.17 M objects — a new allocation the refactor added. Net is still strongly positive, but worth noting.
```

adonovan:
Thanks for the data. Please take a look at https://github.com/google/starlark-go/pull/642, and perhaps patch it into your project and see what kinds of gains it delivers.

jbedard:
Replaced with https://github.com/google/starlark-go/pull/642

PR #640, opened 2026-06-12
fix: off-by-one in Opcode.String

jbedard:
The highest-numbered opcode (`OpcodeMax`, currently `CALL_VAR_KW`) was reported as "illegal op" by `String()`.

### DIFF
--- a/internal/compile/opcode_test.go
+++ b/internal/compile/opcode_test.go
@@ -1,0 +1,23 @@
+package compile_test
+
+import (
+	"strings"
+	"testing"
+
+	"go.starlark.net/internal/compile"
+)
+
+// TestOpcodeNames checks that every opcode has a name, including the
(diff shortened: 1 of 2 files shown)


internal/compile/opcode_test.go
```diff
@@ -0,0 +1,21 @@
+package compile
```
adonovan: Since this tests only public API, let's put it in package compile_test and import compile.

adonovan (approved):
Thanks for this fix!

internal/compile/opcode_test.go
jbedard: Updated 👍

adonovan (approved):
Thanks!

internal/compile/opcode_test.go
js9011678-coder: Push/

Issue #641, opened 2026-06-15
UnpackArgs(&x) causes local variable x to escape


adonovan:
See discussion on https://github.com/google/starlark-go/pull/639: the UnpackArgs function causes all parameter local variables to escape. This is due to two of the cases in the big type switch: the Unpacker interface, and the default case (which uses reflection).

```go
func myBuiltin(thread *Thread, fn *Builtin, args Tuple, kwargs []Tuple) (Value, error) {      
        var ( x string; y int; z custom )
        err := UnpackArgs(fn.Name(), args, kwargs, "x", &x, "y", &y, "z", &z)
        ...
}
```
```go
func UnpackArgs(fnname string, args Tuple, kwargs []Tuple, pairs ...any) error {
	...
	for i, arg := range args {
		err := UnpackArg(arg, pairs[2*i+1])
		...
	}
	...
}

func UnpackArg(v Value, ptr any) error {
	switch ptr := ptr.(type) {
	case Unpacker:
		return ptr.Unpack(v) // ptr escapes!
	case *Value:
		*ptr = v
	case *string:
		s, ok := AsString(v)
		...
		*ptr = s

	... lots more simple cases ...
	
	default:
		ptrv := reflect.ValueOf(ptr) // ptr escapes!
		... complex reflection ...
	}
	...
}
```

Ideally UnpackArgs (or a new API like it) should not cause local vars to escape. In the PR author's application, string.starts_with was found to spend 5% of CPU allocating local string variables (just the headers). An immediate narrow fix for that hot spot (and others like it) seems warranted. Some measurements would be good to justify attacking the  larger problem with new APIs.

One approach is to return values out of the decoder instead of passing variables into the decoder, something like this:
```go
    func myBuiltin(thread *Thread, fn *Builtin, args Tuple, kwargs []Tuple) (Value, error) {      
        d := ArgDecoder(fn, args, kwargs)
        x := d.String(0, "x")
        y := d.Int(1, "y")
        z := d.Unpack(2, "z", MyCustomUnpacker{})
        if err := d.Err(); err != nil {
            return nil, err
        }
        ... 
    }                                                                                             
```
The errors would be statefully accumulated and the user must not forget to check the error at the end (a footgun similar to bufio.Scanner.Err). 

The decoder would not get to see all the names at once, which either forecloses certain features (e.g. spellcheck) or makes them more complicated to implement.

[...]

PR #642, opened 2026-06-26
starlark: reduce escaping in argument unpacking

adonovan:
The Unpack functions, which receive a ptr any value containing a *T pointer, all escape the T variable to which they point. Fundamentally this is because the Unpacker case makes a dynamic method call on the *T value, so the compiler cannot prove that the callee doesn't let the variable escape.

This CL creates a parallel set of internal functions,
 unpackArgsNoEscape
 unpackPositionalArgsNoEscape
that do not support the Unpacker case and take great care not to let *ptr escape in all the other cases.
For example, Sprintf("%T", ptr) entails reflect.ValueOf(ptr), which defeats escape analysis, so we must avoid that. TestUnpackArgNoEscape asserts the invariant.

Nearly all unpack operations from the interpreter's built-in functions now use these no-escape variant functions. Unfortunately these functions are not ideal for broader exposure.

The added benchmark, which just calls string.startswith in a loop, runs 20% faster due to reduced allocation.

For #641

### DIFF
--- a/starlark/eval_test.go
+++ b/starlark/eval_test.go
@@ -946,17 +946,29 @@ func (b *badType) Freeze()               {}
 
 var _ starlark.Value = new(badType)
 
+type badType2 string
+
+func (b badType2) String() string        { return "badType2" }
+func (b badType2) Type() string          { return "badType2" }
+func (b badType2) Truth() starlark.Bool  { return true }
+func (b badType2) Hash() (uint32, error) { return 0, nil }
+func (b badType2) Freeze()               {}
+
+var _ starlark.Value = new(badType)
+
 // TestUnpackErrorBadType verifies that the Unpack functions fail
 // gracefully when a parameter's default value's Type method panics.
 func TestUnpackErrorBadType(t *testing.T) {
+	var v1 *badType
+	var v2 badType2
 	for _, test := range []struct {
-		x    *badType
+		ptr  any
 		want string
 	}{
-		{new(badType), "got NoneType, want badType"},       // Starlark type name
-		{nil, "got NoneType, want *starlark_test.badType"}, // Go type name
+		{&v1, "got NoneType, want *starlark_test.badType"}, // Go type name: (*badType)(nil).Type() panics
+		{&v2, "got NoneType, want badType"},                // Starlark type name: badType2{}.Type() works
 	} {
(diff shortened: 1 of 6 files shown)


jbedard:
This solves the same as #639, with a slightly better result since this has a wider impact.

I was under the impression that we had to drop the pointers in `vars ...any` to solve this, but I guess we just need to prevent certain usages of them which you've done here?

adonovan:
> This solves the same as https://github.com/google/starlark-go/pull/639, with a slightly better result since this has a wider impact.

Great, thanks for checking! Would love to see updated benchmarks for your workload.

> I was under the impression that we had to drop the pointers in vars ...any to solve this, but I guess we just need to prevent certain usages of them which you've done here?

Exactly. Pointers are fine, but it depends what you do with them. Passing them to reflection, or making interface method calls on them, are beyond the escape analysis so it gives up and make conservative judgements. If we can avoid those, we can ensure that the function doesn't force its argument variables to be allocated on the heap.

jbedard:
Here's data similar to [in my pr](https://github.com/google/starlark-go/pull/639#issuecomment-4812426523). This is your PR compared to the same "base" that my PR was compared to. I think this is all via sampling so nothing is perfect but the robot summarizes it well IIUC.

```
┌────────────────────────┬──────────┬──────────┬──────────────────┐
│ Metric (starlark only) │   Base   │ PR #642  │        Δ         │
├────────────────────────┼──────────┼──────────┼──────────────────┤
│ alloc_space            │ 745.3 MB │ 674.8 MB │ −70.6 MB (−9.5%) │
├────────────────────────┼──────────┼──────────┼──────────────────┤
│ alloc_objects          │ 17.60 M  │ 14.09 M  │ −3.50 M (−19.9%) │
└────────────────────────┴──────────┴──────────┴──────────────────┘

Starlark's share of all objects dropped from 27.6% → 22.8%. The whole run actually allocated less in total (objects 63.8 M → 61.8 M, bytes 4995 → 4954 MB), so this is a real reduction, not displacement.

From the -base diff, the win is concentrated exactly where the PR targets it:
- string_startswith: −39.9 MB, −2.56 M objects — the PR's named benchmark. The no-escape unpack path stops startswith's arguments from escaping to the heap.
- (*Function).CallInternal: −27 MB flat / −87 MB cum, −3.78 M cumulative objects — because no-escape unpacking helps every builtin's argument handling, not just one method.
```

jayconrod (approved):

PR #643, opened 2026-07-03
feat: fused CALL_ATTR opcode to avoid BindReceiver alloc

jbedard:
This fulfills a TOOD for creating a fast-path `Opcode` for simple method invocations, especially for builtin methods.

This removes all `(*Builtin).BindReceiver` invocations from my profiling of a large app:

```
┌───────────────────────────────────┬──────────────────┬───────────────────────┐
│ Metric (built-in method binding)  │   Before         │     After             │
├───────────────────────────────────┼──────────────────┼───────────────────────┤
│ (*Builtin).BindReceiver — bytes   │ 51.5 MB          │ 0 (gone from profile) │
├───────────────────────────────────┼──────────────────┼───────────────────────┤
│ (*Builtin).BindReceiver — objects │ 1,125,084 allocs │ 0                     │
└───────────────────────────────────┴──────────────────┴───────────────────────┘
```

### DIFF
--- a/internal/compile/compile.go
+++ b/internal/compile/compile.go
@@ -1660,7 +1660,7 @@ func (fcomp *fcomp) call(call *syntax.CallExpr) {
 	// opt: fuse x.method(pos...) to avoid materializing a closure.
 	if dot, ok := call.Fn.(*syntax.DotExpr); ok {
 		// plain positional args only for CALL_ATTR
-		if npos, ok := plainPositional(call); ok && npos < 0x100 {
+		if npos := len(call.Args); npos < 0x100 && plainPositional(call) {
 			if name := fcomp.pcomp.nameIndex(dot.Name.Name); name < 1<<24 {
 				fcomp.expr(dot.X) // receiver
 				for _, arg := range call.Args {
@@ -1680,18 +1680,18 @@ func (fcomp *fcomp) call(call *syntax.CallExpr) {
 	fcomp.emit1(op, arg)
 }
 
-// plainPositional reports whether all of call's args are plain positional (no name=value, *args, **kwargs), and their count.
-func plainPositional(call *syntax.CallExpr) (int, bool) {
+// plainPositional reports whether all of call's args are plain positional (no name=value, *args, **kwargs).
+func plainPositional(call *syntax.CallExpr) bool {
 	for _, arg := range call.Args {
 		if binary, ok := arg.(*syntax.BinaryExpr); ok && binary.Op == syntax.EQ {
-			return 0, false // named argument
+			return false // named argument
 		}
 		if unary, ok := arg.(*syntax.UnaryExpr); ok &&
 			(unary.Op == syntax.STAR || unary.Op == syntax.STARSTAR) {
-			return 0, false // *args or **kwargs
+			return false // *args or **kwargs
 		}
 	}
-	return len(call.Args), true
+	return true
 }
 
 // args emits code to push a tuple of positional arguments
--- a/internal/compile/compile.go
+++ b/internal/compile/compile.go
@@ -46,7 +46,7 @@ var Disassemble = false
 const debug = false // make code generation verbose, for debugging the compiler
 
 // Increment this to force recompilation of saved bytecode files.
-const Version = 15
+const Version = 16
 
 type Opcode uint8
 
@@ -146,11 +146,14 @@ const (
 	CALL_KW     // fn positional named       **kwargs CALL_KW<n>     result
 	CALL_VAR_KW // fn positional named *args **kwargs CALL_VAR_KW<n> result
 
-	// n>>8 is the method-name index and n&0xff is #positional args.
-	CALL_ATTR // recv positional CALL_ATTR<n> result   (fused x.method(pos...))
+	// ATTR_METHOD+CALL_METHOD implement x.method(pos...) without materializing
+	// a bound-method closure. ATTR_METHOD resolves the method at the receiver
+	// (preserving evaluation order); CALL_METHOD calls it once args are pushed.
+	ATTR_METHOD //           recv ATTR_METHOD<name> method     (method bound to recv)
+	CALL_METHOD // method positional CALL_METHOD<npos> result
 
 	OpcodeArgMin = JMP
-	OpcodeMax    = CALL_ATTR
+	OpcodeMax    = CALL_METHOD
 )
 
 // TODO(adonovan): add dynamic checks for missing opcodes in the tables below.
@@ -159,8 +162,9 @@ var opcodeNames = [...]string{
 	AMP:          "amp",
 	APPEND:       "append",
 	ATTR:         "attr",
+	ATTR_METHOD:  "attr_method",
 	CALL:         "call",
-	CALL_ATTR:    "call_attr",
+	CALL_METHOD:  "call_method",
 	CALL_KW:      "call_kw ",
 	CALL_VAR:     "call_var",
 	CALL_VAR_KW:  "call_var_kw",
@@ -235,8 +239,9 @@ var stackEffect = [...]int8{
 	AMP:          -1,
 	APPEND:       -2,
 	ATTR:         0,
+	ATTR_METHOD:  0,
 	CALL:         variableStackEffect,
-	CALL_ATTR:    variableStackEffect,
+	CALL_METHOD:  variableStackEffect,
 	CALL_KW:      variableStackEffect,
 	CALL_VAR:     variableStackEffect,
 	CALL_VAR_KW:  variableStackEffect,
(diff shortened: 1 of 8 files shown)


internal/compile/compile.go
```diff
@@ -146,8 +146,11 @@ const (
 	CALL_KW     // fn positional named       **kwargs CALL_KW<n>     result
 	CALL_VAR_KW // fn positional named *args **kwargs CALL_VAR_KW<n> result
 
+	// n>>8 is the method-name index and n&0xff is #positional args.
```
adonovan: Would it make more sense to use the lower byte for the number of positional args, like the four preceding opcodes?

internal/compile/compile.go
```diff
@@ -1648,14 +1657,21 @@ func (fcomp *fcomp) binop(pos syntax.Position, op syntax.Token) {
...
+	if dot, ok := call.Fn.(*syntax.DotExpr); ok {
+		// plain positional args only for CALL_ATTR
+		if npos, ok := plainPositional(call); ok && npos < 0x100 {
+			if name := fcomp.pcomp.nameIndex(dot.Name.Name); name < 1<<24 {
+				fcomp.expr(dot.X) // receiver
+				for _, arg := range call.Args {
+					fcomp.expr(arg)
+				}
+				fcomp.setPos(call.Lparen)
+				fcomp.emit1(CALL_ATTR, name<<8|uint32(npos))
```
adonovan: Consider `recv.nonesuch(a, b, c)`: the correct semantics are that recv is evaluated, then `recv.nonesuch`, then the argument expressions a, b, c, then the function is called. Won't this optimization cause the effects of the `recv.nonesuch` ATTR operation to be observed only after the side effects of evaluating a, b, c?

internal/compile/compile.go
```diff
@@ -1664,6 +1680,20 @@ func (fcomp *fcomp) call(call *syntax.CallExpr) {
...
+	for _, arg := range call.Args {
+		if binary, ok := arg.(*syntax.BinaryExpr); ok && binary.Op == syntax.EQ {
+			return 0, false // named argument
+		}
+		if unary, ok := arg.(*syntax.UnaryExpr); ok &&
+			(unary.Op == syntax.STAR || unary.Op == syntax.STARSTAR) {
+			return 0, false // *args or **kwargs
+		}
+	}
+	return len(call.Args), true
```
adonovan: I think this is the caller's concern, and that the function's range should be boolean.

starlark/interp.go
```diff
@@ -384,6 +384,45 @@ loop:
...
+			// Copy args: the callee (a built-in) can't be trusted not to mutate them.
+			var positional Tuple
+			if npos > 0 {
+				positional = slices.Clone(stack[sp-npos : sp])
+			}
+
+			var z Value
+			var err2 error
+			if method := builtinMethod(recv, name); method != nil {
+				// Bind recv in a pooled *Builtin to avoid a BindReceiver alloc.
```
adonovan: This assumes the callee will not hang on to the Builtin after it returns (for example by inserting it into the result). I'm not sure that's safe; we don't make that assumption about the ordinary arguments at L393.

starlark/eval.go
```diff
@@ -67,6 +67,28 @@ type Thread struct {
 
 	// proftime holds the accumulated execution time since the last profile event.
 	proftime time.Duration
+
+	// builtinPool is a LIFO free list of *Builtin reused by CALL_ATTR to bind receivers without allocating.
+	builtinPool []*Builtin
+}
+
+// acquireBoundBuiltin returns a *Builtin (pooled if possible) bound to fn and recv.
+// Pair with releaseBoundBuiltin; the *Builtin must not escape the call (as in CALL_ATTR).
```
adonovan: See my comment below: I don't think we can assume it does not escape.

adonovan:
Thanks, this is an interesting idea. I've added some comments and questions about whether it is a compatible change, both because of reordering of effects and escape of Builtins.

Your benchmark shows that this reduces the allocations at this particular site. What fraction of total allocation (or CPU time, or wall time) does this represent?

jbedard:
Did you forget to press submit on the comments/questions? I see none...

I'll try to benchmark it properly again and come up with real (non robot-generated) benchmarks, but my [favourite pprof analyze](https://pprof.me/) has been failing lately so I'll have to find something new.

IIRC once again in my use case the CPU % is very minimal, but the memory % is high relative to the small amount of wall time which IMO was worth addressing.

adonovan:
> Did you forget to press submit on the comments/questions? I see none...

Yes I did--I (happily) don't spend much time working in GitHub PRs. Done now.

internal/compile/compile.go
jbedard: I was primarily just uncommenting your TODO really. I'll look into this more though...

jbedard:
FWIW the robot claims to recall the original benchmarks (before this PR):
```
- ~1.0% of total alloc bytes (51.5 MB / 4.9 GB)
- ~1.85% of total alloc count (1.13 M / 60.8 M objects), dropping to 0
- negligible CPU/wall (GC-side only)
```

For the negligible amount of CPU/wall-time I think that is a very high allocation count which is what this PR aims to address.

jbedard:
Fixing your concerns (by asking a robot, and me not truly understanding the project) made things the PR a bit more complicated, but the benchmarks still show the improvements.

The golang benchmarks:
```
time and allocations (method-call-dominated)

┌─────────────────────────┬────────┬────────────────────┐
│        Benchmark        │  time  │     allocs/op      │
├─────────────────────────┼────────┼────────────────────┤
│ bench_string_startswith │ −29.5% │ −66.7% (3→1)       │
├─────────────────────────┼────────┼────────────────────┤
│ bench_method_mix        │ −18.6% │ −49.9% (8013→4013) │
├─────────────────────────┼────────┼────────────────────┤
│ bench_builtin_method    │ −18.3% │ −50.0% (2001→1001) │
└─────────────────────────┴────────┴────────────────────┘
```

jbedard:
IMO there should be a better way of doing this that doesn't involve pooling and which would work for more then just builtin methods. I see no reason that `x.y(z)` needs to temporarily create an object binding `y` to `x` before invoking it, builtin or not - but that might be a bigger change? Or maybe I need to restart and do this differently?

adonovan:
> IMO there should be a better way of doing this that doesn't involve pooling and which would work for more then just builtin methods. I see no reason that `x.y(z)` needs to temporarily create an object binding `y` to `x` before invoking it, builtin or not - but that might be a bigger change? Or maybe I need to restart and do this differently?

That's funny, I started writing a long  comment to this effect on your PR, then went to measure something, then some hours flew by, and now I have a sketch of the alternative approach I had in mind. Take a look at https://github.com/google/starlark-go/pull/645.

I'm still not sure I like it: it may hurt the performance in the "legacy" case and it adds yet more complexity to the API surrounding built-in calls, all for a small amount of performance gain. And it causes the frame.callable to lose recv information, which hurts the debug API. But at least the mechanism is not private to the interpreter, so application-defined builtins  can benefit too.

jbedard:
#645 definitely looks better then what this PR has turned into, however it might not be covering as many cases as this PR (possibly because this PR is buggy and covers too much).

Here's an interesting comparison, note "main" doesn't have our previous change so it still has all the allocations for `string_startswith`:

```
┌─────────────────────┬──────────┬──────────┬──────────┬─────────┐
│  Metric (starlark)  │   main   │ PR #645  │ PR #643  │  Best   │
├─────────────────────┼──────────┼──────────┼──────────┼─────────┤
│ alloc_space         │ 745.6 MB │ 684.6 MB │ 648.9 MB │ PR #643 │
├─────────────────────┼──────────┼──────────┼──────────┼─────────┤
│ alloc_objects       │ 17.35 M  │ 13.37 M  │ 13.96 M  │ PR #645 │
├─────────────────────┼──────────┼──────────┼──────────┼─────────┤
│ BindReceiver (objs) │ 1.34 M   │ 1.16 M   │ 0.16 M   │ PR #643 │
├─────────────────────┼──────────┼──────────┼──────────┼─────────┤
│ String.Attr (cum)   │ 1.90 M   │ 0        │ ~0.10 M  │ PR #645 │
├─────────────────────┼──────────┼──────────┼──────────┼─────────┤
│ string_startswith   │ 2.65 M   │ 0        │ 0        │ tie     │
├─────────────────────┼──────────┼──────────┼──────────┼─────────┤
│ CPU duration        │ 17.49 s  │ 17.40 s  │ 17.11 s  │ ~noise  │
├─────────────────────┼──────────┼──────────┼──────────┼─────────┤
│ starlark CPU        │ 0.18 s   │ 0.21 s   │ 0.15 s   │ ~noise  │
└─────────────────────┴──────────┴──────────┴──────────┴─────────┘
```

Note how this has essentially zero impact on CPU usage or timing for this app, but I still think the memory impact (space and object count) is significant enough to justify some time/thought put into it.

jbedard:
Note that #645 doesn't (significantly) reduce the amount of `BindReceiver` objs which IMO implies it's not the right solution either.

Like I previously said, I see no reason that `x.y(z)` needs to temporarily create an object binding `y` to `x` before invoking it, builtin or not. But I'm looking at this from the outside and might be missing something crucial when making that statement 😅

PR #644, opened 2026-07-07
syntax: reject excessively nested bracketed expressions

adonovan:
Both the scanner and the parser contain recursions driven by user input, so an excessively nested input string can cause them to exceed the stack and crash the program.

This change rejects such inputs in both the scanner and parser, using the existing depth counter in the scanner and a new counter in the parser.

The interpreter already imposes a limit.

The tests include a static check that the call graph within the parser has no cycles except those including at least one use of p.enter/p.leave.

This commit addresses GHSA-wcqm-92f8-mh2v, which concerns a denial-of-service vulnerability since attacker-controlled inputss may cause a Starlark-based application to crash.

### DIFF
--- a/syntax/parse.go
+++ b/syntax/parse.go
@@ -143,6 +143,24 @@ type parser struct {
 	in      *scanner
 	tok     Token
 	tokval  tokenValue
+	depth   int
+}
+
+// enter increments the depth and checks the recursion limit.
+// It should be paired with a deferred call to [parser.leave].
+//
+// Every cycle in the parser's static call graph must include at
+// least one function that uses enter+leave to break cycles.
+// TestParserCallGraphCycles statically verifies this property.
+func (p *parser) enter() {
+	p.depth++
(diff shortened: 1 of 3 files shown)


tetromino (approved):

adonovan:
Reviewers: hold off reviewing till I fix the failing test. Sorry...

adonovan:
> Reviewers: hold off reviewing till I fix the failing test. Sorry...

Ok, fixed. The problem was that we need to check whether each cycle contains at least one cycle-breaking node; it is not sufficient to check that all cyclic paths include a cycle-breaking node as it may be on the acyclic initial part of the path. This turned up a couple of other functions that need cycle breaking: parseLambda and parseTestPrec.

jayconrod (approved):
When reading the first half of this change, I was thinking, "It's going to be hard to catch all cases of recursion in the parser". But `TestParserCallGraphCycles` is a neat solution. Very cool.

Issue #304, opened 2020-09-26
Possible to create a simple interactive debugger?


maxmcd:
I created this simple builtin:

```go
debugger := starlark.NewBuiltin("debugger", func(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
	repl.REPL(thread, b.predeclared)
	return nil, nil
})
```

But, unsurprisingly, when it is run, the various local variables or declared global variables are not available in the interactive session.

Is implementing this kind of interactive debugger feasible with the current lib implementation? 

Thanks!

alandonovan:
Not yet. At one point I started writing an asynchronous concurrent debugger with breakpoints, but got distracted. Let me see if I can find it, then perhaps we can define a reasonable API for accessing the environment to support your use case, even if we don't add complete support for breakpoints etc.

maxmcd:
Wonderful, thank you

maxmcd:
Hey @alandonovan any update on this?

adonovan:
Sorry, no progress.

This feature requires that the bytecode interpreter has a single-byte BREAKPOINT instruction, plus new public APIs to
- overwrite existing instructions with BREAKPOINT, saving the overwritten byte so it can be restored later;
- map program counter offsets to file/line/column and back again;
- read and update global and local variables;
- compile and evaluate expressions provided by the user within the environment that exists at a given breakpoint;
- notify the debugger of thread events such as thread creation, function entry/exit, and so on;
And it requires a great deal of care with respect to concurrency.

I think this is probably several weeks' work for someone who understands the implementation well.

AlmogBaku:
In #408 I suggested a somehow naive implementation for basic breakpoints:
I suggest adding a callback in [this](https://github.com/google/starlark-go/blob/master/starlark/interp.go#L66) [few](https://github.com/google/starlark-go/blob/master/starlark/interp.go#L116) [lines](https://github.com/google/starlark-go/blob/master/starlark/interp.go#L319) where the `vmdebug` is placed.

This way we can have a programmatically way for users to implement their own breakpoint mechanism naively.

WDYT?

adonovan:
A debugger is much more than a call to a user-defined function before and after every function call; it requires control over execution at a statement or instruction level, a means of inspecting and perhaps altering program state, and helpers for mapping between source and executable representations of the program.

How would the proposed hook allow someone to step through the iterations of a loop looking at the local variables?

[...]

PR #647, opened 2026-08-09
add Thread.CallDecorator hook for calls

wistful:
Related to #304.

A new optional field on `Thread`:

```go
CallDecorator func(thread *Thread, fn Callable, args Tuple, kwargs []Tuple) (Value, error)
```

If it is `nil`, nothing changes. If set, `starlark.Call` invokes it instead of `fn.CallInternal`. The decorator runs its own logic and calls `fn.CallInternal` itself to run the real function.

Implemented as @adonovan suggested in [#304 (comment)](https://github.com/google/starlark-go/issues/304#issuecomment-1113483537).

### Notes

- The frame of `fn` is already pushed, so `CallFrame(0)` describes `fn`; stack traces and error positions are unchanged.
- Same contract as `CallInternal`: do not keep `args`/`kwargs` after returning, they may alias the caller's operand stack.
- Delegate via `fn.CallInternal`, not `starlark.Call`, because the latter re-enters the decorator.
- Doc comment on `Callable` updated: a `CallDecorator` is the one exception to "never call `CallInternal`".

Tests in `starlark/eval_test.go` cover functions/built-ins/methods, the `nil` case, stack and error positions, replacing the result or error, argument pass-through, and panics.

### Cost

With `CallDecorator == nil` the extra work is one nil check per call. `BenchmarkCallDecorator` compares `nil` against a pass-through decorator, for a built-in and for a Starlark function (`-benchtime=2s -count=10`, ns/op):

```
builtin/nil        10.05 ± 0.08     0 allocs/op
builtin/delegate   11.07 ± 0.14     0 allocs/op
function/nil       44.55 ± 0.71     1 allocs/op
function/delegate  45.46 ± 0.63     1 allocs/op
```

Installing a decorator costs about 1 ns per call (indirect call plus one more stack frame), and no extra allocation. For a built-in that is ~10%, because a built-in call is itself very cheap; for a Starlark function it is ~2%.

### DIFF
--- a/starlark/bench_test.go
+++ b/starlark/bench_test.go
@@ -168,3 +168,50 @@ func BenchmarkProgram(b *testing.B) {
 		}
 	})
 }
+
+// BenchmarkCallDecorator measures the cost of the CallDecorator hook.
+// The "delegate" sub-benchmarks set a decorator that just forwards to fn.CallInternal;
+// comparing "delegate" to "nil" shows the extra cost per call of having a decorator installed.
+// The cost paid by callers that never set CallDecorator is not measured here.
+func BenchmarkCallDecorator(b *testing.B) {
+	builtin := starlark.NewBuiltin("nop", func(thread *starlark.Thread, fn *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+		return starlark.None, nil
+	})
+
+	globals, err := starlark.ExecFile(new(starlark.Thread), "calldecorator.star", "def nop(): pass", nil)
+	if err != nil {
+		b.Fatal(err)
+	}
+	function := globals["nop"].(*starlark.Function)
+
+	b.Run("builtin", func(b *testing.B) {
+		b.Run("nil", func(b *testing.B) {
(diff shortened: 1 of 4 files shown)


adonovan:
Thanks for getting this ball moving. Sorry for the delay--I'm still on a (long) vacation.

Overall the implementation looks straightforward, and the tests are good. The main question I want to think about is the high-level usage of the debug API: if various applications start using CallDecorator for their own special needs, how will a general Starlark debugger that uses CallDecorate (among other APIs) work? Would it disrupt those special features, or cleanly compose with them?

wistful:
> Thanks for getting this ball moving. Sorry for the delay--I'm still on a (long) vacation.
> 
> Overall the implementation looks straightforward, and the tests are good. The main question I want to think about is the high-level usage of the debug API: if various applications start using CallDecorator for their own special needs, how will a general Starlark debugger that uses CallDecorate (among other APIs) work? Would it disrupt those special features, or cleanly compose with them?

[Answered](https://github.com/google/starlark-go/issues/304#issuecomment-5385252241) over in #304, since this now spans #647 and #651, and both hook designs started in that thread.

PR #650, opened 2026-08-23
add exact per-thread call profiler

wistful:
## Summary
This adds opt-in execution profiling through `Thread.Profile`.
It records exact call times and returns the results directly to Go code.
It complements the existing `StartProfile` sampling profiler.

```go
var p starlark.ExecProfile
thread.Profile = &p
_, err := starlark.ExecFile(thread, filename, src, nil)
p.WriteText(os.Stderr)
```

The new profiler:

- records short runs that may not produce a `StartProfile` sample;
- reports calls, self time, cumulative time, and maximum call time;
- makes results available as structured records or a text table;
- can merge results from separate threads; and
- avoids counting time twice during Go-to-Starlark re-entry.

Profiling is off by default. It starts when an `ExecProfile` is assigned to Thread.Profile`. No other setup is needed.

## Timing model

Each row groups calls to one function. 
`Self` is time spent in the function itself. It excludes called functions and time waiting for `Load`. 
`Cum` is the full call time, including both. For recursion, `Cum` counts only the outermost call so the same time is not counted more than once.

Only one function adds to `Self` at a time. The profiler pauses the caller while another function or `Load` runs. This prevents double counting, including when a Go callable calls back into Starlark.

## Performance

When profiling was off, the combined benchmark result was 0.26% slower. 
Most benchmarks did not show a clear change. A few call-heavy results were 1.5% to 2.6% slower, but only `bench_bigint` repeated that result in a second run. The likely cause is that each interpreter frame grew from 56 to 72 bytes to hold profile state.

With profiling enabled, an empty call was about 2.4 to 2.6 times slower. 

## Tests

Tests cover timing and call counts, recursion, Go-to-Starlark re-entry, errors, panics, `Load`, profile changes during a run, merging, parallel threads, and use with `StartProfile`. They also check that total self time does not exceed wall time.

### DIFF
--- a/starlark/eval.go
+++ b/starlark/eval.go
@@ -50,6 +50,14 @@ type Thread struct {
 	// The default behavior is to call thread.Cancel("too many steps").
 	OnMaxSteps func(thread *Thread)
 
+	// Profile records the exact wall time of Starlark calls on this thread.
+	// A nil Profile disables this profiling. See [ExecProfile].
+	//
+	// Like Print and Load, set or clear Profile only from the goroutine that
+	// runs this thread. Read it only after execution returns to the calling
+	// Go code.
+	Profile *ExecProfile
+
 	// Steps a count of abstract computation steps executed
 	// by this thread. It is incremented by the interpreter. It may be used
 	// as a measure of the approximate cost of Starlark execution, by
@@ -67,6 +75,8 @@ type Thread struct {
 
 	// proftime holds the accumulated execution time since the last profile event.
 	proftime time.Duration
+
+	execProfileSpan profileSpan
 }
 
 // ExecutionSteps returns the current value of Steps.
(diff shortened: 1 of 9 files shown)


adonovan:
Thanks, this looks interesting. I have a pile of Starlark PRs that need reviewing and hope to set aside a day to look at them together, but I am currently overwhelmed with Go-related reviews, so this will likely not happen very soon. Thanks for your patience.

adonovan:
Also: this PR adds a wall-time profiler. Such profiles may be useful for finding where time is spent, but often the answer is "sleeping" or "waiting for I/O", and what you really want to know is why are my CPUs busy.

I've been mulling various ideas for integrating into the Go runtime's CPU profiler, so that each Go frame in starkark.Function.CallInternal is somehow replaced by a unique dummy frame corresponding to the specific Starlark function, with correct name/file/line. This may need runtime hacks to achieve (to mutate the PC->virtual frame mapping) but I'd like to experiment with it before we commit to wall-time profiling. Perhaps it's something you'd be interested in trying.

One snag is that, if you're working on macOS, its kernel's setitimer-based CPU profiler is [broken](https://github.com/golang/go/issues/57722), so you need to use Linux or somesuch to get reliable data.

adonovan:
I got totally nerdswiped and sketched the Go runtime integration I had in mind. See #659.

PR #652, opened 2026-08-25
Embed the parse.go source for TestParserCallGraphCycles.

happygiraffe:
This upstreams a patch internal to Google.

### DIFF
--- a/syntax/parse_test.go
+++ b/syntax/parse_test.go
@@ -7,6 +7,7 @@ package syntax_test
 import (
 	"bufio"
 	"bytes"
+	_ "embed"
 	"fmt"
 	"go/ast"
 	"go/build"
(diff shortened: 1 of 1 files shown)


happygiraffe:
Rationale: in the internal Google build system (~Bazel), the source is not always available when the test executes. Embedding ensures it's available for the test.

adonovan:
I think the patch belongs in the Google code, not here. Go tests may (and do) assume the invariant that the test process's working directory is that of the package under test, so this code is valid as is.

Blaze and Bazel go_test rules instead set the working directory to the root of the tree, which seems like a recipe for problems of this kind, but the solution (it seems to me) is in Blaze and Bazel, not in every Go test. Of course the bug may be too hard to fix at this point.

happygiraffe:
Understood. I'll revert this change.

Unfortunately I too think the bug is likely too hard to fix now.

PR #653, opened 2026-09-03
Inline iteration methods

ash2k:
This project requires Go 1.25, so build tags are no longer necessary.


starlark/value.go
```diff
@@ -291,6 +292,33 @@ type Iterator interface {
 	Done()
 }
 
+// Elements returns an iterator for the elements of the iterable value.
```
adonovan: Could you move these functions further down? The file up to this point consists only of type declarations.

starlark/value.go
```diff
@@ -291,6 +292,33 @@ type Iterator interface {
...
+//
+// Example of go1.23 iteration:
```
adonovan: Let's just say "Example:" here.

starlark/value.go
```diff
@@ -316,6 +344,43 @@ type IterableMapping interface {
 
 var _ IterableMapping = (*Dict)(nil)
 
+// Entries returns an iterator over the entries (key/value pairs) of
```
adonovan: ditto here

starlark/value.go
```diff
@@ -316,6 +344,43 @@ type IterableMapping interface {
...
+	// If available (e.g. *Dict), use specialized push iterator,
+	// as it gets k and v in one shot.
+	type hasEntries interface {
+		Entries() iter.Seq2[Value, Value]
+	}
+	if mapping, ok := mapping.(hasEntries); ok {
+		return mapping.Entries()
+	}
+
+	iter := mapping.Iterate()
```
adonovan: This looks like a latent bug: the Iterate call should occur inside the function below, so that (a) asking for a Seq2 doesn't prematurely increase the mapping's active iterator count and (b) iterating over the same Seq2 multiple times doesn't call Done more times than Iterate and underflow the mapping's active iterator count.

The Elements function has the same bug.

Could you make a separate commit to fix it, and test it? Many thanks.

starlark/value.go
```diff
@@ -291,6 +292,33 @@ type Iterator interface {
...
+func Elements(iterable Iterable) iter.Seq[Value] {
+	// Use specialized push iterator if available (*List, Tuple, *Set).
+	type hasElements interface {
+		Elements() iter.Seq[Value]
+	}
+	if iterable, ok := iterable.(hasElements); ok {
+		return iterable.Elements()
+	}
+
+	iter := iterable.Iterate()
```
adonovan: See comment below -- same bug here.

adonovan (commented):
This looks reasonable, and in skimming this code again I noticed a preexisting bug.

starlark/value.go
```diff
@@ -1011,6 +1013,25 @@ func (l *List) Iterate() Iterator {
 	return &listIterator{l: l}
 }
 
+// Elements returns a go1.23 iterator over the elements of the list.
```
adonovan: ...the sequence of elements...

starlark/value.go
```diff
@@ -1122,6 +1143,21 @@ func (t Tuple) Slice(start, end, step int) Value {
 
 func (t Tuple) Iterate() Iterator { return &tupleIterator{elems: t} }
 
+// Elements returns a go1.23 iterator over the elements of the tuple.
```
adonovan: ...the sequence of elements...

starlark/value.go
```diff
@@ -1634,6 +1676,70 @@ func Iterate(x Value) Iterator {
 	return nil
 }
 
+// Elements returns an iterator for the elements of the iterable value.
```
adonovan: ...the sequence of elements...

starlark/value.go
```diff
@@ -1634,6 +1676,70 @@ func Iterate(x Value) Iterator {
...
+	return func(yield func(Value) bool) {
+		iter := iterable.Iterate()
+		defer iter.Done()
+		var x Value
+		for iter.Next(&x) && yield(x) {
+		}
+	}
+}
+
+// Entries returns an iterator over the entries (key/value pairs) of
```
adonovan: the sequence of entries

starlark/value_test.go
```diff
@@ -155,3 +156,132 @@ func TestParamDefault(t *testing.T) {
 		})
 	}
 }
+
```
adonovan, line 159: Can you put all these supporting definitions beneath the tests that use them?

starlark/value_test.go
```diff
@@ -155,3 +156,132 @@ func TestParamDefault(t *testing.T) {
...
+// A fakeMapping is a starlark.IterableMapping that deliberately does not
+// implement the Elements or Entries fast paths, so that the standalone
+// starlark.Elements and starlark.Entries functions must use the generic
+// Iterate/Done code path.
+//
+// Like *List and *Dict it counts its active iterators, but as a signed
+// integer so that the tests can observe underflow.
+type fakeMapping struct {
+	items     []starlark.Tuple
```
adonovan: If you simply embed an existing value (e.g. a Set) here and override the Elements method with the wrong signature, effectively removing it, then Set will do most of the work and you won't need to declare all the methods below, including the existing iterator. You can test that the itercount is zero by attempting to mutate the Set.

adonovan (commented):
I added a few minor suggestions for comment improvements since go1.23 is far in the rear view mirror now. Also I think the test (thanks for that) can be simplified.

starlark/value_test.go
```diff
@@ -155,3 +156,61 @@ func TestParamDefault(t *testing.T) {
...
+
+	// Asking for the sequence must not start an iteration.
+	_ = starlark.Entries(m)
+	err := d.SetKey(starlark.String("one"), starlark.MakeInt(1))
+	if err != nil {
+		t.Error("Entries(m) started an iteration")
+	}
+}
+
+// A fakeMapping is a starlark.IterableMapping that deliberately does not
```
adonovan, line 186: Much better. Thanks!

adonovan (approved):

PR #655, opened 2026-09-04
Struct: add `Len()` and `AttrAt()`

### LINKED ISSUE #654: Interface for objects with `string` -> `Value` mapping
`StringDict`, `Module`, and `Struct` all have `string` keys/attributes and `Value` values. I'd like to add a common method to these types that returns a `iter.Seq2[string, Value]`. Perhaps it can be called `Entries`. Would that be accepted? I can also add an interface for this, but not sure how to call it and if its necessary.

Separately, I'd like to also add/expose the `Len() int` method on these types. `Struct` has a private one, `Module` and `StringDict` don't have it (want to add it for consistency). Would that be accepted?

ash2k:
Closes #654.

### DIFF
--- a/starlarkstruct/struct.go
+++ b/starlarkstruct/struct.go
@@ -246,6 +246,19 @@ func (s *Struct) AttrNames() []string {
 	return names
 }
 
+// AttrAt returns the value of the field at the specified index.
+func (s *Struct) AttrAt(i int) (starlark.Value, error) {
+	if !(0 <= i && i < len(s.entries)) {
+		var ctor string
+		if s.constructor != Default {
+			ctor = s.constructor.String() + " "
+		}
+		return nil, starlark.NoSuchAttrError(
+			fmt.Sprintf("%sstruct has no attribute at index %d", ctor, i))
+	}
+	return s.entries[i].value, nil
+}
+
 func (x *Struct) CompareSameType(op syntax.Token, y_ starlark.Value, depth int) (bool, error) {
 	y := y_.(*Struct)
 	switch op {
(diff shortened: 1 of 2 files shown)


starlarkstruct/struct.go
```diff
@@ -246,6 +246,19 @@ func (s *Struct) AttrNames() []string {
 	return names
 }
 
+// AttrAt returns the value of the field at the specified position.
+func (s *Struct) AttrAt(pos int) (starlark.Value, error) {
```
adonovan: i int

(pos sounds like a source position)

starlarkstruct/struct.go
```diff
@@ -246,6 +246,19 @@ func (s *Struct) AttrNames() []string {
...
+	if pos < 0 || pos >= len(s.entries) {
```
adonovan: Where possible, express interval predicates positively in ascending order (lo < x < hi, or <=):
```
if !(0 <= i && i < len(s.entries)) {
```
as this is make it much easier to correctly comprehend what precondition was established.

starlarkstruct/struct.go
```diff
@@ -246,6 +246,19 @@ func (s *Struct) AttrNames() []string {
 	return names
 }
 
+// AttrAt returns the value of the field at the specified position.
```
adonovan: s/position/index/

adonovan (commented):
Thanks, looks good but for a few minor comments.

adonovan (approved):
LGTM. Thanks!

PR #657, opened 2026-09-06
Update Struct.AttrAt to return name and value

ash2k:
Updates the implementation from #655, as discussed in #654.


starlarkstruct/struct.go
```diff
@@ -247,16 +247,13 @@ func (s *Struct) AttrNames() []string {
 }
 
 // AttrAt returns the value of the field at the specified index.
-func (s *Struct) AttrAt(i int) (starlark.Value, error) {
+// Returns "", nil when field does not exist.
```
adonovan: Since we've exposed Len, we can just make it a precondition that i is in range and let e.entries[i] panic otherwise. (I'm not sure why this wasn't obvious to me before. I suspect I am doing too many tasks, badly.)

adonovan (approved):

adonovan:
Thanks!

